Issue #87, opened 2014-12-19
Provide a CLI interface


mojavelinux:
It would be amazing if Asciidoctor.js could provide the same CLI interface (or at the very least, a minimal CLI interface) just like Asciidoctor. That way, people could install Asciidoctor.js (perhaps using npm) and get the asciidoctor command (or perhaps the asciidoctorjs command, to be consistent with AsciidoctorJ and not cause conflicts).

Mogztter:
See https://github.com/asciidoctor/asciidoctor-cli.js

mojavelinux:
Just tried it and it works great. Nice work!

mojavelinux:
Now that we've published the "asciidoctor-cli" package, it makes me wonder whether we should rename (or alias) the core package as "asciidoctor" instead of (or in addition to) "asciidoctor.js".

```
$ npm install asciidoctor
$ npm install asciidoctor-cli
```

Is it possible to add a package alias? If not, I'm okay with renaming the package. I think we agreed that it's understood that this is a JavaScript package given that we are using npm.

PR #89, opened 2014-12-21
Asciidoctor.js CLI

Mogztter:
Work in progress to discuss implementation detail

Ref #87

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
   "homepage": "https://github.com/asciidoctor/asciidoctor.js",
   "dependencies": {
     "cli": "^0.6.5",
+    "mkdirp": "^0.5.0",
     "opal-npm-wrapper": "git://github.com/anthonny/opal-npm-wrapper.git#0.1.1",
     "xmlhttprequest": "~1.6.0"
   },
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   },
   "homepage": "https://github.com/asciidoctor/asciidoctor.js",
   "dependencies": {
+    "cli": "^0.6.5",
     "opal-npm-wrapper": "git://github.com/anthonny/opal-npm-wrapper.git#0.1.1",
     "xmlhttprequest": "~1.6.0"
   },
--- a/src/npm/asciidoctor-cli.js
+++ b/src/npm/asciidoctor-cli.js
@@ -2,27 +2,84 @@
 var cli = require('cli').enable('status');
 var fs = require('fs');
 var path = require('path');
+var mkdirp = require('mkdirp');
 var asciidoctor = require('../../dist/npm/asciidoctor-core.js')();
 var opal = asciidoctor.Opal;
 var processor = asciidoctor.Asciidoctor(true);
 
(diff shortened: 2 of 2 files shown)


Mogztter:
@mojavelinux @anthonny I'd appreciate review or feedback to get me going :smile:

anthonny:
I'll make a review asap,
regarding the relative path, would it not make sense to create another project for the Cli ? like Grunt with grunt-cli ?
This project will have a dependency on asciidoctor.js
wdyt ?

Mogztter:
> would it not make sense to create another project for the Cli ? like Grunt with grunt-cli ?

Yes why not. Actually I tried to add a dependency on `asciidoctor.js` for asciidoctor-cli (inside `asciidoctor.js`) but that's not working :smile: 

Maybe it's better to create another repository `asciidoctor-cli.js` because the release process is already quite complex in `asciidoctor.js` with one package...

mojavelinux:
I support the idea of a separate repository as that embraces the convention created by grunt. Ideally, that repository name would be `asciidoctor.js-cli`.

Mogztter:
I will create a new repository.

@mojavelinux are you sure about the name ? I think it's better to keep the suffix `.js` at the end of the name. Furthermore, the name of the npm package will be `asciidoctor-cli` (not `asciidoctor.js-cli`).

mojavelinux:
Yep, asciidoctor-cli.js works for me.

I do think that the cli needs to be asciidoctor.js (or asciidoctorjs though). That makes it consistent with how we do it in AsciidoctorJ. Otherwise, the commandline can get mixed up with the Ruby-based one. wdyt?

Mogztter:
> I do think that the cli needs to be asciidoctor.js (or asciidoctorjs though).

I agree

> That makes it consistent with how we do it in AsciidoctorJ. Otherwise, the commandline can get mixed up with the Ruby-based one. wdyt?

Hmm I see, now `asciidoctor.js-cli` makes sense.
I didn't know that AsciidoctorJ has a CLI, the name of the commandline is `asciidoctorj` ?

Our Bower and NPM packages are named `asciidoctor.js`. By convention we should add `-cli` suffix but we already have a suffix ! 
We could remove the `.js` suffix because Bower and NPM are JavaScript tools but the package name must contains the commandline name... Argh naming things! :smile: 

To be conservative I think that `asciidoctor.js-cli` for the package name and the repository name is a safe bet. The commandline should probably be `asciidoctorjs` (dot in commandline name is uncommon).

mojavelinux:
> I didn't know that AsciidoctorJ has a CLI, the name of the commandline is asciidoctorj ?

Yes. You can try it out, btw, by using gvm. See http://gvmtool.net/

I'm okay with renaming the package to `asciidoctor` (and `asciidoctor-cli`) since it's understood that it's JavaScript. I agree with, and prefer, if we can name the commandline `asciidoctorjs`.

Mogztter:
Closing in favor of: https://github.com/asciidoctor/asciidoctor-cli.js

mojavelinux:
\o/

Nice job!

Issue #86, opened 2014-12-15
shell:rakeDist failed


aidanhs:
```
$ grunt dist
[...]
Running "shell:rakeDist" (shell) task
Warning: Command failed: Cannot handle dynamic require :asciidoctor/helpers:20
Cannot handle dynamic require :asciidoctor/converter/factory:49
Cannot handle dynamic require :asciidoctor/converter/factory:197
Cannot handle dynamic require :asciidoctor/converter/factory:202
Cannot handle dynamic require :asciidoctor/converter/factory:207
Cannot handle dynamic require :asciidoctor/converter/factory:215
Cannot handle dynamic require :asciidoctor/converter/factory:218
rake aborted!
NoMethodError: undefined method `write_to' for nil:NilClass
/adb/asciidoctor.js/Rakefile:45:in `block (2 levels) in <top (required)>'
/adb/asciidoctor.js/Rakefile:45:in `each'
/adb/asciidoctor.js/Rakefile:45:in `block in <top (required)>'
Tasks: TOP => dist
(See full trace by running task with --trace)
 Use --force to continue.
```

With --trace added to Gruntfile.js:
```
Running "shell:rakeDist" (shell) task
Warning: Command failed: * Invoke dist (first_time)
* Execute dist
Cannot handle dynamic require :asciidoctor/helpers:20
Cannot handle dynamic require :asciidoctor/converter/factory:49
Cannot handle dynamic require :asciidoctor/converter/factory:197
Cannot handle dynamic require :asciidoctor/converter/factory:202
Cannot handle dynamic require :asciidoctor/converter/factory:207
Cannot handle dynamic require :asciidoctor/converter/factory:215
Cannot handle dynamic require :asciidoctor/converter/factory:218
rake aborted!
NoMethodError: undefined method `write_to' for nil:NilClass
/adb/asciidoctor.js/Rakefile:45:in `block (2 levels) in <top (required)>'
/adb/asciidoctor.js/Rakefile:45:in `each'
/adb/asciidoctor.js/Rakefile:45:in `block in <top (required)>'
/var/lib/gems/1.9.1/gems/rake-10.4.2/lib/rake/task.rb:240:in `call'
/var/lib/gems/1.9.1/gems/rake-10.4.2/lib/rake/task.rb:240:in `block in execute'
/var/lib/gems/1.9.1/gems/rake-10.4.2/lib/rake/task.rb:235:in `each'
[...]

PR #106, opened 2015-03-29
Print a more friendly error message when the extensions-lab directory is missing

Mogztter:
If `extensions-lab` directory is missing (probably because the git submodule is not initialized) you get the following error:
```
rake aborted!
NoMethodError: undefined method `write_to' for nil:NilClass
/workspace/asciidoctor.js/Rakefile:47:in `block (2 levels) in <top (required)>'
/workspace/asciidoctor.js/Rakefile:47:in `each'
/workspace/asciidoctor.js/Rakefile:47:in `block in <top (required)>'
Tasks: TOP => default => dist
(See full trace by running task with --trace)
```

The error message is unclear and the current documentation doesn't tell you to initialize git submodule so there's a high chance that user will see this error message. See https://github.com/asciidoctor/asciidoctor.js/issues/86

I now check that the directory exists and if not prints the following message:
```
Unable to cross-compile extensions because git submodule 'extensions-lab' is not initialized.
To initialize the submodule use the following command `git submodule init` and `git submodule update`.
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -42,9 +42,14 @@ task :dist do
   env['asciidoctor/converter/docbook45'].write_to "build/asciidoctor-docbook45.js#{compress ? '.gz' : nil}"
   env['asciidoctor/converter/docbook5'].write_to "build/asciidoctor-docbook5.js#{compress ? '.gz' : nil}"
 
-  env.append_path 'extensions-lab/lib'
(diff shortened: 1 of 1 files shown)


mojavelinux:
This looks good. Could you rebase so it can be merged?

PR #119, opened 2015-08-01
Add new slide template converters

Mogztter:
Based on @rahmanusta work


Mogztter:
@rahmanusta It's almost working but I got the following exception:

```
document.attr is undefined on line 2
```

I think this is related to this template https://github.com/asciidocfx/asciidoctor.js-reveal-demo/blob/master/revealjs/document.jade#L2

My test page:

```html
<!DOCTYPE html>
<html lang="en" xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
  <head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
    <title>Asciidoctor in JavaScript powered by Opal</title>
    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,300italic,400,400italic,600,600italic%7CNoto+Serif:400,400italic,700,700italic%7CDroid+Sans+Mono:400,700">
    <link rel="stylesheet" href="asciidoctor.css">
	<script src="http://cdnjs.cloudflare.com/ajax/libs/jquery/1.8.3/jquery.js"></script>
  </head>
  <body>
    <div id="content"></div>
    <script src="jade.js"></script>
    <script src="../bower_components/opal/opal/current/opal.js"></script>
    <script src="asciidoctor-core.js"></script>
    <script src="asciidoctor-factory.js"></script>
    <script src="asciidoctor-slide.js"></script>
    <script src="asciidoctor-deckjs.js"></script>
    <script src="asciidoctor-revealjs.js"></script>
    <script>
	  var options = Opal.hash2(['safe', 'attributes'], {
        safe: 'safe',
        attributes: 'showtitle icons=font@ source-highlighter=highlight.js platform=opal platform-opal env=browser env-browser idprefix idseparator=- '
      });
	  var content = ":backend: revealjs\n:templatedir: .\n\n= Title\n\n== Slide 1\n\nContent 1"
	  var doc = Opal.Asciidoctor.$load(content, options);
      var rendered = doc.$convert();
	  console.log("rendered ", rendered);
    </script>
  </body>
</html>
```

rahmanusta:
@Mogztter Let me work on https://github.com/Mogztter/asciidoctor.js/tree/slide. After check, I will make a PR for that.

rahmanusta:
@Mogztter I tried with current jade templates https://github.com/asciidocfx/asciidoctorjs-slide-templates  , it works like a charm! :)

Mogztter:
@rahmanusta I was using https://github.com/asciidocfx/asciidoctor.js-reveal-demo, I will try with https://github.com/asciidocfx/asciidoctorjs-slide-templates (and finish the build task to automate things)

Mogztter:
@rahmanusta Thanks it's working! You can give it a try with `rake examples`, then open in Firefox `build/slide.html`

rahmanusta:
@Mogztter Yes, thanks, it works!!

Also I want to say, I used "grunt examples" command and need to submodule init and submodule update manually.

Also, deckjs slide may work like reveal with templates.

Mogztter:
> Also I want to say, I used "grunt examples" command and need to submodule init and submodule update manually.

Yes we already have to do this manually to initialize extension-lab module.

> Also, deckjs slide may work like reveal with templates.

Yes I need to add another submodule on https://github.com/imakewebthings/deck.js

@mojavelinux The current template converter converts the whole document, is it possible to create an embedded document to ease the integration into the chrome extension and firefox addon ?

mojavelinux:
> @mojavelinux The current template converter converts the whole document, is it possible to create an embedded document to ease the integration into the chrome extension and firefox addon ?

:dizzy_face: 

Sure, I suppose we just need to implement the embedded template.

> I need to add another submodule on https://github.com/imakewebthings/deck.js

Actually, I'd rather just add the template converter infrastructure to Asciidoctor.js and have the actual implementation for deck.js in the deck.js module. That way, the Chrome extension (and such) can depend on asciidoctor-deck.js. Nothing specific to deck.js should be in the asciidoctor.js repo.

Mogztter:
> Sure, I suppose we just need to implement the embedded template.

You're right there's no embedded template, but why the template converter is choosing document template by default ?

> Actually, I'd rather just add the template converter infrastructure to Asciidoctor.js and have the actual implementation for deck.js in the asciidoctor-deck.js component. That way, the Chrome extension (and such) can depend on asciidoctor-deck.js. Nothing specific to deck.js should be in the asciidoctor.js repo.

Sure this is only temporary, I just want to make sure that everything is working before splitting things.
Can we proceed with https://github.com/asciidoctor/asciidoctor-deck.js/pull/1 and https://github.com/asciidoctor/asciidoctor-reveal.js/pull/39 ? I'm really eager to play with this converter in the Chrome extension :smile:

rahmanusta:
> Sure this is only temporary, I just want to make sure that everything is working before splitting things.
Can we proceed with asciidoctor/asciidoctor-deck.js#1 and asciidoctor/asciidoctor-reveal.js#39 ? I'm really eager to play with this converter in the Chrome extension

My PRs isn't accepted yet, I dont know why?

mojavelinux:
I think we're still sorting out where to put all the files.

One thing I noticed in the jade templates for both deck.js and reveal.js is that sometimes you use 2 spaces indent and sometimes you use 4 spaces. I'd prefer if you used 2 spaces. Could you make them consistent?

I also noticed that you are loading firebug by default. I don't think firebug should be loaded by default as the console in most browsers is sufficient for basic usage. I think firebug should be added using a docinfo file if the user wants it.

rahmanusta:
> I think we're still sorting out where to put all the files.
Okey

> One thing I noticed in the jade templates for both deck.js and reveal.js is that sometimes you use 2 spaces indent and sometimes you use 4 spaces. I'd prefer if you used 2 spaces. Could you make them consistent?

My preference is also 2 speaces. I reformatted the template for 2 spaces.

>I also noticed that you are loading firebug by default. I don't think firebug should be loaded by default as the console in most browsers is sufficient for basic usage. I think firebug should be added using a docinfo file if the user wants it.

I'm loading firebug by default in AsciiDocFx, both we think it is not necessary for the common template work. I removed the firebug dependency from document templates.

Mogztter:
> You're right there's no embedded template, but why the template converter is choosing document template by default ?

By default `header_footer` is false ([source](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/document.rb#L247)), I don't understand why the template factory choose document template instead of embedded template.

rahmanusta:
I don't know too @Mogztter , @mojavelinux ?

examples/slide.html
```diff
@@ -0,0 +1,74 @@
...
+<button id="convertBtn">Convert</button>
+<script src="jade.js"></script>
+<script src="../bower_components/opal/opal/current/opal.js"></script>
+<script src="asciidoctor-core.js"></script>
+<script src="asciidoctor-factory.js"></script>
+<script src="asciidoctor-slide.js"></script>
+<script src="asciidoctor-deckjs.js"></script>
+<script src="asciidoctor-revealjs.js"></script>
+<script>
+  var options = Opal.hash2(['safe', 'attributes'], {
```
mojavelinux: Remember we don't need Opal.hash2 anymore. We can just use Opal.hash and wrap a JSON object.

mojavelinux:
> I don't understand why the template factory choose document template instead of embedded template.

It's a bit more complicated than that. What it comes down to is if the header_footer option is not set, it defaults to the "document" template. There are two conditions under which this option is set if not otherwise specified, and we must not be meeting either condition.

See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L1440-L1442

mojavelinux:
When using the API, it's just generally good to specify the header_footer option one way or the other.

mojavelinux:
> I reformatted the template for 2 spaces.

Thank you!

rahmanusta:
Thank you @mojavelinux .

@Mogztter How is work going ? Do you need anything from me ?

Mogztter:
See #219

PR #124, opened 2015-08-27
Fixes #123: Runs TravisCI builds in a container

### LINKED ISSUE #123: Update TravisCI to use a container
TravisCI offers the option to run in a container, supposedly is faster, better, ...
From my experience I've seen that builds start immediately on a PR, other than that I cannot appreciate much differences.

I have used a asciidoctor.js fork as a test ground and confimed it only requires one line as long as you don't need to use `sudo`: https://github.com/abelsromero/asciidoctor.js/blob/master/.travis.yml#L2.

Details here:
http://docs.travis-ci.com/user/migrating-from-legacy/

PS: This may be interesting for more projects, just copy who you consired.

gastaldi:


Mogztter:
Thanks :+1:

PR #125, opened 2015-09-20
Resolves #116, implement methods to read assets

### LINKED ISSUE #116: When header_footer is set to true and safeMode is set to safe or unsafe, building .html files fails.
Hi,

I'm using Asciidoctor.js and grunt to configure it. I would like to use includes, when safeMode should be set to safe or unsafe (instead of secure that is default), but after I change that, building .html files fails. Only error that is given is to log is "undefined is not a function". If I remove option "header_footer: true" from configuration .html files are built, but I'm missing all the information that is related to styling and also Table of Contents is missing after that. Any Ideas, what I'm doing wrong?

Here are my configuration parameters in Gruntfile:

        asciidoctor: {
            default_options: {
                options: {
                    cwd: 'app/docs/',
                    showTitle: true,
                    showNumberedHeadings: true,
                    showToc: true,
                    safeMode: 'unsafe',
                    //header_footer: true,
                    doctype: 'article',
                    backend: 'html5'
                },
                files: {
                    'app/docs_output/': ['**/*.adoc', '!includes/**/*.adoc', '**/*.md']
                }
            }
        },

I also have other problem, where table of contents is not placed to right block of the pace. Even if I add these line to .adoc file:

:toc: right

nothing happens.

Mogztter:
Ref #116

### DIFF
--- a/spec/share/common-specs.js
+++ b/spec/share/common-specs.js
@@ -53,6 +53,12 @@ var commonSpec = function(Opal, Asciidoctor) {
       expect(html).toContain('<h3 id="_test">Test</h3>');
     });
 
+   it('=== Test should embed assets', function() {
+     var options = Opal.hash({doctype: 'article', safe: 'unsafe', header_footer: true, attributes: ['showtitle', 'stylesheet=asciidoctor.css', 'stylesdir=../../dist/css']});
+     var html = Asciidoctor.$convert('=== Test', options);
+     expect(html).toContain('Asciidoctor default stylesheet');
+    });
+
     it('backend=docbook45 should produce a docbook45 document', function() {
       var options = Opal.hash({'attributes': ['backend=docbook45', 'doctype=book'],'header_footer':true});
       var html = Asciidoctor.$convert(':doctitle: DocTitle\n:docdate: 2014-01-01\n== Test', options);


Mogztter:
@pavelhoral You can see an example on how to workaround the issue: https://github.com/asciidoctor/asciidoctor.js/pull/125/files#diff-d6b82947a72b34a905618203e8806467R25

Note that you will need to set two attributes:

```
stylesheet=asciidoctor.css
stylesdir=node_modules/asciidoctor.js/dist/css
```

I don't know yet the best approach but I think Asciidoctor.js should set this values.

mojavelinux:
Since we are so near to releasing 1.5.3 core (days away), I think we should get these changes into 1.5.3 and then just upgrade. That way, we don't need to duplicate it in Asciidoctor.js.

mojavelinux:
But definitely keep the test.

Mogztter:
The test will fail until we use Asciidoctor core 1.5.3 or until #132 is merged (and this branch rebased onto master)

mojavelinux:
> The test will fail until we use Asciidoctor core 1.5.3 or until #132 is merged (and this branch rebased onto master)

Then hopefully we can get this wrapped up soon :)

mojavelinux:
Can you make the test conditional on the Asciidoctor version, perhaps? If you want to wait, that's fine too. We can merge this last.

Mogztter:
And green bars :smile:

mojavelinux:
:+1:

PR #133, opened 2015-10-15
Resolves #132, Move opal_ext from core to Asciidoctor.js

### LINKED ISSUE #132: Move opal_ext from core to Asciidoctor.js
Since Asciidoctor.js now has its own build and tests (i.e., its own lifecycle), I think we should move the [opal_ext](https://github.com/asciidoctor/asciidoctor/tree/master/lib/asciidoctor/opal_ext) folder from core to Asciidoctor.js.

This will also allow us to make any changes necessary when Opal is upgraded without waiting on a core release. The only Opal-related code that would remain in core are the internal patches necessary to handle logic that is specific to the Opal environment.

Mogztter:


mojavelinux:
:+1: However, I think we should still call the directory `opal_ext` instead of `core_ext`. While it's true that the code is (mostly) modifying core behavior, it's really a patch to fix deficiencies in Opal. That's why I like `opal_ext`. Plus, we don't want to accidentally catch modifications of gems that use the core_ext path (not very likely, but still a risk). If you agree, can you update?

I'll prepare a pull request to remove these from core.

Rakefile
```diff
@@ -38,10 +38,12 @@ task :dist do
   #env['asciidoctor'].write_to "build/asciidoctor.js#{compress ? '.gz' : nil}"
   asciidoctor = env['asciidoctor']
   # NOTE hack to make version compliant with semver
-  asciidoctor_src = env['asciidoctor/core_ext/match_data'].source + env['asciidoctor'].source
+  core_ext_files = ['dir', 'error', 'file', 'match_data', 'opal_ext']
```
mojavelinux: Why not just require `asciidoctor/opal_ext` and declare requires in that file instead of itemizing each one in the build?

mojavelinux:
You seem to be missing this file:

https://github.com/asciidoctor/asciidoctor/blob/e9f908f4c6c344fa0cc07c7d0793b42546dca0e6/lib/asciidoctor/opal_ext.rb

mojavelinux:
Also I don't think you should lose this comment:

```
require 'encoding' # needed for String.bytes method
```

(I don't understand why Opal is not including this library by default).

Mogztter:
> You seem to be missing this file...which defines the JAVASCRIPT_PLATFORM constant.

Oh you're right, tests are running fine because Asciidoctor core still contains this file.

> Why not just require asciidoctor/opal_ext and declare requires in that file instead of itemizing each one in the build?

Yes it's way better with `require` inside `opal_ext.rb`

> Also I don't think you should lose this comment:

Done

mojavelinux:
Excellent!

I've got the PR read for core to remove these files. When do you think we should make the switch? If I merge in core, will you switch the build over to use the development version of core until the 1.5.3 release? I'll wait for your thumbs up before changing core.

mojavelinux:
Here's the core PR: https://github.com/asciidoctor/asciidoctor/pull/1518

Mogztter:
> If I merge in core, will you switch the build over to use the development version of core until the 1.5.3 release? 

Yes sure, you can merge the PR in core! Moving forward :+1:

Rakefile
```diff
@@ -35,13 +34,11 @@ task :dist do
...
 
-  #env['asciidoctor'].write_to "build/asciidoctor.js#{compress ? '.gz' : nil}"
   asciidoctor = env['asciidoctor']
+  asciidoctor_src = env['asciidoctor/opal_ext'].source + env['asciidoctor'].source
   # NOTE hack to make version compliant with semver
-  asciidoctor_src = env['asciidoctor/core_ext/match_data'].source + env['asciidoctor'].source
   asciidoctor_src = asciidoctor_src.sub(/'VERSION', "(\d+\.\d+.\d+)\.(.*)"/, '\'VERSION\', "\1-\2"')
-  asciidoctor_src = asciidoctor_src.sub(/^( *)(self.\$require\("asciidoctor\/core_ext"\);)/,
-      %(\\1\\2\n\\1self.$require("asciidoctor/core_ext/match_data");))
+  asciidoctor_src = asciidoctor_src.sub(/^( *)(self\.\$require\("set"\);)/, %(\\1\\2\n\\1self.$require("asciidoctor/opal_ext");))
```
Mogztter: I'm searching for this line of code but I'm not convinced that this is a good idea: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L8

@mojavelinux Thoughts ?

mojavelinux: We're going to go ahead and add the require statement back in so we don't need to do this.

Mogztter:
I think this commit broke our tests: https://github.com/asciidoctor/asciidoctor/commit/03b78e1b1c2b9b4ac2aa5b066f31bf5b341691ae

I don't know why yet, I just saw this error on Travis: `NameError: uninitialized constant DocBook5Converter::DLIST_TAGS`

I will try to reproduce this issue tomorrow on my local copy, stay tuned!

mojavelinux:
Uh oh. I can reason why that's the case. Probably because the scope of constant isn't available within the same assignment statement. That's likely a bug in Opal, though I'm now wondering if it's defined in Ruby or just happens to work. We might be able to patch it with a regexp for now. It seems like the DLIST_TAGS is the only case where it tries to backreference a constant in the assignment of default.

mojavelinux:
I need to make sure to add the step to look at the Asciidoctor.js test results in the release process of core. (I'm putting together a step-by-step guide).

mojavelinux:
Btw, the way I would patch it is to look for `DLIST_TAGS['labeled']` and just replace it with the resolved value of `{ :list  => 'variablelist', :entry => 'varlistentry', :term  => 'term', :item  => 'listitem' }`

Mogztter:
You're right, I fixed the issue, now tests are running fine.

Mogztter:
Green bars \o/

mojavelinux:
Excellent! Good work!

PR #161, opened 2016-01-06
add benchmark script

mojavelinux:


Mogztter:
LGTM but I will postpone the merge because I rewrote the "fetch sample data" task with Node in #167. I will move this method in a reusable module so we can use it with the benchmark script.

mojavelinux:
No problem at all. In fact, I'm glad because I couldn't figure out how to download something in JavaScript (the script would terminate before the download even started...so I gave up and just did it in Ruby).

Mogztter:
Follow-up #170

mojavelinux:
Excellent. Let's close this one then. Nice work!

PR #164, opened 2016-01-09
Resolves #73, Publish dist files to a tag only

### LINKED ISSUE #73: Publish dist files to a tag only
The dist files should only be published to the git repository under a tag. Other option is to push the dist files to a dedicated git repository like asciidoctor.js-dist (though, I'm leading more towards tags). Otherwise, we are constantly checking in the build output, which is unnecessary overhead.

What might help to kick off this task is a link to an article about how to manage published artifacts generated by grunt so that bower / npm can still consume them.

Mogztter:


mojavelinux:
Awesome!

PR #166, opened 2016-01-09
Migrate Rake task 'dist' to npm

Mogztter:
One more step toward a 100% JavaScript build :wink:


npm/rake-dist.js
```diff
@@ -0,0 +1,49 @@
...
+var Build = require('./build.js');
+var build = new Build();
+var log = new Log();
+
+if (!fs.existsSync('build')) {
+  fs.mkdirSync('build');
+}
+
+var opalCompile = function(require, output) {
+  var stdout = child_process.execSync('opal --compile -I lib -I asciidoctor -g asciidoctor -r ' + require + ' --dynamic-require ignore --no-opal --no-exit > build/' + output);
```
mojavelinux: Perhaps move the opal options (e.g., `--dynamic-require ignore --no-opal --no-exit` and maybe `--compile`) to a global variable so we can easily tweak them?

npm/rake-dist.js
```diff
@@ -0,0 +1,49 @@
...
+opalCompile('asciidoctor/converter/docbook5', 'asciidoctor-docbook5.js');
+opalCompile('asciidoctor/converter/docbook45', 'asciidoctor-docbook45.js');
+opalCompile('asciidoctor/extensions', 'asciidoctor-extensions.js');
+opalCompile('asciidoctor', 'asciidoctor-core.js');
+
+build.replaceFileSync('build/asciidoctor-core.js', /'VERSION', "(\d+\.\d+.\d+)\.(.*)"/, "\1-\2");
+
+log.title('compile extensions-lab libraries');
+if (fs.existsSync('extensions-lab/lib')) {
+  opalCompileExtension('chrome-inline-macro');
```
mojavelinux: Perhaps make an `opalCompileExtensions` method that takes a collection?

npm/rake-dist.js
```diff
@@ -0,0 +1,49 @@
...
+  process.exit(9);
+}
+
+log.title('copy resources');
+log.debug('copy asciidoctor.css');
+// FIXME: Find a more robust implementation to get the Asciidoctor gem install path
+asciidoctorGemPath = child_process.execSync("ruby -e \"print (Gem::Specification.find_by_name 'asciidoctor').full_gem_path\"");
+asciidoctorCSSFile = asciidoctorGemPath + '/data/stylesheets/asciidoctor-default.css';
+fs.createReadStream(asciidoctorCSSFile).pipe(fs.createWriteStream('build/asciidoctor.css'));
+fs.createReadStream(asciidoctorCSSFile).pipe(fs.createWriteStream('examples/asciidoctor.css'));
```
mojavelinux: I don't think we need the stylesheet in the examples directory. That was only there because of the legacy Rake task. I think we can just reference the one in build.

mojavelinux:
Fantastic! I added a few inline comments, but overall, solid!

npm/rake-dist.js
Mogztter: Done in #167: https://github.com/asciidoctor/asciidoctor.js/pull/167/files#diff-217cda23d47654a773d88858ee3c9177R14

mojavelinux: Fantastique !

npm/rake-dist.js
```diff
@@ -0,0 +1,51 @@
...
+if (!fs.existsSync('build')) {
+  fs.mkdirSync('build');
+}
+
+var opalCompile = function(require, output) {
+  var stdout = child_process.execSync('opal --compile -I lib -I asciidoctor -g asciidoctor -r ' + require + ' --dynamic-require ignore --no-opal --no-exit > build/' + output);
+  process.stdout.write(stdout);
+}
+
+var opalCompileExtensions = function(names) {
```
mojavelinux: This is a bit nit picky, but you could also make this varargs by using the implicit arguments parameter.

```
var opalCompileExtensions = function() {
  arguments.forEach(opalCompileExtension);
}
```

mojavelinux:
One more inline comment. Then, we're good to go!

npm/rake-dist.js
Mogztter: In your example `arguments` is a JSON object so `forEach` is undefined but I can write:

```javascript
var opalCompileExtensions = function(names) {
  names.forEach(opalCompileExtension);
}
```

mojavelinux: Perfect!

Mogztter:
@mojavelinux Opal compile command line produce a slightly different result, the compiled code is not put inside an Opal module.

Before:
```javascript
Opal.modules["emoji-inline-macro/extension"] = function(Opal) {
  // code...
}
```

After:
```javascript
(function(Opal) {
  // code
}
```

This mean that the code (module) is loaded automatically (same behavior as Opal < 0.8). Overall I think this is better, we could even remove `autoLoadModule` logic: https://github.com/asciidoctor/asciidoctor.js/blob/master/manual.adoc#disable-auto-require-module

What do you think ?

mojavelinux:
Sounds good to me. I think that simplifies things.

One a side note, I'm slightly concerned that the extensions are not under the asciidoctor namespace, though I guess that issue goes away if the code doesn't get wrapped in a module.

PR #167, opened 2016-01-09
Migrate Rake task 'examples' to npm

Mogztter:


npm/rake-examples.js
```diff
@@ -0,0 +1,44 @@
...
+var build = new Build();
+var log = new Log();
+
+var examplesBuildDir = 'build/examples';
+if (!fs.existsSync(examplesBuildDir)) {
+  fs.mkdirSync(examplesBuildDir);
+}
+
+log.title('compile examples');
+stdout = child_process.execSync('opal --compile examples/asciidoctor_example.rb --dynamic-require warning --no-opal > ' + examplesBuildDir + '/asciidoctor_example.js');
```
mojavelinux: Is it possible to extract an "Opal compile" method that is shared across npm scripts? In the future, Opal may even provide an API...wouldn't that be nice.

mojavelinux:
:+1:

PR #177, opened 2016-03-20
Resolves #168, rewrite Rake jdk8_ea and jdk9_ea tasks in Node

### LINKED ISSUE #168: Rewrite Rake jdk8_ea and jdk9_ea task in Node
Once this is done we can remove Rake from the build process.

Related to #72

Mogztter:
Resolves #168


mojavelinux:
Sweet!

Not only does this help consolidate the build, it's also a really useful reference for using Node to do various tasks. I will definitely be taking notes!

Mogztter:
Yes! 
I'm also making progress on compiling Ruby to JS in JS. I just need to find a solution to resolve `require` without using Gem module (otherwise we need to re-implement this module in JavaScript).

Issue #152, opened 2015-12-17
Upgrade to Opal 0.9.2


Mogztter:
https://github.com/opal/opal/tree/v0.9.2

Mogztter:
0.9.0 is released: https://github.com/opal/opal/tree/v0.9.0

https://github.com/opal/opal/blob/v0.9.0/CHANGELOG.md#090-edge

Require attention:
> Renamed:
Hash.keys => Hash.$$keys
Hash.map => Hash.$$map
Hash.smap => Hash.$$smap

mojavelinux:
I wonder if we should alias these for compatibility.

Mogztter:
@mojavelinux Yes definitively! A good candidate for the first public API of Asciidoctor.js (https://github.com/asciidoctor/asciidoctor.js/issues/92). We could also provide a simpler syntax because I'm not a big fan of `doc.attributes.smap`. `doc.attributes.smap['sectnums']` -> `doc.attributes['sectnums']`

mojavelinux:
I absolutely agree. We want to present something as close to the Ruby API as possible so there isn't extra friction.

mojavelinux:
Seems like we're up to Opal 0.9.2 now.

Mogztter:
I saw 0.9.1 yesterday, must have been an error during the release.

mojavelinux:
:+1:

PR #179, opened 2016-04-03
fix(commonJs): change the test to check the JS platform

anthonny:
To know if the script is use in a browser, it now check the typeof window instead of XMLHttpRequest
Tests for bower have some code to fake browser and XMLHttpRequest
I have added Karma with PhantomJS to run the CommonJS tests run in a real browser context
The CommonJS tests use browserify

The CommonJS tests contain a test that generate an error because of a typo in Opal, it must pass after an fix in the Opal dependencies.

The problem is here https://github.com/anthonny/opal-npm-wrapper/blob/master/index.js#L18320 and 18321:
`typeof(process) === 'object'`

The test should be:
`typeof(process) === 'object' && !process.browser`

Tested and approuved after some tests on HubPress

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,35 @@
+'use strict';
+
+module.exports = function(karma) {
+  karma.set({
+    frameworks: [ 'jasmine', 'browserify' ],
+
+    files: [
+      {pattern: 'spec/share/include.adoc', watched: false, included: false, served: true},
+      {pattern: 'spec/share/sample.csv', watched: false, included: false, served: true},
+      {pattern: 'build/asciidoctor.css', watched: false, included: false, served: true},
+      'spec/commonJs/asciidoctor-all.spec.js'
+    ],
+
+    reporters: [ 'dots' ],
+
+    preprocessors: {
+      'spec/commonJs/asciidoctor-all.spec.js': [ 'browserify' ]
+    },
+
+    browsers: [ 'PhantomJS' ],
+
+    logLevel: 'LOG_DEBUG',
+
+    singleRun: true,
+    autoWatch: false,
+    // force the port to use
+    port: 9876,
+
+    // browserify configuration
+    browserify: {
+      debug: true,
+      transform: [ ]
+    }
+  });
+};
--- a/lib/asciidoctor/opal_ext.rb
+++ b/lib/asciidoctor/opal_ext.rb
@@ -1,23 +1,36 @@
 %x(
-  var value;
-  if (typeof module !== 'undefined' && module.exports) {
+  var isNode = typeof module !== 'undefined' && module.exports,
+      isBrowser = typeof window !== 'undefined',
+      isNashorn = typeof Java !== 'undefined' && Java.type,
+      isRhino = typeof java !== 'undefined',
+      value;
+
+
+  // With browserify we can use module and XMLHttpRequest, so for node
+  // we must test that we have not XMLHttpRequest
+  if (isNode && !isBrowser) {
     value = 'node';
   }
-  else if (typeof XMLHttpRequest !== 'undefined') {
+  else if (isBrowser) {
   // or we can check for document
   //else if (typeof document !== 'undefined' && document.nodeType) {
     value = 'browser';
   }
-  else if (typeof Java !== 'undefined' && Java.type) {
+  else if (isNashorn) {
     value = 'java-nashorn';
   }
-  else if (typeof java !== 'undefined') {
+  else if (isRhino) {
     value = 'java-rhino';
   }
   else {
     // standalone is likely SpiderMonkey
     value = 'standalone';
   }
+  // Extra lines to debug tests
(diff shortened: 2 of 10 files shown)


anthonny:
I just made a new branch in https://github.com/anthonny/opal-npm-wrapper called [fix-stderr-stdout](https://github.com/anthonny/opal-npm-wrapper/tree/fix-stderr-stdout)
I will use it in my fork of __asciidoctor.js__ and use my fork in HubPress.

Tell me if you want to merge this quick fix in the master

Mogztter:
Thanks @anthonny 

> The problem is here https://github.com/anthonny/opal-npm-wrapper/blob/master/index.js#L18320 and 18321:
> `typeof(process) === 'object'`
> The test should be:
> `typeof(process) === 'object' && !process.browser`

I think we need to open a pull request with your fix in Opal. The corresponding lines are https://github.com/opal/opal/blob/master/opal/corelib/io.rb#L72-L73

In the meantime you could monkey patch in `opal-npm-wrapper` and publish a new release. Could you do a [0.9.2 release](https://github.com/opal/opal/releases/tag/v0.9.2) of [opal-npm-wrapper](https://github.com/anthonny/opal-npm-wrapper) ?

> Tell me if you want to merge this quick fix in the master

Sure as soon as we can build Asciidoctor.js against a release of Opal npm wrapper.

anthonny:
the 0.9.2 of Opal is not available yet, there is not tag in [opal-cdn](https://github.com/opal/opal-cdn/releases) :/

capture d ecran 2016-04-03 a 22 10 35

Mogztter:
Then I need to work on that ;)
Sorry I though I've published the release on the CDN.
Le 3 avr. 2016 10:11 PM, "Anthonny Quérouil" <notifications@github.com> a
écrit :

mojavelinux:
Karma! Yeah! These tests are getting serious now :)

anthonny:
I made some tests with the version 0.9.2 ... `smap` is now `$$smap` and `smap` is `undefined`
:cry:

anthonny:
I made some tests with the version 0.9.2 ... `smap` is now `$$smap` and `smap` is `undefined`
:cry:

anthonny:
I think that we should really work on stable API that will not change for each version of Opal

mojavelinux:
:+1:

Mogztter:
> I made some tests with the version 0.9.2 ... smap is now $$smap and smap is undefined

https://github.com/asciidoctor/asciidoctor.js/issues/152#issuecomment-170220427

> I think that we should really work on stable API that will not change for each version of Opal

:+1:

Mogztter:
> I think that we should really work on stable API that will not change for each version of Opal

@anthonny I will be grateful if you could set the foundation for building a stable API above Asciidoctor.js/Opal :wink:

Mogztter:
> I think that we should really work on stable API that will not change for each version of Opal

Unfortunately there's no easy way to be backward compatible. `Hash.smap` is an internal property (instance variable) and we cannot add another instance variable...

The only thing we can do is create a new method:

```javascript
// before
doc.attributes.smap['icons'];

// after (choose one)
doc.attributes.smap('icons');
doc.attributes.map('icons');
doc.attributes.fetch('icons');
doc.attributes.get('icons');
```

@mojavelinux @anthonny Wdyt ?

mojavelinux:
We really shouldn't be using the attributes map directly, unless there is a special case. The call should be:

```js
doc.$attr('icons')
```

Would you agree?

In my tests, I could still access the icons through the smap too:

```js
var opts = Opal.hash({safe: 'safe'});
var doc = Opal.Asciidoctor.$load('= Document Title\n:icons: font\n\ncontent', opts);
doc.attributes.smap['icons'] === 'font'
```

The tests seem to confirm this.

Mogztter:
doc.$attr('icons') is fine. Though, can we remove the $ symbol if we create
a wrapper/API around Opal ?

smap is definitely gone in Opal 0.9.2. If you can still access this
attribute, I think you are running in Opal 0.9.0.beta2 (master is still in
this version)

mojavelinux:
> doc.$attr('icons') is fine. Though, can we remove the $ symbol if we create a wrapper/API around Opal?

Indeed. This is where the stable API comes into play.

>  I think you are running in Opal 0.9.0.beta2 (master is still in this version)

Yep, that explains it.

Mogztter:
@anthonny Could you please rebase on master ? I've switched master to Opal 0.9.2

anthonny:
Done

Mogztter:
Thanks @anthonny

Mogztter:
On my machine I get this error:

```
PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  ReferenceError: Can't find variable: require
  at /home/guillaume/workspace/opensource/asciidoctor.js/spec/commonJS/asciidoctor-all.spec.js:1
```

@anthonny Any idea?

Issue #92, opened 2014-12-23
Create a JavaScript API to make extensions easier to write and register


Mogztter:
@mojavelinux 

> Just like we have the Groovy DSL for AsciidoctorJ to make extensions easier to write and register, I think we can offer a JavaScript API that is optimized for the author. It's completely reasonable to hide some of the Opal interfaces behind a simpler, clearer JavaScript API.
> 
> We should aim to make it as similar to the original Asciidoctor API as possible, where possible, so it's easy to reason about the relationship.

Related to #84 #76

Mogztter:
This is harder than I though :persevere:

Mogztter:
I define methods explicitly and provide a convenient method to create `Opal.hash2`:
```javascript
Opal.Asciidoctor.Extensions.register = function (block, name) {
  Opal.block_send(Opal.Asciidoctor.Extensions, 'register', block, name);
};

Opal.Asciidoctor.Extensions.Registry._proto.inline_macro = function (block, name) {
  return Opal.block_send(this, 'inline_macro', block, name);
};

Opal.Asciidoctor.Extensions.InlineMacroProcessor._proto.process = function (block) {
  return Opal.block_send(this, 'process', block);
};

Opal.to_hash = function (data) {
  var keys = [];
  for (var key in data) {
    if (data.hasOwnProperty(key)) {
      keys.push(key);
    }
  }
  return Opal.hash2(keys, data);
};
```
**Before:**
```javascript
Opal.block_send(Opal.Asciidoctor.Extensions, 'register', function () {
  Opal.block_send(this, 'inline_macro', function () {
    var inlineMacro = this;
    inlineMacro.$parse_content_as("text");
    Opal.block_send(inlineMacro, 'process', function (parent, target, attrs) {
      var link = "chrome:" + target;
      var text = attrs.map['text'];
      if (!text) {
        text = link;
      }
      return (inlineMacro.$create_anchor(parent, text, Opal.hash2(["type", "target"], {"type": "link", "target": link}))).$render();
    });
  }, "chrome");
}, "uri_schemes");
```

**After:**
```javascript
Opal.Asciidoctor.Extensions.register(function () {
  this.inline_macro(function () {
    var inlineMacro = this;
    inlineMacro.$parse_content_as("text");
	this.process(function (parent, target, attrs) {
      var link = "chrome:" + target;
      var text = attrs.map['text'];
      if (!text) {
        text = link;
      }
[...]

PR #180, opened 2016-04-16
Create a wrapper API around Opal

### LINKED ISSUE #94: Support pure JSON objects in place of Opal.hash
Ideally, the Asciidoctor APIs should be able to accept a regular JSON data structure where a Hash is required instead of Opal.hash2. The latter is awkward and redundant (keys have to be listed twice). It would feel much more comfortable for developers if they could continue to use the familiar JSON data structure.

Mogztter:
This is work in progress.
One of my concern is that we need to cover 100% of the Asciidoctor core API if we want to be consistent. Otherwise users will never know if they have to use `$` or not when calling a method.

#### Before
```javascript
var options = Opal.hash({'attributes': 'icons=font@'});
var doc = Asciidoctor.$load('== Test', options);
console.log(doc.attributes.smap['icons']); // display font
```
#### After
```javascript
var doc = Asciidoctor.load('== Test', {'attributes': 'icons=font@'});
console.log(doc.getAttr('icons')); // display font
```

#### Before
```javascript

Opal.block_send(Opal.Asciidoctor.Extensions, 'register', function () {
  Opal.block_send(this, 'inline_macro', function () {
    var inlineMacro = this;
    inlineMacro.$parse_content_as("text");
    Opal.block_send(inlineMacro, 'process', function (parent, target, attrs) {
      var link = "chrome:" + target;
      var text = attrs.map['text'];
      if (!text) {
        text = link;
      }
      return (inlineMacro.$create_anchor(parent, text, Opal.hash2(["type", "target"], {"type": "link", "target": link}))).$render();
    });
  }, "chrome");
}, "uri_schemes");
```
#### After
```javascript
Opal.Asciidoctor.Extensions.register(function () {
  this.inline_macro(function () {
    var inlineMacro = this;
    inlineMacro.$parse_content_as("text");
    this.process(function (parent, target, attrs) {
      var link = "chrome:" + target;
      var text = attrs.map['text'];
      if (!text) {
        text = link;
      }
      return (inlineMacro.$create_anchor(parent, text, Opal.to_hash({"type": "link", "target": link}))).$render();
    });
  }, "chrome");
}, "uri_schemes");
```

Resolves #94 #92

/cc @anthonny @mojavelinux


mojavelinux:
I really like where this is headed. So much cleaner.

Is it typical in JavaScript to use camelcase instead of underscore-delimited words for method names? (`loadFile` or `load_file`)?

In addition to `attr`, we should add `isAttr` (or `is_attr`).

The real question is whether this API should look more like the core API or the AsciidoctorJ API. We should definitely pick one or the other and try to align with it.

Mogztter:
> Is it typical in JavaScript to use camelcase instead of underscore-delimited words for method names?

I think camelCase is a typical naming convention in JavaScript:
 * https://github.com/airbnb/javascript#naming--camelCase
 * https://github.com/rwaldron/idiomatic.js/#naming
 * https://google.github.io/styleguide/javascriptguide.xml#Naming
 * http://contribute.jquery.org/style-guide/js/#naming-conventions

> In addition to attr, we should add isAttr (or is_attr).

`attr?` translates easily to "is this attribute exists ?".
`is` something is a well know convention (ie. the method will returns a boolean value) but I think `isAttr` is unclear. In my opinion, `hasAttr` or `attrExists` is more explicit.
AsciidoctorJ already adopts this convention so I think we should stick to it but I'm not a big fan :wink: 

In this case `attr` should be renamed `getAttr` right ?

mojavelinux:
> In this case `attr` should be renamed `getAttr` right ?

I'm kind of torn by this. My preference is to keep it as simple as possible. I suppose we could support both `attr` and `getAttribute`. I don't like the abbreviated form, `getAttr`.

> I think `isAttr` is unclear.

You're right, it doesn't fully represent what this method does and it does read well.

Perhaps `attrSet` (or `isAttrributeSet`) or `attrDefined` (or `isAttributeDefined`)? To see an example:

```js
doc.attrSet('icons', 'font')

doc.isAttributeSet('icons', 'font')

doc.attrDefined('icons', 'font')

doc.isAttributeDefined('icons', 'font')
```

I don't think there is a wrong answer here. We just need to find something that feels comfortable (that's how I define most of the APIs...it just feels right).

Mogztter:
I do like `isAttributeDefined` (but I'm using autocompletion on Java :wink:)

//cc @anthonny @hsablonniere Any preference or advice for this API ?

hsablonniere:
I two cents on this (easier to say than to implement :trollface:) :

* I agree on the **We should align APIs as much as possible** while avoiding unconventional stuffs.
* I prefer longer and more explicit stuffs than short abbreviation but if the Java or Ruby API is an abbreviation, I would keep it that way.

What does `isAttributeDefined` do? It looks like it's being called with a name and a value.

Mogztter:
> I agree on the We should align APIs as much as possible while avoiding unconventional stuffs.
> I prefer longer and more explicit stuffs than short abbreviation but if the Java or Ruby API is an abbreviation, I would keep it that way.

Ruby API is using succinct verbs and words delimited with underscores, I think this is pretty unconventional in the JavaScript world.

Java API is using camel case naming with `get`, `set`, `is`, `has` prefix: https://github.com/asciidoctor/asciidoctorj/blob/master/asciidoctorj-core/src/main/java/org/asciidoctor/ast/AbstractNode.java

> I don't like the abbreviated form, getAttr.

@mojavelinux Maybe we should stick with the abbreviated form for now ? or we could also convince AsciidoctorJ to switch to keep APIs aligned as much as possible ?

> What does isAttributeDefined do? It looks like it's being called with a name and a value.

The second argument is the expected value. So you are asking is this attribute defined with this value.

hsablonniere:
> Java API is using camel case naming with get, set, is, has prefix:

Then maybe we just need to have the same API as Java. I mean, isn't it the [same thing](https://twitter.com/cyril_lakech/status/664027910936911872)? :trollface:

> The second argument is the expected value. So you are asking is this attribute defined with this value.

OK, thx.

Mogztter:
I'm sure @ldez has an opinion on this ? :wink:

ldez:
It's important to respect specific language conventions.

Don't respect specific language conventions, and an user will think:
> This code have been writing by a language noob.
> I'm not secured with library, it's can be unstable.

Use smake-case instead of camel-case, for a JS or Java dev it's a bad sign.

A user who use Asciidoctor in Java also use Asciidoctor in JS or Ruby? I don't think it's a common case.

If you must begin a name change somewhere, why not here?

And delegation is always possible:

```js
getAttribute: function(args) {...}
getAttr: function() {
  getAttribute(args);
}
```

my 2 cents.

mojavelinux:
> @mojavelinux Maybe we should stick with the abbreviated form for now ? or we could also convince AsciidoctorJ to switch to keep APIs aligned as much as possible ?

I was just looking at AsciidoctorJ the other day thinking, "why isn't it `getAttribute`?" I think it should be. There is still a chance to make this change for AsciidoctorJ 1.6.x.

I agree with @ldez that the API should adhere the conventions of the community consuming it. In Ruby, everything is super short and abbreviated, so that is totally normal. But in Java / JavaScript, the formal names are more common.

Mogztter:
### TODO List

All the public functions defined in `asciidoctor.rb`, `abstract_node.rb` and `document.rb`.

#### Asciidoctor API

 * [x] `load input, options = {}`: `load(input, options)`
 * [x] `load_file filename, options = {}`: `loadFile(filename, options)`
 * [x] `convert input, options = {}`: `convert(input, options)`
 * [x] `convert_file filename, options = {}`: `convertFile(filename, options)`

#### AbstractNode API

 * [ ] `inline?`: `isInline()`
 * [ ] `block?`: `isBlock()`
 * [x] `attr(name, default_value = nil, inherit = true)`: `getAttribute(name, defaultValue, inherit)`
 * [x] `attr?(name, expect = nil, inherit = true)`: `isAttribute(name, expectedValue, inherit)`
 * [x] `set_attr name, value, overwrite = true`: `setAttribute(name, value, overwrite)`
 * [ ] `set_option(name)`
 * [ ] `option?(name)`
 * [ ] `update_attributes(attributes)`
 * [ ] `converter`
 * [ ] `role?(expect = nil)`
 * [ ] `role`
 * [ ] `has_role?(name)`
 * [ ] `roles`
 * [ ] `add_role(name)`
 * [ ] `remove_role(name)`
 * [ ] `reftext?`
 * [ ] `reftext`
 * [ ] `icon_uri name`
 * [ ] `media_uri(target, asset_dir_key = 'imagesdir')`
 * [ ] `image_uri(target_image, asset_dir_key = 'imagesdir')`
 * [ ] `generate_data_uri(target_image, asset_dir_key = nil)`
 * [ ] `generate_data_uri_from_uri image_uri, cache_uri = false`
 * [ ] `read_contents target, opts = {}`
 * [ ] `read_asset(path, opts = {})`
 * [ ] `normalize_web_path(target, start = nil, preserve_uri_target = true)`
 * [ ] `normalize_system_path target, start = nil, jail = nil, opts = {}`
 * [ ] `normalize_asset_path(asset_ref, asset_name = 'path', autocorrect = true)`
 * [ ] `relative_path(filename)`
 * [ ] `is_uri? str`


#### Document API

 * [ ] `parse data = nil`
 * [ ] `counter(name, seed = nil)`
 * [ ] `counter_increment(counter_name, block)`
 * [ ] `register(type, value, force = false)`
 * [ ] `footnotes?`
 * [ ] `footnotes`
 * [ ] `nested?`
 * [ ] `embedded?`
 * [ ] `extensions?`
 * [ ] `source`
 * [ ] `source_lines`
 * [ ] `doctype`
 * [ ] `backend`
 * [ ] `basebackend? base`
 * [ ] `title`
 * [ ] `title=(title)`
 * [ ] `doctitle opts = {}`
 * [ ] `author`
 * [ ] `revdate`
 * [ ] `notitle`
 * [ ] `noheader`
 * [ ] `nofooter`
 * [ ] `first_section`
 * [ ] `has_header?`
 * [ ] `<< block`
 * [x] `set_attribute(name, value)`: `setAttribute(name, value)`
 * [ ] `delete_attribute(name)`
 * [ ] `attribute_locked?(name)`
 * [ ] `update_backend_attributes new_backend, force = false`
 * [ ] `update_doctype_attributes new_doctype`
 * [ ] `create_converter`
 * [ ] `convert opts = {}`
 * [ ] `write output, target`
 * [ ] `convert_to target, opts = {}`
 * [ ] `content`
 * [ ] `docinfo location = :head, suffix = nil`
 * [ ] `resolve_docinfo_subs`
 * [ ] `docinfo_processors?(location = :head)`
 * [ ] `to_s`

@mojavelinux Did I miss something or did I forget to remove an internal function ?

Mogztter:
Rebased on master with more API and more tests!

mojavelinux:
@Mogztter There are a number of methods which are intended to be private. (I'll make that declaration more clear once we move to 1.6.0).

Here's a list of the methods you can drop:

#### AbstractNode

* update_attributes(attributes)
* generate_data_uri_from_uri image_uri, cache_uri = false
* generate_data_uri(target_image, asset_dir_key = nil)
* relative_path(filename)
* is_uri? str

#### Document

* register(type, value, force = false)
* first_section
* update_backend_attributes new_backend, force = false
* update_doctype_attributes new_doctype
* create_converter
* convert_to target, opts = {}
* resolve_docinfo_subs
* to_s

The methods for setting and deleting attributes on Document is something that is likely going to undergo some change in the future, but for now we can still map it.

Mogztter:
Thanks, I've updated my todo list :+1:

mojavelinux:
:+1:

mojavelinux:
Btw, if you have any feedback on any of the APIs, don't hesitate to critique them. We want the API to be the best it can be.

Mogztter:
> Btw, if you have any feedback on any of the APIs, don't hesitate to critique them. We want the API to be the best it can be.

I found something confusing in the API.
I've mapped `Document.title()` to `Document.getTitle()` and `Document.title=(title)` to `Document.setTitle(title)`.

```javascript
console.log(doc.getTitle()) // foo
doc.setTitle('bar')
console.log(doc.getTitle()) // foo
```

`Document.title` returns the title explicitly defined in the document attributes but `Document.title=(title)` does not modify the title explicitly defined in the document attributes.

I think `Document.title=(title)` should be renamed to `Document.set_doctitle(title)`

Mogztter:
Also I think `Document.Title.sanitized?` should return a Boolean: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/document.rb#L60

mojavelinux:
I fully admit that the API around the doctitle is weird. The distinction between title and doctitle is not clear in AsciiDoc Python and this was an attempt to blend the old behavior and the new.

I think we need to go back to core and draw up some requirements for these two properties. We'll just have to mark these methods in the Asciidoctor.js API as unstable and subject to change.

Mogztter:
I think we are good to go. @mojavelinux @hsablonniere would you like to do a final review before the merge ?

We obviously need more tests on the API but we can add them gradually. For this purpose, I've shamelessly copy/paste tests from the great AsciidoctorJ tests suite :stuck_out_tongue_closed_eyes:

mojavelinux:
Looks good to me! Let's see how this plays out.

Mogztter:
Let's go! 🍻

mojavelinux:
:beers: Great work! Lots to drink to tonight!

PR #182, opened 2016-04-23
Compile Ruby to js in js

Mogztter:


npm/builder.js
```diff
@@ -481,8 +479,8 @@ Builder.prototype.compile = function() {
 
   log.title('copy resources');
   log.debug('copy asciidoctor.css');
-  // FIXME: Find a more robust implementation to get the Asciidoctor gem install path
-  var asciidoctorGemPath = child_process.execSync("ruby -e \"print (Gem::Specification.find_by_name 'asciidoctor').full_gem_path\"");
+  // REMIND: Asciidoctor is now a Git submodule
+  var asciidoctorGemPath = 'asciidoctor';
```
mojavelinux: I think we should build against a release instead of master. Can we just download the zip from GitHub?

mojavelinux:
Nice work!

I noticed a VIM swap file got caught up in the commit.

Is there a reason set.rb and strscan.rb are now required? Does Opal not make those available?

npm/builder.js
Mogztter: I'm not building against master because the Git submodule points on the 1.5.4 tag but I don't like git submodules.
We will need to figure out how to avoid downloading the zip each time we are building Asciidoctor.js (ie. `build` directory is deleted)

Mogztter:
> I noticed a VIM swap file got caught up in the commit.

:eyes: 

> Is there a reason set.rb and strscan.rb are now required? Does Opal not make those available?

Since we are not using Ruby Gem anymore, Opal builder don't know where to find these two files.
I've decided to put them in the Node package [opal-compiler](https://github.com/Mogztter/opal-node-compiler/tree/master/src/stdlib).
I've tried to add the JavaScript version of these files but in this case Opal compiler is adding the following line `Opal.loaded("set");` which prevent these files from being loaded... See [builder_processors.rb](https://github.com/opal/opal/blob/master/lib/opal/builder_processors.rb#L65)
I don't know the exact reason but for now I'm using Ruby files. Note that `opal-compiler` is only a development dependency.

npm/builder.js
mojavelinux: I agree. git submodules are :hankey:. +1 to downloading the zip.

mojavelinux:
> Since we are not using Ruby Gem anymore, Opal builder don't know where to find these two files.
I've decided to put them in the Node package opal-compiler.

Got it. :+1:

Mogztter:
@mojavelinux Green bar!!! First build without Ruby involved: [Travis](https://travis-ci.org/asciidoctor/asciidoctor.js/builds/126859101) / [Appveyor](https://ci.appveyor.com/project/asciidoctor/asciidoctor-js/build/690) :beers: :tada:

npm/builder.js
Mogztter: @mojavelinux Could we also remove git submodule on Asciidoctor extensions lab ? (ie. publish a version 0.1.0 on GitHub ?).
I think moving each extension into a separate repository will be too much work.

mojavelinux: I agree that moving the extensions is a separate (discrete) effort and not something we want to wait on.

I think we could mark a tag in the repository without releasing a gem. What do you think about 0.0.1 instead or perhaps even a pre-release name, like 1.0.0.preview.1?

mojavelinux:
:green_apple: :green_apple: :green_apple: :green_apple: :green_apple:

:+1:

Mogztter:
I'm still using an integration branch but once https://github.com/opal/opal/pull/1456 is merged and a new release of Opal is available, I will merge this pull request.

mojavelinux:
This truly is a moment to commemorate! \o/

PR #185, opened 2016-05-26
Set environment variable cross-platform compatible

rahmanusta:


mojavelinux:
More context here: https://www.npmjs.com/package/cross-env

Mogztter:
We do need to have a cross environnent compatible build but cross-env has two dependencies which have more dependencies which have... (welcome to the Node.js world).
I think we can find a simpler solution that doesn't require a library.
I will do some testing on Windows to better understand this issue.

One idea is to use program arguments instead of environment variables since we have full control over the build.

Wdyt?

rahmanusta:
> One idea is to use program arguments instead of environment variables since we have full control over the build.

I agree with you, we can use program arguments instead of environment variables with no dependency.

Mogztter:
Merged in 9fdb21c5315718b7facc867229e32e4c23beab24

rahmanusta:
Thank you @Mogztter !

PR #189, opened 2016-06-11
Windows support for Electron

ldez:
Related to https://github.com/electron/electron/issues/2033
Fix https://github.com/asciidoctor/atom-asciidoc-preview/issues/159

I don't know how to write specs for this case.

### DIFF
--- a/lib/asciidoctor/opal_ext.rb
+++ b/lib/asciidoctor/opal_ext.rb
@@ -1,5 +1,6 @@
 %x(
   var isNode = typeof module !== 'undefined' && module.exports,
+      isElectron = typeof process === 'object' && process.versions === 'object' && process.versions.electron === 'string',
       isBrowser = typeof window !== 'undefined',
       isNashorn = typeof Java !== 'undefined' && Java.type,
       isRhino = typeof java !== 'undefined',
@@ -10,7 +11,11 @@
     value = 'browser';
   }
   else if (isNode) {
-    value = 'node';
+    if (isElectron) {
+      value = 'node-electron';
+    } else {
+      value = 'node';
+    }
   }
   else if (isNashorn) {
     value = 'java-nashorn';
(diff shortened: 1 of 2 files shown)


Mogztter:
> I don't know how to write specs for this case.

Indeed this is not a simple task and we will just test our implementation anyway :neutral_face:

lib/asciidoctor/opal_ext.rb
```diff
@@ -1,5 +1,6 @@
 %x(
   var isNode = typeof module !== 'undefined' && module.exports,
+      isElectron = process.versions !== undefined && process.versions.electron !== undefined,
```
Mogztter: maybe we should check that `process` is defined ?

ldez: [process](https://nodejs.org/api/process.html) is always defined with Node

ldez: oops `isXX` are not functions, I must test if `process` is defined.

Mogztter:
@mojavelinux I don't remember if we agreed on a naming to define the runtime environment ?

Electron should be:
```javascript
JAVASCRIPT_ENGINE=electron
JAVASCRIPT_PLATFORM=node // or maybe nodejs ?
```

ldez:
`JAVASCRIPT_ENGINE` is V8 with Node.
Electron is a context perhaps `JAVASCRIPT_CONTEXT`

Mogztter:
Yes you are right, that was part of the [discussion](https://github.com/asciidoctor/asciidoctor/issues/897#issuecomment-63955096) but I think we "formalize" something (the problem is that I can't remember where)

Mogztter:
I think we can track this improvement in another issue

Mogztter:
:+1: Thanks @ldez

PR #190, opened 2016-06-25
Accept WebWorker platform as Browser platform

rahmanusta:
There is no window in WebWorker environment. It supports XmlHTTPRequest or WebWorker function like importScripts

### DIFF
--- a/lib/asciidoctor/opal_ext.rb
+++ b/lib/asciidoctor/opal_ext.rb
@@ -1,7 +1,7 @@
 %x(
   var isNode = typeof module !== 'undefined' && module.exports,
       isElectron = typeof process === 'object' && process.versions === 'object' && process.versions.electron === 'string',
-      isBrowser = typeof window !== 'undefined',
+      isBrowser = typeof window !== 'undefined' || typeof importScripts === 'function',
       isNashorn = typeof Java !== 'undefined' && Java.type,
       isRhino = typeof java !== 'undefined',
       value;


Mogztter:
Thanks @rahmanusta 

@mojavelinux @ldez Just wonder if we should create a dedicated "javascript platform" ?

```javascript
isWebWorker = typeof importScripts === 'function'
```

ldez:
WebWorkers are not limited to browser.
Node is detected before browser, this can produce false-positive (e.g. detect Browser instead of Node)

`isWebWorker` is mandatory.

```js
var isNode = typeof module !== 'undefined' && module.exports,
    isElectron = typeof process === 'object' && process.versions === 'object' && process.versions.electron === 'string',
    isBrowser = typeof window !== 'undefined',
    isWebWorker = typeof importScripts === 'function'
    isNashorn = typeof Java !== 'undefined' && Java.type,
    isRhino = typeof java !== 'undefined',
    value;
// The order of the if statements is important because 'module' will be defined in a Browserify environment
if (isBrowser) {
  value = 'browser';
}
else if (isNode) {
  if (isElectron) {
    value = 'node-electron';
  } else {
    value = 'node';
  }
}
else if (isWebWorker) {
  value = 'browser-webworker';
}
else if (isNashorn) {
  value = 'java-nashorn';
}
else if (isRhino) {
  value = 'java-rhino';
}
else {
  // standalone most likely SpiderMonkey
  value = 'standalone';
}
```

Mogztter:
Ok!
@rahmanusta If you agree with this change could you please update your pull request ? thanks

rahmanusta:
@Mogztter It is okay for me but include:: block macro doesn't work in this case.

Mogztter:
Then this is *not* okay. We need to figure out why this is not working properly :smile: 
Do you think we can add a test case (even with a fake WebWorker environment) to reproduce this issue ?

Mogztter:
I think we should work on a matrix table:

 | Platform | Engine | Framework | I/O
--- | --- | ---  | ---  | ---  | ---
Electron | `node` | `v8` | `electron` | `fs` module
Node.js | `node` | `v8` | | `fs` module
Browser | `browser` | | | `XmlHTTPRequest`
WebWorker Browser | `browser` | | `webworker` | `XmlHTTPRequest`
WebWorker Node | `node` | `v8` | `webworker` | `fs` module
Nashorn | `java` | `nashorn` | | `java/nio/file/Files`
Rhino | `java` | `rhino` | | ???
SpiderMonkey | `standalone` | | | `read(path)`
Webpack | `browser` | | `webpack` | `XmlHTTPRequest`
JSPM | `browser` | | `jspm` | `XmlHTTPRequest`

Mogztter:
@rahmanusta I've improved the detection of the JavaScript environment. Could you please give it a try ? See https://github.com/asciidoctor/asciidoctor.js/pull/248. Do you have a code sample of a WebWorker using Asciidoctor.js ?

Mogztter:
Closed in favor of #248

nerk:
Just to make things even more complicated, I'm trying to figure out how to make this work with the Brackets plugin. This is basically a browser environment with Asciidoctor.js running inside a webworker, using importScripts to load 'asciidoctor.js'. 

No joy so far... Has this actually been tested with a webworker running inside a browser?

Mogztter:
> No joy so far...

What is not working ? Do you have a stack trace ?

> Has this actually been tested with a webworker running inside a browser?

Sort of, I setup a simple project with a WebWorker and it was working. But we do need integration tests because integrating in a simple WebWoeker environment is not the same as integrating in Brackets

nerk:
I need to look at it more closely. What I did was, I invoked 

```
importScripts('asciidoctor.js');
```

at the end of my worker. When the worker's `convert` function is invoked, `Opal` is defined inside this function, but `Opal.Asciidoctor` is undefined. I can access it by using the plain name `Asciidoctor`, but  none of its exported methods, e.g. `load` is defined. I can invoke `Asciidoctor.$load`, but unlike with 1.5.4, this seems to interpret the passed content string as a pathname.

nerk:
Just to make it clear, the `convert` function is my own function called by the worker's `onmessage`

Mogztter:
I can confirm that `Opal.Asciidoctor` is not exposed anymore. `Opal` global variable should not be used in the client code (i.e. Opal is an implementation detail and is now a "private" variable).

More information about this breaking change: https://github.com/asciidoctor/asciidoctor.js/releases/tag/v1.5.5-4

Also, you can use the new wrapper API to remove the dollar sign:

```javascript
importScripts('asciidoctor.js');
Asciidoctor.load("= Hello world", {attributes: 'icons=font@'});
```

Do you have a working branch on GitHub so I can help you migrate to the latest version ?

nerk:
Thanks for pointing me to the release notes. I read manual.adoc, seems to be out-of-date. 

I just created a branch https://github.com/asciidoctor/brackets-asciidoc-preview/tree/1.5.5 but I did not check in any changes. The relevant stuff is in lib/converter_worker.js

I'm kind of busy right now, but when I get the time, I'll check something in.

PR #191, opened 2016-06-27
Export environment variables in a cross-compatible way

Mogztter:
@rahmanusta Could you please tell me if this is working for you (ie. Windows) ?

### DIFF
--- a/npm/arg.js
+++ b/npm/arg.js
@@ -1,0 +1,27 @@
+module.exports = Arg;
+
+function Arg() {
+}
+
+function isEnvironmentVariable(arg) {
+  var envSuffix = 'env.';
+  // startWith 'env.'
+  if (arg.lastIndexOf(envSuffix, 0) === 0) {
+    var equalSignIndex = arg.indexOf('=');
+    var name = arg.substring(envSuffix.length, equalSignIndex);
+    var value = arg.substring(equalSignIndex + 1, arg.length);
+    // Export to environment variable
+    process.env[name] = value;
+    // Filter out this argument from process.argv
+    return false;
+  }
+  return true;
+}
+
+// Export environement variables from command line arguments.
+// Argument starting with 'env.' will be removed from process.argv and added to process.env
+Arg.prototype.exportEnvironmentVariables = function() {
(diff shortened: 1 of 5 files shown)


rahmanusta:
Hey @Mogztter. It works very well. Thanks for this work!

Mogztter:
You're welcome, this is not fully working yet, I will try to fix it this week

Mogztter:
This is harder than I thought, I will accept your pull request

mojavelinux:
Amazing that Node makes this so hard. Seems like something that should be a core feature.

Mogztter:
Indeed, I tried several approaches, but none of them worked. `cross-env` package seems to be the "least worst solution" :wink:

npm/arg.js
```diff
@@ -0,0 +1,27 @@
...
+    // Filter out this argument from process.argv
+    return false;
+  }
+  return true;
+}
+
+// Export environement variables from command line arguments.
+// Argument starting with 'env.' will be removed from process.argv and added to process.env
+Arg.prototype.exportEnvironmentVariables = function() {
+  var args = process.argv.slice(2);
```
mojavelinux: Why are we using slice here instead of considering every argument?

mojavelinux:
I had one question about the implementation. Also, I kind of prefer a better name than Args. Perhaps we can name the module something more explanatory?

npm/arg.js
Mogztter: The first argument is `node` and the second is the program name. `process.argv.slice(2)` gives me the remaining arguments in the command line. Eg:

```javascript
process.argv = ["node", "test.js", "env.MINIFY=1", "env.PRODUCTION=1" ];
process.argv.slice(2) // [ "env.MINIFY=1", "env.PRODUCTION=1" ]
```

Mogztter:
@mojavelinux I've canceled this pull-request and merged https://github.com/asciidoctor/asciidoctor.js/commit/9fdb21c5315718b7facc867229e32e4c23beab24 instead.

mojavelinux:
Oh, my mistake! Thanks for the clarifications!

PR #194, opened 2016-07-17
Upgrade build versions

Mogztter:
Ruby 1.9.3 is no longer maintained and Node 4.4.7 is the current LTS.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 gem 'opal', '0.9.0.beta2'
 #gem 'opal', :github => 'opal'
+gem 'rack', '1.6.4' # Opal compiler is *not* compatible with Rack 2.x
 gem 'asciidoctor', '1.5.4'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 gem 'opal', '0.9.0.beta2'
 #gem 'opal', :github => 'opal'
+gem 'rack', '1.6.4' # Opal compiler is *not* compatible with Rack 2.x
```
Mogztter: @mojavelinux Do you know why suddenly `bundle install` is downloading rack 2.x ? Maybe a "non-strict" transitive dependency somewhere ?

mojavelinux: opal depends on rack-test, which depends on rack >= 1.0. rack 2 was released June 30. So that explains it. (This is why I like depending on concrete versions. Other projects end up deciding when you upgrade.

Mogztter: That's was i thought, thanks for the explanation :smile:

PR #195, opened 2016-07-17
Mutable String methods are not supported in Opal

### LINKED ISSUE #173: Mutable String methods are not supported in Opal.
NotImplementedError : String#sub! not supported. Mutable String methods are not supported in Opal.

Asciidoctor.js Live Preview 1.5.4.100
Chrome Version 48.0.2564.97 (64-bit)

Some AsciiDoc files are reporting this error in the Chrome extension. They were rendering before, so perhaps this issue was introduced in an update?

Mogztter:
Replaces unsupported methods in asciidoctor-core.js during the build phase.
Adds tests to check that asciidoctor-core.js does *not* contains unsupported features in Opal.

Resolves #173

### DIFF
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -91,6 +91,7 @@ Builder.prototype.build = function(callback) {
     function(callback) { builder.clean(callback); }, // clean
     function(callback) { builder.downloadDependencies(callback); }, // download dependencies
     function(callback) { builder.compile(callback); }, // compile
+    function(callback) { builder.replaceUnsupportedFeatures(callback); }, // replace unsupported features
     function(callback) { builder.concatJavaScripts(callback); }, // concat
     function(callback) { builder.uglify(callback); } // uglify (optional)
   ], function() {
@@ -519,6 +520,16 @@ Builder.prototype.compile = function(callback) {
   var asciidoctorPath = 'build/asciidoctor';
   var asciidoctorCSSFile = asciidoctorPath + '/data/stylesheets/asciidoctor-default.css';
   fs.createReadStream(asciidoctorCSSFile).pipe(fs.createWriteStream('build/asciidoctor.css'));
+  callback();
+};
+
+Builder.prototype.replaceUnsupportedFeatures = function(callback) {
+  log.title('Replace unsupported features');
+  var path = 'build/asciidoctor-core.js';
+  var data = fs.readFileSync(path, 'utf8');
+  log.debug('Replace (g)sub! with (g)sub');
+  data = data.replace(/(\(\$\w+ = \(\$\w+ = (\w+)\))\['(\$g?sub)!'\]/g, "$2 = $1['$3']");
+  fs.writeFileSync(path, data, 'utf8');
   callback();
 };
 
--- a/npm/test/unsupported-features.js
+++ b/npm/test/unsupported-features.js
@@ -1,0 +1,11 @@
+var fs = require('fs');
+var Log = require('../log.js');
+var log = new Log();
+
+log.title('Check unsupported features');
+var data = fs.readFileSync('build/asciidoctor-core.js', 'utf8');
+var mutableStringPattern = /\['\$(g)?sub!'\]/;
+if (mutableStringPattern.test(data)) {
+  log.error("Mutable String methods are not supported in Opal, please replace sub! and gsub! methods");
+  process.exit(1);
+}
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "README.adoc"
   ],
   "scripts": {
-    "test": "node npm/test/jasmine-bower.js && node npm/test/jasmine-bower-min.js && node npm/test/jasmine-npm.js && node npm/test/jasmine-bower-latex.js && karma start",
+    "test": "node npm/test/unsupported-features.js && node npm/test/jasmine-bower.js && node npm/test/jasmine-bower-min.js && node npm/test/jasmine-npm.js && node npm/test/jasmine-bower-latex.js && karma start",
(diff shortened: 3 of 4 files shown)


mojavelinux:
I want to look a little closer at this one...mostly so that I make sure I don't forget about the upstream change.

Mogztter:
@mojavelinux Sure no problem, let me know when you are ready to proceed

Mogztter:
@mojavelinux Does this look good to you ? If you want I can create an issue in asciidoctor core repository to track the upstream change ?

mojavelinux:
I'm going to get to this this week. I've been really busy this summer, but starting to get caught up.

Mogztter:
@mojavelinux No worries, a *lot* of things is happening right now!

mojavelinux:
Tell me about it.

npm/builder.js
```diff
@@ -487,6 +490,17 @@ Builder.prototype.compile = function() {
   fs.createReadStream(asciidoctorCSSFile).pipe(fs.createWriteStream('build/asciidoctor.css'));
 };
 
+Builder.prototype.replaceUnsupportedFeatures = function() {
+  log.title('Replace unsupported features');
+  var path = 'build/asciidoctor-core.js';
+  var data = fs.readFileSync(path, 'utf8');
+  log.debug('Replace sub! with sub');
+  data = data.replace(/\['\$sub!'\]/gi, '[\'$sub\']');
```
mojavelinux: The `i` flag is not necessary here since the method name is always lowercase. I also recommend combining these two replacements into a single call to avoid scanning the input twice.

```js
data = data.replace(/(\['\$g?sub)!(?='\])/g, "$1")
```

or a bit less fancy:

```js
data = data.replace(/\['(\$g?sub)!'\]/g, "['$1']")
```

mojavelinux:
Okay, so this is interesting. It turns out that even when Opal converts gsub! and sub!, it still captures and passes on the return value. So simply dropping the exclamation point works. I also like that you have a test to verify that the patch works correctly, so we know right away if that assumption breaks.

I would still like to see Opal perform an automatic translation of string mutator methods. Clearly, Opal is almost there because it's already capturing the return value and passing that to the next step. All that's left is to drop the `!`. I think this would allow Opal to handle more libraries without modification.

mojavelinux:
Actually, I was incorrect. Opal does not reassign the return value. It just so happens that in our logic, if the first tab replacement doesn't stick (as it does not), it just handles the tab replacement again in the condition...which doesn't need to be reassigned because it becomes the method's return value. In other words, Asciidoctor.js does extra work to replace tabs.

npm/builder.js
mojavelinux: So the regexp needs to be a bit more complex in order to capture the return value.

The following pattern:

```
($d = ($e = result)['$gsub!']
```

must become

```
result = ($d = ($e = result)['$gsub!']
```

which gives us:

```js
data.replace(/(\(\$\w+ = \(\$\w+ = (\w+)\))\['(\$g?sub)!'\]/g, "$2 = $1['$3']")
```

mojavelinux: I tested this replacement and it does exactly what we want and doesn't seem to cause any issues.

mojavelinux:
I provided the correct replacement in the line comment. Once you apply that, I'm prepared to merge this change.

I'm still trying to decide what to do in core. One thing I could do is always perform the assignment even if I use `g?sub!`. That way, all we would need to do in the Asciidoctor.js build in the future is alias `g?sub!` to `g?sub`. We'd be safe knowing that it will return the value anyway...so it's nothing more than an optimization.

mojavelinux:
Btw, the reason I use `g?sub!` is because it avoids creating a new string internally in Ruby. If we assign the result to the same variable, that still avoids allocation.

mojavelinux:
Shoot, that doesn't work because if the replacement doesn't happen, Ruby returns nil. So the two methods don't have the same behavior.

mojavelinux:
Well, either way, our patch will work. I'll figure out a way to make it obsolete eventually ;)

Mogztter:
Thanks for the full detailed explanation. I don't quite follow you on your last comment:

> Shoot, that doesn't work because if the replacement doesn't happen, Ruby returns nil. So the two methods don't have the same behavior.

You are saying that `sub!` and `gsub!` don't have the same behavior ? So could we still use the following replacement ?

```javascript
data = data.replace(/(\(\$\w+ = \(\$\w+ = (\w+)\))\['(\$g?sub)!'\]/g, "$2 = $1['$3']")
```

mojavelinux:
> You are saying that sub! and gsub! don't have the same behavior ?

I'm saying that gsub and gsub! don't have the same return value. gsub always returns the input string, whether or not it was modified. gsub! only returns the input string if it was modified. (I was mostly just talking to myself about how I might write it in core. It doesn't affect what we are doing here.).

mojavelinux:
>  So could we still use the following replacement ?

That's correct.

mojavelinux:
To clarify what we need to do in Asciidoctor.js, we need to not only drop the exclamation point, we also need to assign the return value to the variable on which the method is run.

Mogztter:
> I'm saying that gsub and gsub! don't have the same return value. gsub always returns the input string, whether or not it was modified. gsub! only returns the input string if it was modified. (I was mostly just talking to myself about how I might write it in core. It doesn't affect what we are doing here.).

Ok thank you for clearing that up.

Mogztter:
@mojavelinux I made the changes, does this look good now ?

mojavelinux:
Looks great!

PR #196, opened 2016-07-20
Compile asciidoctor latex

### LINKED ISSUE #192: Asciidoctor-latex
Implement the asciidoctor-latex extension in asciidoctor.js.  I've implemented an alternative to using Ruby's prepend, so this should now be possible.  See [L571](https://github.com/asciidoctor/asciidoctor-latex/blob/master/lib/asciidoctor/latex/converter.rb#L571) of the converter code.  Code quality of `asciidoctor-latex` may need to be improved for this.

Reason: I would like be able to render asciidoc-latex documents in the client so as to avoid hitting the server so often for [scripta.io](http:www.scripta.io) and  [shoobox.io](http:www.shoobox.io). This will also set the stage for a better editor on these sites with a javascript frontend.

Mogztter:
Based on #182 

Resolves #192

### DIFF
--- a/lib/asciidoctor/opal_ext.rb
+++ b/lib/asciidoctor/opal_ext.rb
@@ -33,6 +33,7 @@ require 'strscan'
 require 'asciidoctor/opal_ext/file'
 require 'asciidoctor/opal_ext/match_data'
 require 'asciidoctor/opal_ext/kernel'
+require 'asciidoctor/opal_ext/thread_safe'
 
 case JAVASCRIPT_PLATFORM
   when 'java-nashorn'
--- a/lib/asciidoctor/opal_ext/thread_safe.rb
+++ b/lib/asciidoctor/opal_ext/thread_safe.rb
@@ -1,0 +1,5 @@
+module ThreadSafe
+end
+
+class Cache < ::Hash
+end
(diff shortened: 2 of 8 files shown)


Mogztter:
@jxxcarlson Compilation is working fine but you need to make some changes in asciidoctor-latex:

 * [ ] Opal does not define a `__dir__` method, you need to replace [this line in latex.rb#L3](https://github.com/asciidoctor/asciidoctor-latex/blob/master/lib/asciidoctor/latex.rb#L3) with [something like asciidoctor.rb#L172](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L172)

mojavelinux:
Wow! Pretty cool!

I'll follow up on the ThreadSafe issue. One workaround is to define ThreadSafe::Cache by mapping it to Hash in opal/ext.

jxxcarlson:
Thanks so much!  Will get this done in the next day or two.

Mogztter:
> One workaround is to define `ThreadSafe::Cache` by mapping it to Hash in opal/ext.

Good idea thanks!

mojavelinux:
In fact, I'm almost wondering if that is the solution for asciidoctor/asciidoctor#1820. Either way, I will follow-up there.

Mogztter:
In my opinion, your solution is simple enough to not make a special case in Asciidoctor core

mojavelinux:
:+1:

PR #199, opened 2016-08-13
Asciidoctor.load_file method is now working in Node.js

Mogztter:
`File.mtime` is available on Opal 0.10+ and `File.readlines` is implemented in `opal_ext`.

Note: Test is failing because I need to fix a small issue in Opal: https://github.com/opal/opal/issues/1550

### DIFF
--- a/lib/asciidoctor/opal_ext/file.rb
+++ b/lib/asciidoctor/opal_ext/file.rb
@@ -68,7 +68,7 @@ class File
     end
   end
 
-  def readlines()
+  def readlines
     File.readlines(@path)
   end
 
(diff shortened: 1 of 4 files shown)


lib/asciidoctor/opal_ext/file.rb
```diff
@@ -68,6 +68,15 @@ def each_line(separator = $/, &block)
     end
   end
 
+  def readlines()
```
mojavelinux: It's generally accepted in the Ruby community to drop the brackets for methods that have no arguments.

Mogztter: Done, thanks for the code review

PR #209, opened 2016-08-19
Adds tests on convert_file method running with Node.js

Mogztter:
@cmoulliard I've added tests to make sure that Asciidoctor.js can read/write files when running with Node.js. Could you please do a quick review ? Did I missed a use case ?


spec/npm/asciidoctor-all.spec.js
```diff
@@ -21,5 +37,76 @@ describe('Node.js', function () {
       expect(doc.attributes.$fetch('docname')).toBe('test.adoc');
     });
   });
+
+  describe('Converting file', function() {
+    it('should be able to convert a file', function() {
+      var expectFilePath = __dirname + '/test.adoc.html';
```
Mogztter: @mojavelinux I'm surprised with the file name, it should be `test.html` no ?

spec/npm/asciidoctor-all.spec.js
```diff
@@ -21,5 +37,76 @@ describe('Node.js', function () {
...
+      removeFile(expectFilePath);
+      try {
+        var doc = Asciidoctor.$convert_file(__dirname + '/test.adoc', null);
```
Mogztter: @mojavelinux Even without `to_file` the returned value is a `Document` and a file is written `test.adoc.html`. 
https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L1575-L1576

In my case I think `to_file` is `nil`: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L1424
Is that the correct behavior ?

spec/npm/asciidoctor-all.spec.js
```diff
@@ -21,5 +37,76 @@ describe('Node.js', function () {
...
+      } finally {
+        removeFile(expectFilePath);
+      }
+    });
+
+    it('should be able to convert a file with custom css embedded', function() {
+      var expectFilePath = __dirname + '/test.adoc.html';
+      removeFile(expectFilePath);
+      try {
+        var options = Opal.hash({'safe': 'server', 'attributes': ['stylesheet=simple.css', 'stylesdir=css']});
```
Mogztter: @mojavelinux With `'safe': 'server'`, the stylesheet is automatically embedded in the HTML document, is that intended ?

mojavelinux: To be more accurate, the default is to link to the CSS if safe mode is secure and embed the CSS if safe mode is server or below. When safe mode is server or below, you can still control this behavior using the linkcss attribute.

spec/npm/asciidoctor-all.spec.js
mojavelinux: Actually, docname should be extensionless. It seems like the version from lib/asciidoctor/opal_ext/file.rb does not accept the second argument. Fortunately, this method is now implemented in Opal core, so we can either remove this patch or we need to implement the second argument.

See https://github.com/asciidoctor/asciidoctor/blob/v1.5.4/lib/asciidoctor/helpers.rb#L179-L196

spec/npm/asciidoctor-all.spec.js
mojavelinux: > Even without to_file the returned value is a Document and a file is written test.adoc.html. 

This is due to the File.basename problem I mentioned above.

mojavelinux: There are a lot of different permutations here. You need to study the logic in the link below to understand the behavior:

https://github.com/asciidoctor/asciidoctor/blob/v1.5.4/lib/asciidoctor.rb#L1421-L1438

cmoulliard:
@Mogztter. The tests coverage is sufficient for the moment. We will most probably extend them in the future. Can we apply this PR and generate a new release please ?

spec/npm/asciidoctor-all.spec.js
Mogztter: Nice catch! I've removed all methods that are now implemented in Opal core.

spec/npm/asciidoctor-all.spec.js
Mogztter: Thanks for the clarification :smile:

Mogztter:
:+1:

mojavelinux:
:+1:

Mogztter:
Done, version `1.5.5-2` h @cmoulliard

cmoulliard:
Tested successfully this morning ! I will have a look now to see if we can generate the content of the asciidoctor files located within a folder.

PR #224, opened 2016-09-16
Automatically load Opal's nodejs and pathname module

Mogztter:
Avoid the necessity to manually load those modules when running on Node.js.
Currently when an user want to use the `convert_file` method, it's mandatory to load `nodejs` module:

The following code will throw an exception `mtime: undefined method 'mtime' for #<File:0x1b6>`:
```javascript
var asciidoctor = require('asciidoctor.js')();

var opal = asciidoctor.Opal;
var processor = asciidoctor.Asciidoctor();

//opal.load('nodejs');

var doc = processor.$convert_file('test.adoc');
console.log(doc.$doctitle());
```

### DIFF
--- a/lib/asciidoctor/opal_ext.rb
+++ b/lib/asciidoctor/opal_ext.rb
@@ -16,6 +16,8 @@
     } else {
       value = 'node';
     }
+    Opal.load("nodejs");
+    Opal.load("pathname");
   }
   else if (isNashorn) {
     value = 'java-nashorn';


mojavelinux:
One more step that could simplify the API is by allowing the convert methods to accept a raw JSON for the options argument instead of having to wrap it in Opal.hash(). (We could add something to core if necessary). Another solution would be to add an OptionsBuilder like in AsciidoctorJ that abstracts away Opal.

Mogztter:
That will be the case when https://github.com/asciidoctor/asciidoctor.js/pull/180 is merged :wink:

mojavelinux:
+1

PR #227, opened 2016-09-19
Resolves #221, improve the path resolution of the default stylesheet

### LINKED ISSUE #221: Default asciidoctor-default.css file not found
If I use asciidoctor.js to render an adoc file to HTML then I get this error : 

```
Error: ENOENT: no such file or directory, open 'data/stylesheets/asciidoctor-default.css'
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.readFileSync (fs.js:397:15)
    at /Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:872:118
    at ːread (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:903:84)
    at ːread (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:915:33)
    at $Stylesheets.ːprimary_stylesheet_data (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:11152:146)
    at $Stylesheets.ːembed_primary_stylesheet (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:11158:36)
    at $Html5Converter.ːdocument (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:4684:44)
    at $Html5Converter.ː__send__ (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/node_modules/opal-runtime/src/opal.js:3228:21)
    at $Html5Converter.ːconvert (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:4167:24)
    at $Document.ːconvert (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:7110:35)
    at module_constructor.ːconvert (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:12411:22)
    at module_constructor.ːconvert_file (/Users/chmoulli/MyProjects/litoria/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.js:12473:21)
```

* This problem is related to the fact that the packaging of the gem is different from nodejs where the default css file is saved into the folder `dist/css` under the name `asciidoctor.css`
* Asciidoctor.js calculates a relative path that it can't retrieve.

I have also observed that Asciidoctor Ruby when it generates the HTML adds the link of the font within the header and this is not the case for asciidoctor.js

```
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="UTF-8">
<!--[if IE]><meta http-equiv="X-UA-Compatible" content="IE=edge"><![endif]-->
<meta name="viewport" content="width=device-width, initial-scale=1.0">
<meta name="generator" content="Asciidoctor 1.5.4">
<title>My Doc</title>
<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,300italic,400,400italic,600,600italic%7CNoto+Serif:400,400italic,700,700italic%7CDroid+Sans+Mono:400,700">
<style>
/* Asciidoctor default stylesheet | MIT License | http://asciidoctor.org */
/* Remove comment around @import statement below when using as a custom stylesheet */
/*@import "https://fonts.googleapis.com/css?family=Open+Sans:300,300italic,400,400italic,600,600italic%7CNoto+Serif:400,400italic,700,700italic%7CDroid+Sans+Mono:400,700";*/
article,aside,details,figcaption,figure,footer,header,hgroup,main,nav,section,summary{display:block}
```

Mogztter:
Asciidoctor Ruby is trying to read the default stylesheet at `data/stylesheets/asciidoctor-default.css`.
In Asciidoctor.js the default stylesheet is available at `${baseDir}/css/asciidoctor.css`.

When running on Node.js, use `__dirname + '../css/asciidoctor.css'`.
When running on other environments, use `'css/asciidoctor.css'`.

Also remove `examples/asciidoctor.css` to avoid confusion.

Resolves #221


lib/asciidoctor/stylesheets.rb
```diff
@@ -0,0 +1,106 @@
+# encoding: UTF-8
+# Monkey patch the whole file because the class is static
```
Mogztter: @mojavelinux Do you think we can do something in Asciidoctor core to make it easier to override ?

mojavelinux: Instead of replacing the entire file, can't we just redefine the constants? (I suppose to avoid any sort of warning message, we could use a global variable instead of a constant in core).

mojavelinux: Or you could just monkeypatch the `primary_stylesheet_data` method. Replacing the whole file is going to be much harder to maintain.

Mogztter: I've tried both but this not working. I think the reason is that the class is a singleton (class is instantiated statically).
If I monkeypatch `primary_stylesheey_data` and `self.instance` then my code is executed but Opal does not find other methods.
Does that make sense ? Did I miss something ?

mojavelinux: Hmm. The plot thickens.

When I used the following, and compiled and ran it with Opal, it worked:

```ruby
require 'asciidoctor'

module Asciidoctor
class Stylesheets
  def primary_stylesheet_data
    @primary_stylesheet_data ||= 'foobar'
  end
end
end

puts Asciidoctor::Stylesheets.instance.primary_stylesheet_data
```

Note that with this approach, we need not worry about the path constants because we just have to return the data.

mojavelinux: I compiled it using:

```
opal -I /path/to/asciidoctor.js/lib -g asciidoctor test.rb
```

mojavelinux: Keep in mind it has to be an extra script not at the same path as the original file. In other words, it needs to be appended to the built-in script. So put it at the path lib/asciidoctor/ext/stylesheets.rb or something.

mojavelinux: One way to do it is to create the following file:

*lib/asciidoctor/ext/stylesheets.rb*
```ruby
module Asciidoctor
class Stylesheets
  def primary_stylesheet_data
    @primary_stylesheet_data ||= 'foobar'
  end 
end
end
```

Then require this path in *lib/asciidoctor/opal_ext.rb* (hijacking our opportunity to require additional scripts in core).

```ruby
module Asciidoctor; DATA_PATH = ''; end
require 'asciidoctor/stylesheets'
require 'asciidoctor/ext/stylesheets'
```

The challenge is that opal_ext gets required before the Asciidoctor module is defined, so we have to predefine the constant. If you have a way to add an extra require after requiring asciidoctor, use that instead.

The key point is that a monkeypatch needs to augment, not replace, the thing it's monkeypatching.

Of course, another solution to the problem is to literally patch the JS file as we do for the gsub stuff. The could be a cleaner approach.

Mogztter: @mojavelinux  Wow thanks for taking the time to dig into this issue!

> Of course, another solution to the problem is to literally patch the JS file as we do for the gsub stuff. The could be a cleaner approach.

Yeah, I will patch the JS file, that's the less intrusive solution and maybe the most comprehensive.

mojavelinux: :+1:

Issue #229, opened 2016-09-20
Extract reusable build modules to their own repository


Mogztter:
* [ ] Download dependencies, files...
* [x] Opal compiler
* [x] Log
* [x] Uglify (based on Google Closure Compiler)
* [x] FS (concat, mkdirs, untar, recursive delete...)
* [x] JDK download

This should significantly reduce the size of `builder.js`.

mojavelinux:
I like it!

I'm curious, what's the origin of the word bestikk?

Mogztter:
Bestikk is Norwegian for Cutlery.
I kinda like the idea of a set of "simple" tools and Bestikk sounds great :wink:

mojavelinux:
I love it!

PR #231, opened 2016-09-25
Extract logging functions into a dedicated module

Mogztter:
Logging functions are now part of https://github.com/bestikk/bestikk-log

Ref : #229


mojavelinux:
Nice refactoring!

PR #236, opened 2016-10-03
Build against master

Mogztter:
@mojavelinux 

```
1) Bower Loading should load document attributes
  Message:
    bytes: undefined method `bytes' for #<Encoding:UTF-16LE>
  Stack:
    bytes: undefined method `bytes' for #<Encoding:UTF-16LE>
        at Opal.defs.TMP_1 [as $new] (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:4853:15)
        at ːmethod_missing (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:3410:54)
        at method_missing_stub [as $bytes] (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:1006:35)
        at ːbytesize (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:15983:19)
        at String.ːbytesize (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:16020:28)
        at String.ːlimit (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:20765:38)
        at $Document.ːset_attribute (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:26435:72)
        at /home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:28613:24
        at ːstore_attribute (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:28616:28)
        at ːprocess_attribute_entry (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:28578:16)
```
https://github.com/asciidoctor/asciidoctor/blob/93063e2500369612e923c04695fbf6e3054b31ba/lib/asciidoctor/core_ext/string/limit.rb#L5

I don't know why this error is thrown because Opal seems to implement these methods: 

 * https://github.com/opal/opal/blob/0cf1870a66c7ff338e33ede1a1941f70ff357828/opal/corelib/string/encoding.rb#L101-L116
 * https://github.com/opal/opal/blob/0cf1870a66c7ff338e33ede1a1941f70ff357828/opal/corelib/string/encoding.rb#L152-L154

Maybe since we are "patching" String class, Opal doesn't find `bytes` method anymore ?

### DIFF
--- a/lib/asciidoctor/opal_ext/string.rb
+++ b/lib/asciidoctor/opal_ext/string.rb
@@ -1,0 +1,8 @@
+class String
+  # Safely truncate the string to the specified number of bytes.
+  def limit size
+    return self.to_s unless size < bytes.length
+    result = byteslice 0, size
+    result.to_s
+  end unless method_defined? :limit
+end
--- a/lib/asciidoctor/opal_ext.rb
+++ b/lib/asciidoctor/opal_ext.rb
@@ -36,6 +36,7 @@ require 'asciidoctor/opal_ext/file'
 require 'asciidoctor/opal_ext/match_data'
 require 'asciidoctor/opal_ext/kernel'
 require 'asciidoctor/opal_ext/thread_safe'
+require 'asciidoctor/opal_ext/string'
 
 case JAVASCRIPT_PLATFORM
   when 'java-nashorn'
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -27,7 +27,6 @@ function Builder() {
     'src/npm/prepend-core.js',
     'build/asciidoctor-core.js'
   ];
-  this.asciidoctorCoreVersion = '1.5.4';
   this.asciidoctorLatexVersion = '0.2';
   this.htmlEntitiesVersion = '4.3.3';
   this.benchmarkBuildDir = 'build' + path.sep + 'benchmark';
@@ -71,7 +70,7 @@ Builder.prototype.downloadDependencies = function(callback) {
 
   var builder = this;
   async.series([
-    function(callback) { builder.getContentFromURL('https://codeload.github.com/asciidoctor/asciidoctor/tar.gz/v' + builder.asciidoctorCoreVersion, 'build/asciidoctor.tar.gz', callback); },
+    function(callback) { builder.getContentFromURL('https://codeload.github.com/asciidoctor/asciidoctor/tar.gz/master', 'build/asciidoctor.tar.gz', callback); },
     function(callback) { builder.getContentFromURL('https://codeload.github.com/asciidoctor/asciidoctor-latex/tar.gz/' + builder.asciidoctorLatexVersion, 'build/asciidoctor-latex.tar.gz', callback); },
     function(callback) { builder.getContentFromURL('https://codeload.github.com/threedaymonk/htmlentities/tar.gz/v' + builder.htmlEntitiesVersion, 'build/htmlentities.tar.gz', callback); },
     function(callback) { bfs.untar('build/asciidoctor.tar.gz', 'asciidoctor', 'build', callback); },
--- a/spec/share/latex-specs.js
+++ b/spec/share/latex-specs.js
@@ -16,14 +16,16 @@ var commonSpec = function(testOptions, Opal, Asciidoctor) {
       it('should render asciidoctor-latex question/answer block', function() {
         var html = Asciidoctor.$convert('[env.question]\n--\nWhat is the speed of light?\n--\n\n[click.answer]\n--\n300,000 km/sec\n--', null);
         expect(html).toContain('<div class="openblock question">\n' +
-                               '<div class="title">Question 1.</div><div class="content">\n' + 
+                               '<div class="title">Question 1.</div>\n' +
+                               '<div class="content">\n' +
                                '<div class=\'click_oblique\'>\n' +
                                'What is the speed of light?\n' +
                                '</div>\n' +
                                '</div>\n' +
                                '</div>\n' +
                                '<div class="openblock click">\n' +
-                               '<div class="title">Answer</div><div class="content">\n' +
+                               '<div class="title">Answer</div>\n' +
+                               '<div class="content">\n' +
                                '<div class=\'click_oblique\'>\n' +
                                '300,000 km/sec\n' +
                                '</div>\n' +


mojavelinux:
This appears to be a bug in Opal. Consider the following script:

```ruby
puts 'hi'.bytesize
```

Run it using:

```
opal test.rb
```

Same exception. It seems like Opal is mixing up the call chain despite the fact that all methods are defined.

It's also a bit strange that Opal defaults to UTF-16LE strings and not UTF-8. I don't know what that's all about.

mojavelinux:
One way to fix this in the interim is just to define bytesize on String in core_ext.

```ruby
def bytesize
  bytes.length
end
```

However, since the default encoding is UTF-16LE for some reason, the return value is twice the length it should be. A quick fix would be to force the encoding:

```ruby
def bytesize
  (force_encoding ::Encoding::UTF_8).bytes.length
end
```

meh:
@mojavelinux in JavaScript **all** strings are UTF-16, the `#bytesize` result is correct based on that, there's no such thing as a string containing binary data in JavaScript other than abusing properties of the engine.

meh:
Fixed on https://github.com/opal/opal/commit/a355b94e31338ee13d1eb6cff95dd78fc7221823

mojavelinux:
@meh Thanks for the clarification.

Mogztter:
For now, I will go with your quick fix:

```ruby
# Temporary fix until opal/opal@a355b94 is available in a release
def bytesize
  (force_encoding ::Encoding::UTF_8).bytes.length
end
```

mojavelinux:
:+1:

Mogztter:
Hmmm I think I missed something: `no implicit conversion of #<Class:0xae> into String`
`0xae` is a copyright symbol in ANSI Hex. ®

@mojavelinux Any idea?

mojavelinux:
:confused:

Mogztter:
If I remove the `force_encoding` this is working but...
```ruby
class String
  # Temporary fix until opal/opal@a355b94 is available in a release
  def bytesize
    bytes.length
  end
end
``` 
... tests are failing:
```
Failures:
1) Bower Loading should load document attributes
  Message:
    Expected attribute-value to be 'attribute-value'.
  Stack:
    Error: Expected attribute-value to be 'attribute-value'.
        at Object.<anonymous> (/home/guillaume/workspace/opensource/asciidoctor.js/build/bower.spec.all.js:35336:44)
```

```javascript
var options = Opal.hash({'attributes': 'icons=font@ data-uri!'});
var doc = Asciidoctor.$load('= Document Title\n:attribute-key: attribute-value\n\ncontent', options);
expect(doc.$attr('attribute-key')).toBe('attribute-value');
```

`doc.$attr('attribute-key')` does not return a `string` type anymore but an `object`:

```javascript
console.log(doc.$attr('attribute-key')); // [String: 'attribute-value']
console.log(typeof doc.$attr('attribute-key')); // object
```

With the wrapper API, I can "cast" object as string but I'm a little bit worried that this is hiding something bigger :unamused:

meh:
That happens because in JavaScript literals and objects behave differently.

```
> typeof("foo")
'string'
> typeof(new String("foo"))
'object'
> "foo" == new String("foo")
true
```

Since literals are immutable, the only way to change a property is to create the corresponding `Object`, in this case `#force_encoding` creates a new `String` object and sets the `encoding` property on it, so it can be handled.

What must be happening is that you're returning that object somewhere, instead of just using the `#force_encoding` for the `#bytesize` call.

Mogztter:
> What must be happening is that you're returning that object somewhere, instead of just using the #force_encoding for the #bytesize call.

You put me on track!
What is happening is that we are now [using a limit method on String](https://github.com/asciidoctor/asciidoctor/blob/e2ed959d484de5543c6cf71be3102d8fd4763745/lib/asciidoctor/document.rb#L838):

```ruby
if @max_attribute_value_size
  resolved_value = (apply_attribute_value_subs value).limit @max_attribute_value_size
else
  resolved_value = apply_attribute_value_subs value
end
```

In the code above, `value` is a literal:
```javascript
if ((($a = self.max_attribute_value_size) !== nil && $a != null && (!$a.$$is_boolean || $a == true))) {
  console.log('typeof value', typeof value); // string
  resolved_value = (self.$apply_attribute_value_subs(value)).$limit(self.max_attribute_value_size)
} else {
  resolved_value = self.$apply_attribute_value_subs(value)
};
```

But in the [`limit` method we are returning "self"](https://github.com/asciidoctor/asciidoctor/blob/e2ed959d484de5543c6cf71be3102d8fd4763745/lib/asciidoctor/document.rb#L838):

```ruby
class String
  # Safely truncate the string to the specified number of bytes.
  # If a multibyte char gets split, the dangling fragment is removed.
  def limit size
    return self unless size < bytesize
    # NOTE JRuby 1.7 & Rubinius fail to detect invalid encoding unless encoding is forced; impact is marginal.
    size -= 1 until ((result = byteslice 0, size).force_encoding ::Encoding::UTF_8).valid_encoding?
    result
  end unless method_defined? :limit
end
```


and "self" is an "object" :disappointed: 

```javascript
return (Opal.defn(self, '$limit', TMP_1 = function ːlimit(size) {
  var $a, $b, self = this, result = nil;

  console.log(typeof self); // object
  // ...
}
```

I guess we could call `toString()` on the wrapper API if Asciidoctor returns a `String` object: 

```
> new String("foo").toString();
"foo"
> typeof new String("yo").toString();
"string"
```

mojavelinux:
I'm glad you are making headway. Instead of overriding bytesize in Opal, you may just want to override limit. It's doing some complicated stuff that perhaps isn't even necessary in JavaScript...so that could simplify things.

Mogztter:
Green bars!

lib/asciidoctor/opal_ext/string.rb
```diff
@@ -0,0 +1,8 @@
+class String
+  # Safely truncate the string to the specified number of bytes.
+  def limit size
+    return self.to_s unless size < bytes.length
```
Mogztter: Direct call to `bytes.length` instead of `bytesize`

lib/asciidoctor/opal_ext/string.rb
```diff
@@ -0,0 +1,8 @@
...
+    result = byteslice 0, size
```
Mogztter: I removed the shenanigans :wink:

meh:
One of the huge semantical pitfalls in JavaScript is that if you implement a method on a literal, for example `String`, the `this` inside the method is implicitly converted to a `String` object, so if you do return that `this`, you're effectively returning an `Object` instead of the literal.

One way to avoid that is calling `.valueOf()` on the method and returning that. This isn't done automatically or internally within corelib methods because it would slow everything down when most of the time it's going to work.

Mogztter:
@meh Thanks for the explanation. I'm using `to_s` in Ruby to return the literal value, that should be equivalent to `valueOf()` ?

meh:
@Mogztter yes, it's equivalent.

mojavelinux:
> ... One way to avoid that is calling .valueOf() on the method and returning that. This isn't done automatically or internally within corelib methods because it would slow everything down when most of the time it's going to work.

I think that's perfectly reasonable. Just something we need to keep in mind when overriding a built-in type like String.

PR #243, opened 2016-10-06
Include extensions API in asciidoctor-core.js 

Mogztter:
* Rename asciidoctor-all.js to asciidoctor.js
 * asciidoctor-core.js contains core + extensions API
 * Rename specs.js to spec.js
 * Update package.json and bower.json accordingly
 * Update documentation


Mogztter:
This is the first toward a more consistent experience when using Asciidoctor:

**Browser**

```html
<script src="asciidoctor.min.js"></script>
```
```javascript
Opal.Asciidoctor.convert("*This is awesome!*");
```

**Server**

```
npm install asciidoctor.js
```
```javascript
var Asciidoctor = require('asciidoctor.js')().Asciidoctor();
Asciidoctor.convert("*This is awesome!*");
```

I can go one step further and break the API:

 ```javascript
var Asciidoctor = require('asciidoctor.js')().Asciidoctor; // Asciidoctor is no longer a function but an object `Opal.Asciidoctor`
Asciidoctor.convert("*This is awesome!*");
```

I don't know if it's a good idea to auto-configure everything in order to remove the instanciation. I think 
we will still need to "configure" the Asciidoctor modules somehow (ie. choose to require `Opal` and/or `xmlhttprequest`) so I'm not sure we will be able to do:

```javascript
var Asciidoctor = require('asciidoctor.js');
Asciidoctor.convert("*This is awesome!*");
```

Currently the double instanciation is weird, maybe the following is the best we can do:

```javascript
var Asciidoctor = require('asciidoctor.js')(/* optional config */);
Asciidoctor.convert("*This is awesome!*");
```

Mogztter:
> I don't know if it's a good idea to auto-configure everything in order to remove the instanciation. I think 
we will still need to "configure" the Asciidoctor modules somehow (ie. choose to require Opal and/or xmlhttprequest) so I'm not sure we will be able to do:

Actually I've found a way to auto-configure Asciidoctor.js, so it's possible to require the Node module as follows:

```javascript
var Asciidoctor = require('asciidoctor.js');
Asciidoctor.convert("*This is awesome!*");
```

See #248

Mogztter:
@mojavelinux @anthonny Could you please review this pull request ?

Issue #178, opened 2016-03-31
Does not work with module loader for browser


kuechlerm:
When using JSPM or Webpack, requiring / importing asciidoctor.js requires the node module "fs".
This does not work, because I want to use it in a browser environment.
When my code runs, the right path (caused by JAVASCRIPT_PLATFORM) is used. But for reasons unknown both JSPM and Webpack tell me that they cannot find the module "fs".

Can you provide a package with browser-only code and please please publish it on npm?
Or is already something out there? Or a setting I miss? 
Thanks.

anthonny:
Hi @kuechlerm 

I have the same problem with HubPress and i'm working on it.
I use babel/browserify and with this tools JAVASCRIPT_PLATFORM is considered as 'node' instead of 'browser'.
This is because in this kind of "compilation" the test `(typeof module !== 'undefined' && module.exports)` return `true` 
https://github.com/asciidoctor/asciidoctor.js/blob/master/lib/asciidoctor/opal_ext.rb#L3

I have changed the test: 
https://github.com/anthonny/asciidoctor.js/blob/master/lib/asciidoctor/opal_ext.rb#L3

Now the JAVASCRIPT_PLATFORM is correct but i still have an issue with `include::[]`: 
`TypeError: Cannot read property 'write' of undefined at $IO.write_proc`

But i'm sure that @Mogztter or @mojavelinux will help us ;)

kuechlerm:
:+1: @anthonny

Mogztter:
Interesting.
@anthonny Could you please open a pull request with your fix on `opal_ext` ?

> Now the JAVASCRIPT_PLATFORM is correct but i still have an issue with include::[]: 
TypeError: Cannot read property 'write' of undefined at $IO.write_proc

Do you have a simple test to reproduce the problem ? 

Thanks

anthonny:
Yes i will open a PR i must rebase before.
I do not have a simple test, i encounter this error during the development of HubPress
But i suppose that i can create a little project to reproduce.

anthonny:
The PR is here https://github.com/asciidoctor/asciidoctor.js/pull/179

anthonny:
@kuechlerm we are working on this problem, we are waiting for the 0.9.2 in the opal-cdn project.
You can already use the fixed version and make some tests by using my repository:
`"asciidoctor.js": "github:anthonny/asciidoctor.js#fix-opal"`

kuechlerm:
[...]

PR #248, opened 2016-10-09
Auto-configure Asciidoctor.js

Mogztter:
* Choose the right IO module
 * Detect the JavaScript platform, engine and framework
 * Auto-configure Asciidoctor.js node module

### DIFF
--- a/src/npm/prepend-core.js
+++ b/src/npm/prepend-core.js
@@ -1,11 +1,6 @@
-module.exports = function(opalParam, xmlHttpRequestParam) {
-  var Opal = opalParam || require('opal-runtime').Opal;
-
-  if (xmlHttpRequestParam !== false) {
-    var XMLHttpRequest = xmlHttpRequestParam || require('xmlhttprequest').XMLHttpRequest;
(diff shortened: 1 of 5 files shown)


src/npm/prepend-core.js
```diff
@@ -1,11 +1,6 @@
-module.exports = function(opalParam, xmlHttpRequestParam) {
-  var Opal = opalParam || require('opal-runtime').Opal;
-
-  if (xmlHttpRequestParam !== false) {
-    var XMLHttpRequest = xmlHttpRequestParam || require('xmlhttprequest').XMLHttpRequest;
-    // Define overrideMimeType, not define by default in wrapper
-    XMLHttpRequest.prototype.overrideMimeType = function() {};
```
Mogztter: I think this is no longer needed because `XMLHttpRequest` is only used when running on browser.
We might need to improve the detection of `XMLHttpRequest` because according to @anthonny, `XMLHttpRequest` is only available on the `window` object (when using module loader like Browserify or Webpack): https://github.com/asciidoctor/asciidoctor.js/issues/178#issuecomment-205062716

PR #249, opened 2016-10-23
Resolves #242, move asciidoctor-latex.js to asciidoctor-latex repository

### LINKED ISSUE #242: Move asciidoctor-latex.js to asciidoctor-latex repository

Mogztter:
Merged once https://github.com/asciidoctor/asciidoctor-latex/pull/48 is merged.


Mogztter:
Moving forward, `asciidoctor-latex.js` must be generated from the Asciidoctor Latex repository

mojavelinux:
This is definitely the right approach (if not a dedicated repository). We don't want to couple the Asciidoctor.js core build with modules that only a subset of people are interested in using. Doing so makes the library seem heavier than it is. Modularity FTW!

Issue #250, opened 2016-10-25
embedding an image using option data-uri fails


cmoulliard:
When we use the attribute `data-uri` to embed an image

```
:data-uri:

== Image
[[id_of_litoria_picture]]
.Title of the picture displayed under the picture
image::image/litoria-chloris.jpg[width="40%"]
```

then we get an Opal error as the flag attribute rb is not supported

```
Error: Unknown file open flag: rb
    at stringToFlags (fs.js:589:9)
    at Object.fs.openSync (fs.js:640:51)
    at $File.b.defn.w [as $initialize] (/Users/chmoulli/Temp/litoria/package/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.min.js:25:207)
    at Opal.defn.TMP_4 [as $new] (/Users/chmoulli/Temp/litoria/package/node_modules/opal-runtime/src/opal.js:3140:23)
    at ːopen (/Users/chmoulli/Temp/litoria/package/node_modules/opal-runtime/src/nodejs.js:187:19)
    at e.b.defn.D [as $generate_data_uri] (/Users/chmoulli/Temp/litoria/package/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.min.js:285:57)
    at e.b.defn.Z [as $image_uri] (/Users/chmoulli/Temp/litoria/package/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.min.js:283:194)
    at p.b.defn.y [as $image] (/Users/chmoulli/Temp/litoria/package/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.min.js:98:431)
    at p.ː__send__ (/Users/chmoulli/Temp/litoria/package/node_modules/opal-runtime/src/opal.js:3228:21)
    at p.b.defn.w (/Users/chmoulli/Temp/litoria/package/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.min.js:32:84)
    at e.b.defn.H (/Users/chmoulli/Temp/litoria/package/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.min.js:297:459)
    at a.$$p.d (/Users/chmoulli/Temp/litoria/package/node_modules/asciidoctor.js/dist/npm/asciidoctor-core.min.js:298:133)
    at Opal.yield1 (/Users/chmoulli/Temp/litoria/package/node_modules/opal-runtime/src/opal.js:1198:14)
[...]

PR #251, opened 2016-10-25
Embed image in base64 (Node.js)

Mogztter:
Ref #250 

The main issue is that we need to provide a JavaScript implementation for Base64 because Asciidoctor needs it [here](https://github.com/asciidoctor/asciidoctor/blob/a16566c34e51e9f1078b67374e2869281a040205/lib/asciidoctor/abstract_node.rb#L372).

* In Node.js we can use [Buffer](https://nodejs.org/docs/latest/api/buffer.html#buffer_buffers_and_character_encodings): `buf.toString('base64')`
* In Browsers, we can use `btoa`: https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/btoa

Or we could use Opal.Base64 module: https://github.com/opal/opal/blob/master/stdlib/base64.rb


lib/asciidoctor/opal_ext/file.rb
```diff
@@ -19,12 +19,19 @@ class File
...
 
-  def initialize(path, mode = 'r')
+  def initialize(path, flags = 'r')
     @path = path
     @contents = nil
     @eof = false
     @lineno = 0
-    @fd = `require('fs').openSync(path, mode)` if JAVASCRIPT_PLATFORM == 'node'
+    # binary flag is unsupported
+    binary_flag_regexp = /b/
```
mojavelinux: I think a better approach here is to use `delete`. No need to be diving into regex.

```ruby
flags = (flags.delete 'b')
```

If you want to check if it was present first, add an `include?`

```ruby
binary = false
if flags.include? 'b'
  binary = true
  flags = flags.delete 'b'
end
```

Always try to find a way to do it without regex first.

mojavelinux: I think to read binary data with node, you set the encoding (second argument) to empty string or null. Then, you get a buffer object. You'll then have to do something with that buffer object...or trust that the next step in the chain properly converts it to base64.

mojavelinux: If you read the image file with the encoding set to 'utf8', you get corrupt data. (I'm not sure if that's what the code is doing. I'm just mentioning that a general comment).

Mogztter: > If you read the image file with the encoding set to 'utf8', you get corrupt data. (I'm not sure if that's what the code is doing. I'm just mentioning that a general comment).

I think you are right, the Ruby code is:
```ruby
bindata = ::File.open(image_path, 'rb') {|file| file.read }
%(data:#{mimetype};base64,#{::Base64.encode64(bindata).delete EOL})
```

So basically we are instantiating a new File with `initialize` and then we are using `File.read` which resolves to:

```javascript
return require('fs').readFileSync(path, 'utf8');
```

Maybe we should implement `IO.binread` ? In this case we will use the following code, instead of `File.read`:
```ruby
bindata = ::IO.binread(image_path)
```

[abstract_node.rb#L367](https://github.com/asciidoctor/asciidoctor/blob/a16566c34e51e9f1078b67374e2869281a040205/lib/asciidoctor/abstract_node.rb#L367)

Mogztter:
```javascript
var buf = fs.readFileSync(image_path);
return buf.toString('binary');
```
> 'binary' - Alias for 'latin1'.
https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings

mojavelinux: :+1:

Mogztter:
@mojavelinux Does this look good to you ?

mojavelinux:
Nice work!

Issue #201, opened 2016-08-17
NilClass_alloc { '$$id': 4, apply: [Function], call: [Function] }


cmoulliard:
This node.js script generates this exception `NilClass_alloc { '$$id': 4, apply: [Function], call: [Function] }` when the content is coming from a file and no stack trace is available

```javascript
var asciidoctor = require('asciidoctor.js')();
var fs = require('fs')

var opal = asciidoctor.Opal;
var processor = null;
var useExtensions = true;

if (useExtensions) {
    processor = asciidoctor.Asciidoctor(true);
}
else {
    processor = asciidoctor.Asciidoctor();
}

// Check if the path of the file to be converted is passed as parameter
if (process.argv[2]) {
    readFile(process.argv[2]);
} else {
    console.error("Please pass as parameter the path to the html file to be converted !")
}

function readFile(path) {
    var content = fs.readFileSync(path, 'utf8')
    convertToHtml(content);
}


function convertToHtml(content) {
    //console.log(content);
    var options = opal.hash({doctype: 'inline', attributes: ['showtitle']});
    try {
        var html = processor.$convert(content, options);
        console.log(html);
    } catch (e) {
        console.error(e.stack);
    }
}
```

How can we get the stack trace and more info about the error ?

Mogztter:
Hello @cmoulliard 

This is not an exception, in fact `NilClass_alloc` is the object returned by `processor.$convert(content, options);`.

In Ruby, the method should return a `Nil` instance translated by Opal to a `NilClass_alloc` instance.

The problem is that `inline` document type can only convert a single paragraph. To convert a document you need to use an `article` document type (this is the default value).
You can read more about document types here: http://asciidoctor.org/docs/user-manual/#document-types

Here's an updated version of the `convertToHtml` function:
```javascript
function convertToHtml(content) {
    var options = opal.hash({attributes: ['showtitle']});
    try {
        var html = processor.$convert(content, options);
        console.log('html', html);
    } catch (e) {
        console.error('error', e.stack);
    }
}
```

mojavelinux:
>  NilClass_alloc is the object returned by processor.$convert(content, options);.

I don't think that's correct. `$convert_file` returns nil because it directs the output to the file, but `$convert` should return the String that is converted.

Mogztter:
`$convert` method returns the String that is converted when the content is a single paragraph otherwise the method returns a `Nil` object.
[...]

PR #252, opened 2016-10-29
Return an empty string instead of Opal.nil

Mogztter:
`console.log(Opal.nil)` produces the following output :

```
NilClass_alloc { '$$id': 4, apply: [Function], call: [Function] }
```

This output can be confusing and is not consistent with Asciidoctor (Ruby).
We now return an empty string.

Ref #201


mojavelinux:
:+1:

PR #263, opened 2016-11-01
Improve auto-detection and allow to configure explicitly Asciidoctor.js

Mogztter:
Should resolve https://github.com/asciidoctor/atom-asciidoc-preview/pull/201

### DIFF
--- a/lib/asciidoctor/opal_ext.rb
+++ b/lib/asciidoctor/opal_ext.rb
@@ -1,64 +1,103 @@
 %x(
-  var isNode = typeof module !== 'undefined' && module.exports,
-      isElectron = typeof process === 'object' && typeof process.versions === 'object' && typeof process.versions.electron === 'string',
-      isBrowser = typeof window !== 'undefined',
-      isNashorn = typeof Java !== 'undefined' && Java.type,
-      isRhino = typeof java !== 'undefined',
-      isPhantomJS = typeof window !== 'undefined' && typeof window.phantom !== 'undefined',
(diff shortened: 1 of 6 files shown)


lib/asciidoctor/opal_ext.rb
```diff
@@ -1,64 +1,104 @@
...
-    Opal.load("nodejs");
-    Opal.load("pathname");
-    Opal.load("base64");
   }
-  else if (isNashorn) {
-    platform = 'java';
-    engine = 'nashorn';
+
+  if (typeof platform === 'undefined') {
+    throw new Error('Unable to auto-detect the platform! Please configure Asciidoctor.js');
```
Mogztter: If we can't determine the `platform` some features (mostly I/O features like include macro, embedded images, etc...) will not work.

@mojavelinux Do you think it's a good idea to throw an exception or should we just "ignore" this issue ? Initially I wanted to print a warning but since we do not know the "platform", we do not know how to print a log...

mojavelinux: I think throwing an exception is correct since we should be able to assume in core that we can perform file operations in some way.

src/npm/prepend-core.js
```diff
@@ -1,7 +1,7 @@
-module.exports = function() {
+module.exports = function(ajsConfig) {
```
mojavelinux: I'd like a better name than ajsConfig. Perhaps something like runtimeConfig or hostConfig?

mojavelinux: :+1:

PR #267, opened 2016-11-12
Run tests on Appveyor

### LINKED ISSUE #266: Activate tests on Appveyor (Windows)

Mogztter:
Resolves #266

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,11 @@ environment:
 install:
 - git submodule update --init --recursive
 - npm install
-- npm install -g bower
 build_script:
 - node npm/appveyor.js
+- node npm/test/unsupported-features.js
+- node npm/test/jasmine-browser.js
+- node npm/test/jasmine-node.js
+- node npm/test/jasmine-webpack.js
+- npm run test_karma_browserify
+- npm run test_karma_requirejs
--- a/lib/asciidoctor/opal_ext/file.rb
+++ b/lib/asciidoctor/opal_ext/file.rb
@@ -12,8 +12,6 @@ module Kernel
 end
 
 class File
-  SEPARATOR = '/'
-  ALT_SEPARATOR = nil
 
   attr_reader :eof
   attr_reader :lineno
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -50,8 +50,9 @@ Builder.prototype.build = function (callback) {
     function (callback) { builder.clean(callback); }, // clean
     function (callback) { builder.downloadDependencies(callback); }, // download dependencies
     function (callback) { builder.compile(callback); }, // compile
+    function (callback) { builder.patchAsciidoctorCore(callback); }, // patch Asciidoctor core. TODO: remove once Asciidoctor 1.5.6 is released
     function (callback) { builder.replaceUnsupportedFeatures(callback); }, // replace unsupported features
-    function (callback) { builder.replaceDefaultStylesheetPath(callback); }, // replace unsupported features
+    function (callback) { builder.replaceDefaultStylesheetPath(callback); }, // replace the default stylesheet path
     function (callback) { builder.generateUMD(callback); }, // generate UMD
     function (callback) { builder.uglify(callback); } // uglify (optional)
   ], function () {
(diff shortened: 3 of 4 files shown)


mojavelinux:
Need any assistance with AppVeyor, or do you have it handled?

Mogztter:
> Need any assistance with AppVeyor, or do you have it handled?

@mojavelinux No, I've got it thanks :wink: 
With https://github.com/asciidoctor/asciidoctor/pull/1925 the tests should passed. I think I will monkey-patch  Asciidoctor.js to get ahead on the Asciidoctor core release.

mojavelinux:
Sounds good. I'll try to do another Asciidoctor core release soon, but you don't have to wait on it.

Mogztter:
Still some issues on Windows with file paths: https://ci.appveyor.com/project/asciidoctor/asciidoctor-js/build/997

 * [ ] `SecurityError: Jail is not an absolute path: C:\projects\asciidoctor-js`
    * `Node.js Converting file should be able to convert a file and embed an image`
    * `Node.js Converting file should be able to convert a file with to_dir and to_file`
    * `Node.js Converting file should be able to convert a file with to_dir`

 * [ ] `SecurityError: Jail is not an absolute path: C:\projects\asciidoctor-js\spec\node`
   * `Node.js Converting file should be able to convert a file with custom css embedded`

 * [ ] `Error: ENOENT: no such file or directory, open 'C:\projects\asciidoctor-js\C:\projects\asciidoctor-js\C:\projects\asciidoctor-js\build\css\asciidoctor.css'`
    * `Node.js Parsing should embed assets`

Mogztter:
I think I found the issue!

In `stdlib/nodejs/file.rb`

```ruby
  if `__path__.sep !== #{Separator}`
    ALT_SEPARATOR = Separator
    Separator = SEPARATOR = `__path__.sep`
  end
```
https://github.com/opal/opal/blob/47c2e60d9ec02d5c8fdaf2feee1308772a145e84/stdlib/nodejs/file.rb#L47-L50

On Windows, `__path__.sep = '\'` and by default `Separator = SEPARATOR = '/'` in `opal/corelib/file.rb`

```ruby
class File < IO
  Separator = SEPARATOR = '/'
  ALT_SEPARATOR = nil
  # ....
end
```
https://github.com/opal/opal/blob/47c2e60d9ec02d5c8fdaf2feee1308772a145e84/opal/corelib/file.rb#L2


So we end up with:

```ruby
path.sep = "\\"
File.ALT_SEPARATOR = "/"
File.SEPARATOR = "\\"
```

In `lib/asciidoctor/path_resolver.rb`:

```ruby
@file_separator = file_separator ? file_separator : (::File::ALT_SEPARATOR || ::File::SEPARATOR)
```
https://github.com/asciidoctor/asciidoctor/blob/7282b0a5864df2f587b5a5a8f3cda10f71ed5547/lib/asciidoctor/path_resolver.rb#L125

`File.ALT_SEPARATOR` is defined so `@file_separator = "/"`

In `lib/asciidoctor/path_resolver.rb`:

```ruby
elsif @file_separator == BACKSLASH && (WindowsRootRx.match? path)
```
https://github.com/asciidoctor/asciidoctor/blob/7282b0a5864df2f587b5a5a8f3cda10f71ed5547/lib/asciidoctor/path_resolver.rb#L146

`(WindowsRootRx.match? path)` is true but `@file_separator` is *not* a `BACKSLASH` :disappointed: 

@mojavelinux I think Opal is wrong, `ALT_SEPARATOR` should be equal to `__path__.sep` no ?

> ALT_SEPARATOR: platform specific alternative separator
PATH_SEPARATOR: path list separator

https://ruby-doc.org/core-2.1.1/File.html

mojavelinux:
In Ruby, `File::SEPARATOR` is always forward slash. `File::ALT_SEPARATOR` is only set on platforms that don't use forward slash as the separator. This is the only place you should find the backslash value. `File::ALT_SEPARATOR` should not be set if the system uses forward slash as the separator.

So Opal is incorrect. It should set `File::ALT_SEPARATOR` to `__path__.sep`, but only if that value is not forward slash (though, technically, I don't think it will hurt if it's the same).

`File::PATH_SEPARATOR` is something completely different. It's the separator used in the PATH environment variable to separate path entries (not segments of a path).

mojavelinux:
One thing that might help is to update this pull request and print the values that Ruby sets on Windows. Add something like this to the AppVeyor build:

```
ruby -e ' p [File::SEPARATOR, File::Separator, File::ALT_SEPARATOR]'
```

On Linux, that prints:

```
["/", "/", nil]
```

mojavelinux:
Opal seems to get it right on Linux. I can't verify it on Windows atm.

mojavelinux:
I want to point out that my aim is to get this working without hacks. I want to avoid having to do any custom path wrangling/mangling in Asciidoctor.js like the Gradle plugin does. (See https://github.com/asciidoctor/asciidoctor-gradle-plugin/blob/cecab658d6f64d14bbbd7921f31acc817be2d477/src/main/groovy/org/asciidoctor/gradle/AsciidoctorTask.groovy#L614-L620). If that's necessary, I consider it a failure of core.

Mogztter:
The full tests suite is running on Windows!!! :tada:

Issue #198, opened 2016-08-11
input.$respond_to? is not a function


bguiz:
I'm getting `input.$respond_to? is not a function` errors when I tried to build my asciidoc files.

Any idea where to start looking, or what would cause this?

```
C:\Users\bguiz\code\book\node_modules\asciidoctor.js\node_modules\opal-npm-wrapper\index.js:4296
      throw exception;
      ^

TypeError: input.$respond_to? is not a function
    at module_constructor.$load (C:\Users\bguiz\code\book\node_modules\asciidoctor.js\dist\npm\asciidoctor-core.js:12265:50)
    at module_constructor.<anonymous> (C:\Users\bguiz\code\book\node_modules\asciidoctor.js\dist\npm\asciidoctor-core.js:12372:20)
    at Object.<anonymous> (C:\Users\bguiz\code\book\build.js:23:22)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
    at startup (node.js:139:18)
    at node.js:968:3
```

Mogztter:
Hmmm not sure... :thought_balloon: 
Could you please give me your code + a document to reproduce this issue ?

Mogztter:
ping @bguiz ?

Mogztter:
pong @bguiz  :smile:

bguiz:
Sorry for the late response - I switched to using the Ruby asciidoctor

I'll close this ticket since I'm not going to follow up

Mogztter:
Ok, thanks for you reply.
By any chance, did you keep your JavaScript code + AsciiDoc files ? I would like to reproduce and fix this issue :smile:

hsablonniere:
I just got the same problem when sending a Buffer instead of a String to `convert`. I think we can close this.

mojavelinux:
Should we handle/convert the Buffer?

Mogztter:
Why not or at least print a useful error message if the user is using the wrapper API.

Mogztter:
Buffer can now be used in the public API on `load` and `convert` method

PR #271, opened 2016-12-22
Handle and convert Node's Buffer

Mogztter:
Ref #198


mojavelinux:
Now that's this API living up to our dreams! \o/

PR #277, opened 2017-01-27
docs(readme): add note on release process

thetutlage:
Asciidoctor.js does not release as per semver and maintains an identical version across different versions. So added a note on same


Mogztter:
Thanks @thetutlage, I've made some adjustments but overall this is really good!
Tell me if this is OK for you, then I'll proceed and merge your pull request.

thetutlage:
Yup, looks good to me

Mogztter:
And merged! thanks again :wink:

mojavelinux:
:+1:

PR #280, opened 2017-03-06
Integrate with Opal master

Mogztter:
RequireJS is failing, otherwise all good!

I don't understand why but RequireJS does not resolve `asciidoctor` but does resolve `asciidoctor/docbook`. @anthonny @hsablonniere any idea ?

`asciidoctor` object is equal to: 

```javascript
function c(c, j, k) {
  var n = [];
  j = 1 == j ? {entropy: !0} : j || {};
  var s = g(f(j.entropy ? [c, i(a)] : null == c ? h() : c, 3), n), t = new d(n), u = function () {
    for (var a = t.g(m), b = p, c = 0; a < q;)a = (a + c) * l, b *= l, c = t.g(1);
    for (; a >= r;)a /= 2, b /= 2, c >>>= 1;
    return (a + c) / b
  };
  return u.int32 = function () {
    return 0 | t.g(4)
  }, u.quick = function () {
    return t.g(4) / 4294967296
  }, u.double = u, g(i(t.S), a), (j.pass || k || function (a, c, d, f) {
    return f && (f.S && e(f, t), a.state = function () {
      return e(t, {})
    }), d ? (b[o] = a, c) : a
  })(u, s, "global" in j ? j.global : this == b, j.state);
}
```

entropy is high!!! :rofl: 

https://github.com/asciidoctor/asciidoctor.js/blob/0866d3335699f728f0a9197f8f01c51ce92525cc/spec/requireJS/asciidoctor.spec.js#L8

### DIFF
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -350,7 +350,7 @@ Builder.prototype.replaceUnsupportedFeatures = function (callback) {
   var path = 'build/asciidoctor-lib.js';
   var data = fs.readFileSync(path, 'utf8');
   log.debug('Replace (g)sub! with (g)sub');
-  data = data.replace(/(\(\$\w+ = \(\$\w+ = (\w+)\))\['(\$g?sub)!'\]/g, '$2 = $1[\'$3\']');
+  data = data.replace(/\$send\(([^,]+), '(g?sub)!'/g, '$1 = $send\($1, \'$2\'');
   fs.writeFileSync(path, data, 'utf8');
   callback();
 };
@@ -367,7 +367,7 @@ Builder.prototype.replaceDefaultStylesheetPath = function (callback) {
     '  stylesheetsPath = "css";\n' +
     '}\n' +
     'return ((($a = self.primary_stylesheet_data) !== false && $a !== nil && $a != null) ? $a : self.primary_stylesheet_data = Opal.get("IO").$read(Opal.get("File").$join(stylesheetsPath, "asciidoctor.css")).$chomp());';
-  data = data.replace(/(ːprimary_stylesheet_data\(\)\ {\n)(?:[^}]*)(\n\s+}.*)/g, '$1' + primaryStylesheetDataImpl + '$2');
+  data = data.replace(/(function \$\$primary_stylesheet_data\(\)\ {\n)(?:[^}]*)(\n\s+}.*)/g, '$1' + primaryStylesheetDataImpl + '$2');
   fs.writeFileSync(path, data, 'utf8');
   callback();
 };
--- a/npm/test/jasmine-browser-min.js
+++ b/npm/test/jasmine-browser-min.js
@@ -22,7 +22,6 @@ jasmine.loadConfig({
 
 // This code is necessary to fake a browser for Opal
 //--------------------------------------------------
-window = {};
 process.browser = true;
 
 if (typeof XMLHttpRequest === 'undefined') {
--- a/npm/test/jasmine-browser.js
+++ b/npm/test/jasmine-browser.js
@@ -18,7 +18,6 @@ jasmine.loadConfig({
 
 // This code is necessary to fake a browser for Opal
 //--------------------------------------------------
-window = {};
 process.browser = true;
 
 if (typeof XMLHttpRequest === 'undefined') {
--- a/npm/test/jasmine-webpack.js
+++ b/npm/test/jasmine-webpack.js
@@ -28,7 +28,7 @@ webpack(config, function () {
 
   // This code is necessary to fake a browser for Opal
   //--------------------------------------------------
-  window = {};
+  window = global;
 
   if (typeof XMLHttpRequest === 'undefined') {
     XMLHttpRequest = require('xmlhttprequest').XMLHttpRequest;
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "asciidoctor.js",
-  "version": "1.5.5-4",
+  "version": "1.5.5-integration1",
   "description": "A JavaScript AsciiDoc processor, cross-compiled from the Ruby-based AsciiDoc implementation, Asciidoctor, using Opal",
   "main": "dist/asciidoctor.js",
   "engines": {
-    "node": ">=0.12",
+    "node": ">=4",
     "npm": ">=3.0.0"
   },
   "files": [
@@ -47,19 +47,19 @@
   },
   "homepage": "https://github.com/asciidoctor/asciidoctor.js",
   "dependencies": {
-    "opal-runtime": "0.10.2",
+    "opal-runtime": "0.11.0-integration6",
     "xmlhttprequest": "~1.7.0"
   },
   "devDependencies": {
-    "asciidoctor-docbook.js": "1.5.5-4",
+    "asciidoctor-docbook.js": "1.5.5-integration2",
     "asciidoctor-reveal.js": "1.0.1",
-    "asciidoctor-template.js": "1.5.5-4",
+    "asciidoctor-template.js": "1.5.5-integration1",
     "async": "^1.5.0",
     "bestikk-download": "^0.1.0",
     "bestikk-fs": "^0.1.0",
     "bestikk-jdk-ea": "0.1.1",
     "bestikk-log": "0.1.0",
-    "bestikk-opal-compiler": "0.2.0",
+    "bestikk-opal-compiler": "0.3.0-integration5",
     "bestikk-uglify": "0.1.1",
     "bower": "^1.5.3",
     "browserify": "^13.0.0",
--- a/spec/browser/asciidoctor.spec.js
+++ b/spec/browser/asciidoctor.spec.js
@@ -5,7 +5,7 @@ var testOptions = {
   baseDir: 'file://' + path.join(__dirname, '..')
 };
 
-var asciidoctor = require('./asciidoctor.js')();
+var asciidoctor = require('./asciidoctor.js')({runtime: {platform: 'browser'}});
 require('asciidoctor-docbook.js');
 
 commonSpec(testOptions, asciidoctor);
--- a/spec/requireJS/main.js
+++ b/spec/requireJS/main.js
@@ -21,7 +21,9 @@ require.config({
 
   config: {
     'asciidoctor': {
-      'ioModule': 'phantomjs'
+      'runtime': {
+        'ioModule': 'xmlhttprequest'
+      }
     }
   },
 
--- a/spec/webpack/spec.js
+++ b/spec/webpack/spec.js
@@ -1,5 +1,5 @@
 var commonSpec = require('../share/common-spec.js');
-var asciidoctor = require('../../build/asciidoctor.js')();
+var asciidoctor = require('../../build/asciidoctor.js')({runtime: {platform: 'browser'}});
 require('asciidoctor-docbook.js');
 
 commonSpec(testOptions, asciidoctor);
--- a/src/template-asciidoctor.js
+++ b/src/template-asciidoctor.js
@@ -4,6 +4,7 @@ if (typeof Opal === 'undefined' && typeof module === 'object' && module.exports)
 
 if (typeof Opal === 'undefined') {
 //#{opalCode}
+  Opal.require('opal');
 }
 
 // UMD Module


mojavelinux:
Amazing. Thanks for your hard work to keep us up with the latest and greatest!

Mogztter:
I found the entropy code: https://github.com/opal/opal/blob/master/opal/corelib/random/seedrandom.js.rb
This changes was introduced in https://github.com/opal/opal/commit/c5d225a2c008f33caf1fcd92e228bfcbc3128a58 but I have no idea why Require.js resolves `asciidoctor` as the `seedrandom` function... :thinking:

Mogztter:
I think this is why: https://github.com/davidbau/seedrandom/blob/cac59aac49c2a8ae7cd48d33e1926425f83e61ec/seedrandom.js#L233-L241

The module is not named and the code is bundled in a file named asciidoctor.js ! :scream: 

> You should avoid naming modules yourself, and only place one module in a file while developing. However, for tooling and performance, a module solution needs a way to identify modules in built resources

http://requirejs.org/docs/whyamd.html#namedmodules

Possible solutions:

 * Ask `seedrandom` to add a module id (but Require.js does not recommend it)
 * Add a module id in [opal/corelib/random/seedrandom.js.rb](https://github.com/opal/opal/blob/master/opal/corelib/random/seedrandom.js.rb) since the file is bundled in opal.js
 * Other ?

mojavelinux:
I'm in favor of the fix in Opal, esp since it's bundled.

Mogztter:
Me too :+1:

Mogztter:
Green bar!

mojavelinux:
green-bar

mojavelinux:
It looks like Opal now partially implements String#unpack, so our workaround breaks. We need to check for the value C3, then delegate to the built-in method otherwise.

mojavelinux:
It seems like getId() is returning some sort of wrapped String object instead of a JavaScript string. Are you aware of this change? I have to add `.toString()` to get the test for ID to pass in spec/node/asciidoctor.spec.js (but oddly, only the first occurrence).

mojavelinux:
It's when the ID is auto-generated.

mojavelinux:
Yep, Opal is getting confused by the syntax. I'll need to update core. We might find other cases of this, which I think you've already spotted.

```ruby
section.id = (id = attributes['id'] || (Section.generate_id section.title, document))
```

It's assigning an array value.

mojavelinux:
Actually, the problem is somewhere else. The following statement returns some sort of array object (but it's not really an Array...it's something in between).

```
"foobar".$chomp("a");
```

When there is nothing to chomp, we don't get a JavaScript String back.

mojavelinux:
This is a major problem in Opal. It's happening in a lot of places. String operations sometimes returns `self`, which is not a JavaScript string. Thus any object that makes it's way out of the API without being combined with another string is this pseudo-String object.

mojavelinux:
Here's another example:

```js
"foobar".$delete("");
```

elia:
@mojavelinux it returns the actual string objects (vs. the string value), you can get the native string by calling `.valueOf()` but I'm not sure how that could be applied generally.

Would you mind opening an issue for this on opal/opal with some more context about your use case? That would help in considering pros/cons of returning `self.valueOf()` instead of `self` from core string methods.

mojavelinux:
@elia Yes, we'll open an issue once we determine exactly what we're looking for. The main issue I have atm is that this changes the API. Tests that were working with Opal 0.10 are now failing without any change to the code Opal is compiling.

elia:
> Tests that were working with Opal 0.10 are now failing without any change to the code Opal is compiling.

@mojavelinux that's unfortunate, anyway I can confirm that 0.10 → 0.11 is a quite major release and fixes/adds a ton of stuff, that's also why it's not out yet. **Anything I can do to help let me know ⛑.**

PS. just checked too and it has the same behavior wrt `"foobar".$chomp("a");` 😃

mojavelinux:
Thanks @elia!

Turns out, you are correct. This is the same behavior. So what changed is that we added a test in the interim.

So this is not a migration problem, but rather one of general spec. When a user is using an Opal-compiled library, should they get different value types from the same String method. An open question for sure.

Btw, I changed the logic in Asciidoctor so that we don't attempt to chomp when there's nothing to chomp (turns it, it is faster anyway). So we won't have to change the test in Asciidoctor.js.

mojavelinux:
@Mogztter Here's the fix we need for unpack:

```ruby
alias :_original_unpack :unpack

def unpack format
  if format == 'C3'
    self[0, 3].bytes.select.with_index {|_, i| i.even? } 
  else
    _original_unpack format
  end
end
```

This makes it future-proof since we just duplicate logic in _original_unpack once fully implemented.

mojavelinux:
(If Opal supports it, we could use a refinement instead. See https://ruby-doc.org/core-2.1.1/doc/syntax/refinements_rdoc.html).

Mogztter:
> Here's the fix we need for unpack in `lib/asciidoctor/opal_ext/string.rb`

@mojavelinux Just to be sure this fix need to be done once we switch to Asciidoctor core master branch ? Tests are running fine with Asciidoctor core 1.5.5 and Opal master. Can we merge this pull request ?

Mogztter:
I was expecting Asciidoctor.js to be faster but currently it's ~10% slower:


_ | Opal 0.11 | Opal 0.10
------------| ------------- | -------------
Run 1     |      `0,942`|	`0,851`
Run 2     |      `0,716`|	`0,663`
Run 3     |      `0,67`  |	`0,614`
Run 4     |      `0,648`|	`0,588`
Average |	     `0,744`|	`0,679`

mojavelinux:
> Just to be sure this fix need to be done once we switch to Asciidoctor core master branch ? 

Yes, the fix is still needed. Opal has a partial implementation of unpack, but does not cover this format string. And this impl is future proof.

mojavelinux:
If you're comfortable with it, I'm ready to merge this pull request. We'll still need to make changes to support Asciidoctor master, but that's a separate issue.

mojavelinux:
I noticed in the new build that Opal is being required and bundled in asciidoctor.js. Perhaps that could account for part of the slowdown?

Btw, the upgrade to Asciidoctor 1.5.6 is going to bring a ~10% improvement in speed, so there's some hope we'll gain this speed back.

mojavelinux:
Actually, I see that the bundling is the same...so that is an unrelated change to the issue at hand.

I just checked and here are the results I got:

0.10 (from the 1.5.5-4 release)

```
Load scripts: 0.423s
Run #1: 1.259s
Run #2: 0.853s
Run #3: 0.682s
Run #4: 0.64s
Run #5: 0.588s
Run #6: 0.598s
Run #7: 0.605s
Run #8: 0.564s
Run #9: 0.582s
Run #10: 0.553s
```

0.11 (from master)

```
Load scripts: 0.462s
Run #1: 1.374s
Run #2: 1.045s
Run #3: 0.842s
Run #4: 0.857s
Run #5: 0.825s
Run #6: 0.79s
Run #7: 0.812s
Run #8: 0.79s
Run #9: 0.78s
Run #10: 0.824s
```

Both are using the Asciidoctor 1.5.5 code base.

mojavelinux:
I thought we had integrated with Opal 0.11. Is there a reason this PR is still pending?

Or is it that we've switched to the 0.11 compiler, but not the runtime.

Mogztter:
> I thought we had integrated with Opal 0.11. Is there a reason this PR is still pending?

Not yet, I wanted to run benchmark first :zap: 
Opal 0.11 is a bit slower but using the latest version of Opal is not really an option since we don't stay behind (and this version fixes a lot of issues on I/O)

I will work on improving the performance.

> Or is it that we've switched to the 0.11 compiler, but not the runtime?

We switched both compiler and runtime in this pull request. Currently master branch is using Opal 0.10 compiler and runtime.

mojavelinux:
Great news!

Is there a benchmark suite for Opal that would allow us to see which functions / methods slowed down between 0.10 and 0.11. Looking at discrete operations should allow us to figure out which ones are the culprit.

The benchmark/ips gem is extremely valuable for this use case. Having something like that for Opal would allow us to keep our eye on performance. (See https://github.com/JuanitoFatas/fast-ruby for some examples).

mojavelinux:
@Mogztter Perhaps this change in Opal will help our performance problem:

https://github.com/opal/opal/pull/1637

I'll admit that I applied it manually and didn't see much of a change, but perhaps I did something incorrectly.

Btw, we can study performance using the following tool:

```
node --prof build/benchmark/run.js
node --prof-process isolate-*.log > profile.txt
```

Here's what I see at the top right now:

```
 [JavaScript]:
   ticks  total  nonlib   name
    589    6.3%    6.4%  Stub: LoadICStub
    390    4.2%    4.2%  Builtin: ObjectHasOwnProperty
    241    2.6%    2.6%  LazyCompile: *constGetMultiple /home/dallen/projects/asciidoctor/asciidoctor.js-pristine/build/      asciidoctor.js:191:28
    182    2.0%    2.0%  KeyedLoadIC: A keyed load IC from the snapshot
    154    1.7%    1.7%  Builtin: CallFunction_ReceiverIsNotNullOrUndefined
     93    1.0%    1.0%  Builtin: Call_ReceiverIsNotNullOrUndefined
     61    0.7%    0.7%  Stub: VectorStoreICStub
     51    0.5%    0.6%  LazyCompile: *$$gsub /home/dallen/projects/asciidoctor/asciidoctor.js-pristine/build/asciidoctor.js: 6824:66
```

The gsub is the slowest part of Asciidoctor, so that's to be expected. Everything above that is a concern.

It would be interesting to do the same against Opal 0.10.

mojavelinux:
Here's what it looks like with Opal 0.10 running on Asciidoctor.js 1.5.5-4.

```
 [JavaScript]:
   ticks  total  nonlib   name
    421    5.6%    5.6%  Stub: LoadICStub
    147    1.9%    2.0%  KeyedLoadIC: A keyed load IC from the snapshot
    137    1.8%    1.8%  Builtin: CallFunction_ReceiverIsNotNullOrUndefined
    130    1.7%    1.7%  LazyCompile: *Opal.ancestors /tmp/asciidoctor.js/node_modules/opal-runtime/src/opal.js:891:28
     57    0.8%    0.8%  Builtin: Call_ReceiverIsNotNullOrUndefined
     50    0.7%    0.7%  Stub: GrowArrayElementsStub
     46    0.6%    0.6%  Stub: CEntryStub
     39    0.5%    0.5%  LazyCompile: *toString native string.js:30:24
     38    0.5%    0.5%  LazyCompile: ~Opal.hash_get /tmp/asciidoctor.js/node_modules/opal-runtime/src/opal.js:1661:27
     33    0.4%    0.4%  LazyCompile: *ːgsub /tmp/asciidoctor.js/node_modules/opal-runtime/src/opal.js:6549:53
     33    0.4%    0.4%  Handler: $$is_boolean
```

So ObjectHasOwnProperty seems to be a real concern.

Mogztter:
`constGetMultiple` is called many times. Here's a top 10 by `const_name`:

|`const_name`|`calls`
-------|-----------
| 'Opal'|557218|
| 'Integer'| 252130|
| 'String'| 251994|
| 'MatchData'| 33304 |
| 'ListRxMap'| 25448 |
| 'Compliance'| 22984| 
| 'DELIMITED_BLOCK_LEADERS'| 22192|
| 'LIST_CONTINUATION'| 16408|
| 'EOL'|14384|
| 'Array'| 11868|

Mogztter:
Since we do not redefine `Opal`, `Integer`, `String`, `MatchData` in another scope, we could replace  

*  `Opal.const_get($scopes, 'Opal', true, true)` by `Opal.Opal`, 
* `Opal.const_get($scopes, 'Integer', true, true)` by `Opal.Integer`, 
* etc...

This is a drastic change but the performance greatly improve.

Mogztter:
@mojavelinux Maybe we should continue the discussion on a dedicated issue ?

mojavelinux:
I agree. Though I wonder if it should be in Opal instead of Asciidoctor.js. We just happen to serve as the case study.

elia:
Just a heads up that I'm reworking the constants system which I think has a bug in 0.11 that causes https://github.com/opal/opal/issues/1607. While I'm there I hope I'll make it faster too. 🏎

Mogztter:
@elia cool!

elia:
@Mogztter @mojavelinux so #1653 has been merged and should have [great performance](https://gist.github.com/elia/66341dae29adf153c912f672e0eff354), looking forward to the update results! 😄

mojavelinux:
I am *thrilled* to hear that. Great job! I can't wait to see the results! :champagne:

PR #281, opened 2017-03-07
make Asciidoctor.js compatible with Asciidoctor 1.5.6

mojavelinux:
- define a limited implementation of String#unpack
- update Kernel#freeze to return self
- rename String#limit to String#limit_bytesize and add alias

### DIFF
--- a/lib/asciidoctor/opal_ext/kernel.rb
+++ b/lib/asciidoctor/opal_ext/kernel.rb
@@ -1,6 +1,8 @@
 module Kernel
+  # Redefine Kernel#freeze to suppress "not supported" warning message from Opal.
+  # TODO Remove once https://github.com/opal/opal/issues/1253 is resolved.
+  # NOTE Opal should define freeze on Object, not Kernel.
   def freeze
-    # noop, suppress "not supported" warning message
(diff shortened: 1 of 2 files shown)


lib/asciidoctor/opal_ext/kernel.rb
```diff
@@ -1,6 +1,8 @@
 module Kernel
+  # Redefine Kernel#freeze to suppress "not supported" warning message from Opal.
+  # TODO Remove once https://github.com/opal/opal/issues/1253 is resolved.
```
Mogztter: I think we can now remove this code because https://github.com/opal/opal/issues/1253 is "resolved" (Elia made some changes, described here: https://github.com/opal/opal/pull/1345#issuecomment-200085541).
What we need to do is to set `Opal.config.unsupported_features_severity = 'ignore'` to suppress the "not supported" warning message.

mojavelinux: I'm 100% in support of that. Whenever you upgrade Opal and it takes over control of one of our overrides, feel free to remove it at the same time.

Mogztter:
LGTM, thanks :wink:

mojavelinux:
:+1:

PR #283, opened 2017-03-12
resolves #282 add support for block filter in findBy API method

### LINKED ISSUE #282: API, AbstractBlock.find_by: allow to use a function as second attribute
Reference API: https://github.com/asciidoctor/asciidoctor/blob/7282b0a5864df2f587b5a5a8f3cda10f71ed5547/lib/asciidoctor/abstract_block.rb#L268

To pass a function, we need to use `Opal.block_send`:
```js
Opal.block_send(doc, 'find_by', toHash(selector), function (block) { block.$lineno() != Opal.nil })
```

Discussion: https://github.com/asciidoctor/atom-asciidoc-preview/pull/216#issuecomment-285847507

mojavelinux:


mojavelinux:
With this change, you can write the following code:

```js
var blocks = doc.findBy({}, function(b) { return b.$lineno() !== Opal.nil });
```

I can write a test for it if you let me know where it should go.

mojavelinux:
I thought about the idea of allowing the block to be the first and only argument. That would be consistent with what is allowed in the Ruby API.

```js
var blocks = doc.findBy(function(b) { return b.$lineno() !== Opal.nil });
```

We just have to check if the "selector" argument is a function first.

Mogztter:
> I can write a test for it if you let me know where it should go.

You can add a test in here (where we are testing blocks): https://github.com/asciidoctor/asciidoctor.js/blob/d058481b193d43ebe54f62ec99ae2d496efc15e0/spec/node/asciidoctor.spec.js#L60

> I thought about the idea of allowing the block to be the first and only argument. That would be consistent with what is allowed in the Ruby API.

:+1: 

> We just have to check if the "selector" argument is a function first.

Indeed.

src/asciidoctor-core-api.js
```diff
@@ -71,8 +71,13 @@ Opal.Asciidoctor.AbstractBlock.$$proto.delegate = function () {
   return this.$delegate();
 };
 
-Opal.Asciidoctor.AbstractBlock.$$proto.findBy = function (selector) {
-  return this.$find_by(toHash(selector));
+Opal.Asciidoctor.AbstractBlock.$$proto.findBy = function (selector, block) {
+  if (typeof block === 'function') {
+    return Opal.block_send(this, 'find_by', block, toHash(selector));
```
Mogztter: Are you sure that `block` should be the first argument and the rest of the arguments after (ie `selector`) ?

mojavelinux:
It looks like the name of the method and the order of arguments changes from Opal 0.10 to Opal 0.11.

Here's what the code needs to be to work with Opal 0.11 (runtime):

```js
Opal.Asciidoctor.AbstractBlock.$$proto.findBy = function (selector, block) {
  if (typeof block === 'undefined' && typeof selector === 'function') {
    return Opal.send(this, 'find_by', null, selector);
  }
  else if (typeof block === 'function') {
    return Opal.send(this, 'find_by', toHash(selector), block);
  }
  else {
    return this.$find_by(toHash(selector));
  }
};
```

Other than that, it looks good!

Mogztter:
Urgh build is failing:

```
  101:44   error  Missing space before function parentheses  space-before-function-paren
  101:81   error  Missing semicolon                          semi
  107:78   error  Missing space before function parentheses  space-before-function-paren
  107:124  error  Missing semicolon                          semi
```

My bad, I didn't run the linter :disappointed: 

> It looks like the name of the method and the order of arguments changes from Opal 0.10 to Opal 0.11.
Here's what the code needs to be to work with Opal 0.11 (runtime)

I've merged #280 
Branch master is now running Opal 0.11, could you please rebase ?

Mogztter:
Let's continue the discussion here: https://github.com/asciidoctor/asciidoctor.js/pull/286

PR #284, opened 2017-03-13
resolves #43 use character ranges from XRegExp to define *Rx constants

### LINKED ISSUE #43: Integrate XRegExp.js for fully-compliant Unicode support
We are starting to hit the limits of the basic regexp engine in JavaScript and it may be time to integrate XRegExp.js to better align Ruby and JavaScript.

Thus far, we've used preprocessor-style conditional blocks to setup regular expressions for the Ruby and JavaScript environments, respectively. However, the JavaScript regular expressions can only match word characters within the basic Latin range because of the limitations of the regexp implementation. Fortunately, the engine can match ranges of unicode characters, so it's technically possible. XRegExp preprocesses the regular expression and replaces the posix character groups like `\p{Word}` with the full set of character ranges across the Unicode spectrum.

You can find more information about XRegExp on the website:

http://xregexp.com/

This may not be as high priority as I first anticipated. It seems that in Firefox and Chrome, the `[a-zA-Z0-9]` character group is matching like `\p{Word}`. However, this may not be true of all browsers. Some research may be required to find out what is and is not supported.

mojavelinux:
- copy character ranges from XRegExp
- define *Rx constants for Opal / JavaScript environment in Asciidoctor.js


mojavelinux:
Note that this requires a corresponding change in Asciidoctor core. You may want to consider trying to make the build work with either Asciidoctor 1.5.5 or Asciidoctor 1.5.6, just in case you need to do a minor release in the interim. The other option is just to hold this PR until Asciidoctor 1.5.6 is released.

After this PR is merged, I think we should reorganize the imports so that asciidoctor/opal_ext.rb only contains patches for Opal and the rest is in asciidoctor/opal.rb (or required from there).

mojavelinux:
As a side note, this file ends up being an interesting font test ;)

mojavelinux:
We should probably credit XRegExp since I used their Unicode property / category list via the `XRegExp#source` method. It's hard to be 100% sure I got everything right given the sheer size of the list. I'll document how I did it.

From the inspect view in the browser development tools, I used the following:

* P_L resolved from `XRegExp('\\p{L}').source`
* P_Nl resolved from `XRegExp('\\p{Nl}').source`
* P_Nd resolved from `XRegExp('\\p{Nd}').source`
* P_Pc resolved from `XRegExp('\\p{Pc}').source`

In each case, I removed the surrounding square brackets since we already only interested in the list itself.

As I understand it:

* Alpha is build from Lu, Ll, Lt, Lm, Lo, Nl. That's effectively the same as L + Nl. (See http://unicode.org/reports/tr44/#Alphabetic).
* Alnum is Alpha + Nd (numbers with digits).
* According to the Regexp docs in the Ruby language API, Word is Letter + Mark + Number + Connector_Punctuation. However, I verified that Number really means Nl + Nd in this case. I don't think Mark is needed in modern Unicode (since most accented characters are now written combined).

I spot checked a few regular expressions in Ruby to ensure we're matching the same things. I wouldn't be surprised if there are a few mistakes somewhere.

Mogztter:
@mojavelinux Do you think we can create a test case ? I can write one but I need to know what I need to test :wink:

mojavelinux:
We should absolutely write a test case. However, we'll have to keep it disabled until Asciidoctor (core) 1.5.6 is released or you run the build against Asciidoctor master since it requires changes in both projects.

Here's a simple test case:

```
[[bière]]
== Bière

La bière est la vie.
```

If the regular expressions do not have proper Unicode support, then the ID won't be recognized. Currently, when Asciidoctor.js checks for letters and numbers, it's only considering A-Z, a-z and 0-9. This patch adds support for any letter or number as defined by Unicode, when Asciidoctor goes looking for them. Thus, it matches the accented character in the ID.

Another way to test is an accented character in the author line. Something like:

```
= HubPress
Anthonny Quérouil

{doctitle} was written by {firstname} {lastname}.
```

(If the author line can't be parsed, then the names don't get properly separated).

You're looking for the use of ALPHA, ALNUM, or WORD in the regular expressions. It most pertains to IDs.

mojavelinux:
We could, of course, add an example with CJK characters. Just use them where ever you see an accented character in the examples I gave. Something like:

```
[[ビール]]
== ビール
```

Mogztter:
Cool now I know how to write beer in Japanese! ビール 🍺

Mogztter:
Let's continue here: https://github.com/asciidoctor/asciidoctor.js/pull/291

PR #285, opened 2017-03-19
Build master against Asciidoctor core master

Mogztter:
https://github.com/asciidoctor/asciidoctor/pull/2070#issuecomment-287218597

### DIFF
--- a/lib/asciidoctor/opal_ext/string.rb
+++ b/lib/asciidoctor/opal_ext/string.rb
@@ -7,11 +7,13 @@ class String
   end unless method_defined? :limit_bytesize
   alias :limit :limit_bytesize unless method_defined? :limit
 
+  alias :_original_unpack :unpack
+
   def unpack format
     if format == 'C3'
       self[0, 3].bytes.select.with_index {|_, i| i.even? }
     else
-      raise %(undefined method `unpack' for #{inspect}#{self.class})
+      _original_unpack format
     end
-  end unless method_defined? :unpack
+  end
 end
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -26,7 +26,11 @@ function Builder () {
     'node_modules/opal-runtime/src/opal.js',
     'build/asciidoctor-lib.js'
   ];
-  this.asciidoctorCoreVersion = '1.5.5';
+  if (process.env.ASCIIDOCTOR_CORE_VERSION) {
+    this.asciidoctorCoreVersion = process.env.ASCIIDOCTOR_CORE_VERSION;
+  } else {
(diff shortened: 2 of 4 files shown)


mojavelinux:
Yes, I think this would avoid me breaking Asciidoctor.js accidentally. Since you are already downloading Asciidoctor from GitHub, it's just a matter of changing the URL you are fetching from. It might be nice to accept an environment variable. That way, we can use a matrix to get Travis to build against different versions of Asciidoctor core, whether it be a tag or master.

lib/asciidoctor/opal_ext/string.rb
```diff
@@ -7,11 +7,13 @@ def limit_bytesize size
   end unless method_defined? :limit_bytesize
   alias :limit :limit_bytesize unless method_defined? :limit
 
+  alias :_original_unpack :unpack
+
   def unpack format
     if format == 'C3'
-      self[0, 3].bytes.select.with_index {|_, i| i.even? }
+      self[0, 3].bytes.select.with_index {|_, i| i.even? }
```
mojavelinux: You've got a trailing space here.

npm/builder.js
```diff
@@ -26,7 +26,7 @@ function Builder () {
     'node_modules/opal-runtime/src/opal.js',
     'build/asciidoctor-lib.js'
   ];
-  this.asciidoctorCoreVersion = '1.5.5';
+  this.asciidoctorCoreVersion = 'master'; // or v1.5.5 to build against a release
```
mojavelinux: I think that this should read from an environment variable, defaulting to a fixed version (though we can hardcode master for the time being).

Mogztter: That's an excellent idea! I wish i thought about it 😀

mojavelinux:
As part of this change, can we skip the download if Asciidoctor has already been downloaded? I always have to comment out these lines in builder.js a) to avoid a redundant download and b) so I can tweak the Asciidoctor code for testing out ideas.

Mogztter:
> As part of this change, can we skip the download if Asciidoctor has already been downloaded? I always have to comment out these lines in builder.js a) to avoid a redundant download and b) so I can tweak the Asciidoctor code for testing out ideas.

I've added a `SKIP_CLEAN` to skip the clean `build` directory. Also the download and untar task are skipped if the file or the directory already exists:

```
SKIP_CLEAN=1 npm run build  
```
```
SKIP_CLEAN environment variable is true, skipping "clean" task

download dependencies
build/asciidoctor.tar.gz file already exists, skipping "download" task
build/asciidoctor directory already exists, skipping "untar" task

compile core lib
 - compile asciidoctor
...
```

PR #287, opened 2017-03-19
Create a simple inline macro

Mogztter:
* Add test
 * Migrate the extension API to Opal 0.11
 * Showcase a simple inline macro


mojavelinux:
Nice!

PR #290, opened 2017-03-20
Update require structure to conform to the latest version of Asciidoctor

Mogztter:


mojavelinux:
This looks really great. So clean and organized. I like how you found a way to preserve backwards compat too.

Now that we have the node and nashorn support for IO, I'm thinking we should start to remove any duplicate code from opal_ext/file.rb.

Mogztter:
Yes!
opal_ext/file.rb should be the base/default implementation (IO agnostic).
We should create a File implementation for each IO/runtime: one based on
XMLHttpRequest, another one based on Node.js IO, etc...

PR #291, opened 2017-03-20
resolves #43 use character ranges from XRegExp to define *Rx constants

### LINKED ISSUE #43: Integrate XRegExp.js for fully-compliant Unicode support

Mogztter:


mojavelinux:
Looks good to me!

I'll merge the corresponding change in core asap.

PR #293, opened 2017-03-21
fix bug in find-by

mojavelinux:
- wrap selector argument in array
- add test to verify that both function and selector get used at once


Mogztter:
Thanks!

mojavelinux:
:tada:

PR #294, opened 2017-03-21
Throw an IOError on fs.readFileSync

Mogztter:


mojavelinux:
We should try to get IO#binread into opal upstream. Would you agree?

Mogztter:
Indeed, pull request opened: https://github.com/opal/opal/pull/1651

PR #296, opened 2017-03-23
improvements to public API

mojavelinux:
- convert attributes in JSON format to Opal hash
- change casing of URI to Uri in methods
- remove trailing endlines
- remove unused delegate method
- change return value of getAttributes() to JSON object
- add alias getDocumentTitle to $doctitle method
- add alias getRevisionDate to $revdate method

### DIFF
--- a/src/asciidoctor-core-api.js
+++ b/src/asciidoctor-core-api.js
@@ -5,13 +5,24 @@ var toHash = function (object) {
   return object;
 };
 
+var prepareOptions = function (options) {
+  if (options = toHash(options)) {
+    var attrs = options['$[]']('attributes');
+    if (attrs && typeof attrs === 'object' && attrs.constructor.name === 'Object') {
+      options = options.$dup();
+      options['$[]=']('attributes', toHash(attrs));
+    }
+  }
+  return options;
+};
+
 // Asciidoctor API
 
 Opal.Asciidoctor['$$class'].$$proto.convert = function (input, options) {
   if (typeof input === 'object' && input.constructor.name === 'Buffer') {
-    input = input.toString('utf8'); 
+    input = input.toString('utf8');
   }
-  var result = this.$convert(input, toHash(options));
+  var result = this.$convert(input, prepareOptions(options));
   if (result === Opal.nil) {
     return '';
   }
@@ -19,18 +30,18 @@ Opal.Asciidoctor['$$class'].$$proto.convert = function (input, options) {
 };
 
 Opal.Asciidoctor['$$class'].$$proto.convertFile = function (filename, options) {
-  return this.$convert_file(filename, toHash(options));
+  return this.$convert_file(filename, prepareOptions(options));
 };
 
 Opal.Asciidoctor['$$class'].$$proto.load = function (input, options) {
   if (typeof input === 'object' && input.constructor.name === 'Buffer') {
-    input = input.toString('utf8'); 
+    input = input.toString('utf8');
   }
-  return this.$load(input, toHash(options));
+  return this.$load(input, prepareOptions(options));
 };
 
 Opal.Asciidoctor['$$class'].$$proto.loadFile = function (filename, options) {
-  return this.$load_file(filename, toHash(options));
+  return this.$load_file(filename, prepareOptions(options));
 };
 
 // AbstractBlock API
@@ -67,10 +78,6 @@ Opal.Asciidoctor.AbstractBlock.$$proto.convert = function () {
   return this.$convert();
 };
 
-Opal.Asciidoctor.AbstractBlock.$$proto.delegate = function () {
-  return this.$delegate();
-};
-
 Opal.Asciidoctor.AbstractBlock.$$proto.findBy = function (selector, block) {
   if (typeof block === 'undefined' && typeof selector === 'function') {
     return Opal.send(this, 'find_by', null, selector);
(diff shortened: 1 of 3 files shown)


mojavelinux:
Let me know if you need me to split this up to discuss.

src/asciidoctor-core-api.js
```diff
@@ -177,15 +184,15 @@ Opal.Asciidoctor.AbstractNode.$$proto.setOption = function (name) {
   return this.$set_option(name);
 };
 
-Opal.Asciidoctor.AbstractNode.$$proto.getIconURI = function (name) {
+Opal.Asciidoctor.AbstractNode.$$proto.getIconUri = function (name) {
```
Mogztter: The reasoning was to keep acronym capitalized but I'm fine with this change

src/asciidoctor-core-api.js
```diff
@@ -292,11 +299,11 @@ Opal.Asciidoctor.Document.$$proto.setTitle = function (title) {
   return this['$title='](title);
 };
 
-Opal.Asciidoctor.Document.$$proto.getDoctitle = function (options) {
+Opal.Asciidoctor.Document.$$proto.getDoctitle = Opal.Asciidoctor.Document.$$proto.getDocumentTitle = function (options) {
```
Mogztter: :+1:

src/asciidoctor-core-api.js
```diff
@@ -292,11 +299,11 @@ Opal.Asciidoctor.Document.$$proto.setTitle = function (title) {
...
   return this.$doctitle(toHash(options));
 };
 
-Opal.Asciidoctor.Document.$$proto.getRevdate = function () {
+Opal.Asciidoctor.Document.$$proto.getRevdate = Opal.Asciidoctor.Document.$$proto.getRevisionDate = function () {
```
Mogztter: :+1:

Mogztter:
PhantomJS does not support `Object.assign`:

```
PhantomJS 2.1.1 (Linux 0.0.0) CommonJS Loading should load document attributes FAILED
	TypeError: undefined is not a function (evaluating 'Object.assign({}, this.attributes.$$smap)') in /tmp/b5c495e0da5e44a8f41e19766210aff6.browserify (line 35926)
	getAttributes@/tmp/b5c495e0da5e44a8f41e19766210aff6.browserify:35926:23 <- build/asciidoctor.js:35924:0
	/tmp/b5c495e0da5e44a8f41e19766210aff6.browserify:80917:33 <- spec/share/common-spec.js:58:0
```

src/asciidoctor-core-api.js
```diff
@@ -292,14 +326,22 @@ Opal.Asciidoctor.Document.$$proto.setTitle = function (title) {
...
   return this.$revdate();
 };
 
+Opal.Asciidoctor.Document.$$proto.getRevnumber = Opal.Asciidoctor.Document.$$proto.getRevisionNumber = function () {
+  return this.getAttributes().revnumber;
```
Mogztter: @mojavelinux Do you think Revnumber and Revremark should be in the Asciidoctor core API ?

Mogztter:
Why continuous integration does not run on this pull request :thinking:

mojavelinux:
> Why continuous integration does not run on this pull request

A few days ago the webhooks were broken. I think if I force push it will fire again.

src/asciidoctor-core-api.js
```diff
@@ -5,32 +5,43 @@ var toHash = function (object) {
   return object;
 };
 
+var prepareOptions = function (options) {
```
mojavelinux: That is correct. It is now possible to pass the attributes as a Hash. We dig down to grab the value of the attributes key and coerce it if necessary. We're trying to shield the user from the low-level details here.

+1 to a test.

src/asciidoctor-core-api.js
```diff
@@ -67,10 +78,6 @@ Opal.Asciidoctor.AbstractBlock.$$proto.convert = function () {
   return this.$convert();
 };
 
-Opal.Asciidoctor.AbstractBlock.$$proto.delegate = function () {
```
mojavelinux: I honestly have no idea what this method refers to. "delegate" isn't anywhere in Asciidoctor core.

src/asciidoctor-core-api.js
```diff
@@ -90,7 +97,7 @@ Opal.Asciidoctor.AbstractBlock.$$proto.convert = function () {
 // AbstractNode API
 
 Opal.Asciidoctor.AbstractNode.$$proto.getAttributes = function () {
-  return this.attributes;
+  return Object.assign({}, this.attributes.$$smap);
```
mojavelinux: +1 to test.

src/asciidoctor-core-api.js
mojavelinux: I'm trying to align with AsciidoctorJ here. See http://static.javadoc.io/org.asciidoctor/asciidoctorj/1.6.0-alpha.3/org/asciidoctor/ast/ContentNode.html

mojavelinux:
> PhantomJS does not support Object.assign

Oops, I didn't realize that. I was just being lazy...we could just do a manual clone like https://gist.github.com/wycliffepeart/c6520bc1c340fd0c59af5b7cb59ea610 (a bit simpler than the polyfill you linked to).

oncletom:
If it's only for testing purposes, you can also consider using https://qa.polyfill.io/v2/docs/
It solved all my testing issues with Phantom.

Mogztter:
Someone could use Phantomjs as a runtime but this is mostly for testing purposes.

Mogztter:
This looks good to me, thanks 😊

src/asciidoctor-core-api.js
Mogztter: @mojavelinux This proposal had been removed (force pushed), you don't like it ?

mojavelinux:
@Mogztter I was wondering what happened the revnumber change. Somehow, it got lost in the pushes. Let's take it up as a separate issue.

PR #297, opened 2017-04-03
API: Add getLineno/getLineNumber function on block

Mogztter:


spec/node/asciidoctor.spec.js
```diff
@@ -111,6 +111,15 @@ describe('Node.js', function () {
       var abstractSectionBlocks = doc.findBy({'context': 'section'}, function (b) { return b.getTitle() === 'Second Section'; });
       expect(abstractSectionBlocks.length).toBe(1);
     });
+
+    it('should be able to find blocks with line number', function () {
```
ldez: I think you must create 2 tests instead of 1: one by assert.

Mogztter: Here I want to make sure that not all blocks have a line number defined.
I've added two more tests to assert that:

*  line number is defined when sourcemap is enabled 
* line number is undefined when sourcemap is disabled

src/asciidoctor-core-api.js
```diff
@@ -94,6 +94,14 @@ Opal.Asciidoctor.AbstractBlock.$$proto.convert = function () {
   return this.$convert();
 };
 
+Opal.Asciidoctor.AbstractBlock.$$proto.getLineno = Opal.Asciidoctor.AbstractBlock.$$proto.getLineNumber = function () {
```
mojavelinux: I think we should just drop `getLineno`. We only keep `getDoctitle` since "doctitle" is so core to the terminology of AsciiDoc. But there's no such understanding for "lineno" so I think we should just use the formal term, `getLineNumber`.

Mogztter: Agreed and done!

mojavelinux:
This one is ready to go as far as I can tell.

PR #301, opened 2017-04-09
drop duplicate require statement

mojavelinux:
Core already requires "asciidoctor/converter" (and always has), so there's no reason to require it in the Asciidoctor.js build.

### DIFF
--- a/lib/asciidoctor/js/postscript.rb
+++ b/lib/asciidoctor/js/postscript.rb
@@ -1,4 +1,3 @@
-require 'asciidoctor/converter'
 require 'asciidoctor/converter/composite'
 require 'asciidoctor/converter/html5'
 require 'asciidoctor/extensions'
--- a/lib/asciidoctor/opal_ext.rb
+++ b/lib/asciidoctor/opal_ext.rb
@@ -1,3 +1,4 @@
 # For backward compatibility with Asciidoctor <= 1.5.5
 require 'asciidoctor/js'
+require 'asciidoctor/converter'
 require 'asciidoctor/js/postscript'


mojavelinux:
Is it necessary to require "asciidoctor/composite" here? Or can that be done where we require "asciidoctor/template"? The composite converter is not needed for any of the main built-in converters.

lib/asciidoctor/js/postscript.rb
```diff
@@ -1,4 +1,3 @@
-require 'asciidoctor/converter'
```
Mogztter: This line is only needed with Asciidoctor 1.5.5.
https://travis-ci.org/asciidoctor/asciidoctor.js/jobs/220190386

```
/home/travis/build/asciidoctor/asciidoctor.js/node_modules/opal-runtime/src/opal.js:4903
      throw exception;
      ^
Converter: uninitialized constant Asciidoctor::Converter
```

Mogztter:
> Is it necessary to require "asciidoctor/composite" here? Or can that be done where we require "asciidoctor/template"?

I'm not sure, we could try to move it here: https://github.com/asciidoctor/asciidoctor-template.js

lib/asciidoctor/js/postscript.rb
mojavelinux: That seems strange because it's required in 1.5.5. See https://github.com/asciidoctor/asciidoctor/blob/v1.5.5/lib/asciidoctor.rb#L1611

In other words, there's no difference between 1.5.5 and 1.5.6.dev in this case, so I don't understand why there would be a difference in Asciidoctor.js.

mojavelinux: Ah, because of where the require was done for the html5 converter. It was done at the top instead of at the bottom. Now I get it.

mojavelinux: So I just need to move it to the other file. I'll update.

mojavelinux:
I'm pretty sure this is ready to go now. We'll deal with the composite import in a separate PR.

Mogztter:
Yes, all good!
Thanks :)

PR #303, opened 2017-04-12
Include processor extension

Mogztter:
This is a really useful extension, that will return the following content when the file extension is `.foo`:
```
foo
foo
```
/cc @hsablonniere


src/asciidoctor-extensions-api.js
```diff
@@ -137,3 +148,18 @@ BlockProcessor.$$proto.named = function (name) {
...
+ * @module Extensions/IncludeProcessor
+ */
+var IncludeProcessor = Extensions.IncludeProcessor;
+
+/**
+ * @memberof Extensions/IncludeProcessor
+ */
+IncludeProcessor.$$proto.handles = function (block) {
+  // FIXME: Dirty monkey patch !
+  IncludeProcessor.$$proto['$handles?'] = block;
```
Mogztter: This is ugly :see_no_evil:

mojavelinux: I filed an issue, sent a PR, and verified that it fixes your hack. I've been meaning to fix this for a long time. I'm glad that :sun_with_face: has come.

https://github.com/asciidoctor/asciidoctor/pull/2120

Mogztter: 😉 
For the record, I just learned about `$$alloc.prototype`, I don't know if this could work as a workaround but the fix in core is 👍

spec/share/extensions/foo-include.js
```diff
@@ -0,0 +1,13 @@
+Opal.Asciidoctor.$$scope.Extensions.register(function() {
+  this.includeProcessor(function() {
+    var self = this;
+    self.handles(function (target) {
+      return target.endsWith('.foo');
```
Mogztter: We already test the include macro (ie. if `handles` always returns `true`, the tests suite is failing) but you are right, I will add an explicit test next to the "foo-include-ex" test.

spec/node/asciidoctor.spec.js
```diff
@@ -9,11 +9,20 @@ var config = {
   }
 };
 var asciidoctor = require('../../build/asciidoctor.js')(config);
+function asciidoctorVersionGreaterThan (version) {
```
Mogztter: Maybe we should add an API to get the Asciidoctor version ?
Maybe we should return a double to make comparison easier ?

```
1.5.6 = 1.56
1.5.5 = 1.55
```
`.dev` version should have a low precedence than the associated normal version:

`1.5.6.dev = 1.559`

Mogztter:
@mojavelinux does this look good to you ?

spec/node/asciidoctor.spec.js
mojavelinux: Adding an API to get the version would be consistent with AsciidoctorJ.

http://www.javadoc.io/doc/org.asciidoctor/asciidoctorj/1.6.0-alpha.3

Though I'd argue that should be `getAsciidoctorVersion`.

As far as comparison goes, it follows the Gem::Version rules. See http://ruby-doc.org/stdlib-2.4.1/libdoc/rubygems/rdoc/Gem/Version.html#method-i-3C-3D-3E. That comparison will be useful to have somewhere in Opal or Asciidoctor.js anyway.

mojavelinux: Let's follow up on this in a separate issue about getAsciidoctorVersion and perhaps a comparison helper method for comparing gem versions.

mojavelinux: Btw, I tried to adopt the semantic version pattern in asciidoctor core, but rubygems.org rejects it. So it's actually impossible as far as I can tell to use the semantic version pattern (specifically the hyphen) for a gem...unless there's something I missed.

mojavelinux: In other words, these rules are strict: http://ruby-doc.org/stdlib-2.4.1/libdoc/rubygems/rdoc/Gem/Version.html

mojavelinux: Versionomy looks interesting: http://dazuma.github.io/versionomy/

PR #306, opened 2017-04-22
Resolves #298, API: get revision remark, number and info

### LINKED ISSUE #298: Public API: Adds getRevisionRemark and getRevisionNumber methods
* `attributes['revremark']`
 * `attributes['revnumber']`

https://github.com/asciidoctor/asciidoctor/blob/4dbbedea3fb2c4682afac0495a2384c68b5d2507/lib/asciidoctor/parser.rb#L1794-L1811

Mogztter:
* Adds `Document.getRevisionInfo` function
* Adds `Document.getRevisionRemark` function
* Adds `Document.getRevisionNumber` function

Resolves #298

### DIFF
--- a/src/asciidoctor-core-api.js
+++ b/src/asciidoctor-core-api.js
@@ -599,10 +599,11 @@ Document.$$proto.getDoctitle = function (options) {
 };
 
 /**
+ * Get the document revision date from document header (document attribute <code>revdate</code>).
  * @memberof Document
  */
 Document.$$proto.getRevisionDate = function () {
-  return this.$revdate();
+  return this.getAttribute('revdate');
 };
 
 /**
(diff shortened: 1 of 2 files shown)


mojavelinux:
I'm noticing the docs HTML page as part of recent pull requests. I think it should only be generated as part of the tag, just like we do with the dist files. Otherwise, we're back to committing generated files, which seems like a smell to me. Can you pull docs/index.html out of the master branch?

Mogztter:
> I'm noticing the docs HTML page as part of recent pull requests. I think it should only be generated as part of the tag, just like we do with the dist files. Otherwise, we're back to committing generated files, which seems like a smell to me. Can you pull docs/index.html out of the master branch?

I will see what I can do but I don't think it's possible to configure GitHub pages on tags.

Mogztter:
> Should getRevisionInfo() return null/undefined if none of the revision info is set? Or would that cause problems for the user of the API?

My personal taste is to avoid "null". As a developer, I think it's tedious to check for null especially in JavaScript where the value can be `null` or `undefined` (or `Opal.nil` if we are not careful).

I think the convention in JavaScript is to return `undefined` or an empty object. `null` should not be used in an API.

But we should definitely agree and follow a general rule to make the API consistent.

mojavelinux:
> I will see what I can do but I don't think it's possible to configure GitHub pages on tags.

Actually, what i was thinking was a single gh-pages branch (which you are correct in saying that's all GitHub supports) with folders for each released version. If you wanted to publish the master too, then I would use some sort of nightly job that updates that folder.

mojavelinux:
> My personal taste is to avoid "null". 

In that case, perhaps we can add `isRevisionInfo()` which checks to see if any revision attributes are set. It's essentially the equivalent of checking for undefined in each of the fields on RevisionInfo. This might be helpful if someone is looking to see if any revision info is available at all.

mojavelinux:
> I think the convention in JavaScript is to return undefined or an empty object. 

:+1:

Mogztter:
> Actually, what i was thinking was a single gh-pages branch (which you are correct in saying that's all GitHub supports) with a folder for each released version (e.g., v1.5.5-4). If you wanted to publish the master too, then I would use some sort of nightly job that updates that folder.

I've moved the documentation to https://github.com/asciidoctor/asciidoctor.js/tree/gh-pages

> In that case, perhaps we can add isRevisionInfo() which checks to see if any revision attributes are set. It's essentially the equivalent of checking if each of the fields in RevisionInfo are undefined. This might be helpful if someone is looking to see if any revision info is available at all.

`hasRevisionInfo` is more appropriated no ?

mojavelinux:
> I've moved the documentation to https://github.com/asciidoctor/asciidoctor.js/tree/gh-pages

Super!!

> hasRevisionInfo is more appropriated no ?

sure, that works

src/asciidoctor-core-api.js
```diff
@@ -614,6 +615,86 @@ Document.$$proto.getRevdate = function () {
 };
 
 /**
+ * Get the document revision number from document header (document attribute <code>revnumber</code>).
+ * @memberof Document
+ */
+Document.$$proto.getRevisionNumber = function () {
+  return this.getAttribute('revnumber', '');
```
Mogztter: One small concern is that:

```javascript
document.getAttribute('revnumber'); # returns undefined
document.getRevisionNumber(); # returns empty string ''
```

mojavelinux: It should definitely default to undefined, IMO.

Mogztter: You are definitely right, it make more sense :smile:

PR #313, opened 2017-04-28
resolves #312 map the Extensions.create API and test

### LINKED ISSUE #312: Map the Extensions.create API
So far we've mapped the Extensions.register API. That method registers an extension globally, to be used by all invocations of Asciidoctor. There's a complimentary method that creates an extension register, but not define it. That method is Exensions.create (formerly Extensions.build_registry). The result of this method is then passed to the extension_registry option (formerly extensions_registry) to be used for a specific invocation of Asciidoctor.

Like we did for the Extensions.register API, map the Extensions.create API.

mojavelinux:


mojavelinux:
@Mogztter I defined the sample registry inline. You could define this in a separate file and require it...but I wasn't sure if we wanted to add the code necessary to do that so it works across environments. Let me know how you want to proceed.

Mogztter:
>  I defined the sample registry inline. You could define this in a separate file and require it...but I wasn't sure if we wanted to add the code necessary to do that so it works across environments. Let me know how you want to proceed.

It's fine.
Just a quick note that since you are defining the test in `spec/node/asciidoctor.spec.js` it will only run in a Node environment where you can use `require`.

At some point, we will need to test extensions across all environments but we will do it globally.

PR #320, opened 2017-05-04
Resolves #319, API: Tree Processor extension

### LINKED ISSUE #319: Map the API needed to write a tree processor extension
Map the APIs needed to write a tree processor extension.

```javascript
Opal.Asciidoctor.$$scope.Extensions.register(function() {
  var self = this;
  this.treeProcessor(function() {
    self.process(function(doc) {
      return doc;
    });
  });
});
```

A treeprocessor can return a replacement document or null to continue using the same document.

Mogztter:


spec/share/extensions/love-tree-processor.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function (registry) {
```
Mogztter: @hsablonniere this is for you :heart:

src/asciidoctor-extensions-api.js
```diff
@@ -45,14 +48,19 @@ var toBlock = function (block) {
...
+// Alias
+Opal.Asciidoctor.Extensions = Extensions;
+
 /**
  * @memberof Extensions
  */
 Extensions.create = function (name, block) {
   if (typeof name === 'function' && typeof block === 'undefined') {
-    return Opal.send(this, 'build_registry', null, toBlock(name));
+    return Opal.send(this, 'create', null, toBlock(name));
```
Mogztter: @mojavelinux `build_registry` is deprecated so now I'm using `create` and I've also added a case when the block is undefined.

mojavelinux: The reason I mapped to build_registry was so that we can still build against 1.5.5. Are we still doing that?

Mogztter: Yes we are! Travis will punch me in the face :fist_oncoming:

PR #322, opened 2017-05-06
resolves #321, API: Postprocessor extension

### LINKED ISSUE #321: Map the API needed to write a postprocessor extension
Map the APIs needed to write a postprocessor extension.

```ruby
Opal.Asciidoctor.$$scope.Extensions.register(function() {
  var self = this;
  this.postprocessor(function() {
    self.process(function(doc, output) {
      return output.replace(/foo/, 'bar');
    });
  });
});
```

Mogztter:
resolves #321

### DIFF
--- a/spec/node/asciidoctor.spec.js
+++ b/spec/node/asciidoctor.spec.js
@@ -255,7 +255,7 @@ describe('Node.js', function () {
       expect(result).toContain('<strong>:)</strong>');
     });
 
-    it('should be able to process love extension', function () {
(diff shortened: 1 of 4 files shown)


mojavelinux:
Looks good, except the title of the PR doesn't match the code.

Mogztter:
Oups and the commit message is wrong :(
Thanks for your review.

Mogztter:
Fixed!

mojavelinux:
Nice! Good job!

PR #324, opened 2017-05-07
resolves #323, API: Preprocessor extension

### LINKED ISSUE #323: Map the API needed to write a preprocessor extension
Map the APIs needed to write a preprocessor extension.

```javascript
asciidoctor.Extensions.register(function() {
  var self = this;
  this.preprocessor(function() {
    self.process(function(doc, reader) {
      var lines = reader.getLines();
      lines.append("IMPORTANT: This is a draft");
      return reader;
    });
  });
});
```

Mogztter:
resolves #323

### DIFF
--- a/src/asciidoctor-extensions-api.js
+++ b/src/asciidoctor-extensions-api.js
@@ -17,7 +17,7 @@ var toBlock = function (block) {
  *
  * The various extensions participate in AsciiDoc processing as follows:
  *
- * 1. After the source lines are normalized, {Preprocessor}s modify or replace
+ * 1. After the source lines are normalized, {{@link Extensions/Preprocessor}}s modify or replace
  *    the source lines before parsing begins. {{@link Extensions/IncludeProcessor}}s are used to
  *    process include directives for targets which they claim to handle.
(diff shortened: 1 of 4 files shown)


src/asciidoctor-extensions-api.js
```diff
@@ -234,3 +245,10 @@ if (typeof Extensions.TreeProcessor !== 'undefined') {
 TreeProcessor.$$proto.process = function (block) {
   return Opal.send(this, 'process', null, toBlock(block));
 };
+
+/**
+ * @namespace
+ * @module Extensions/Preprocessor
+ */
+// eslint-disable-next-line no-unused-vars
+var Preprocessor = Extensions.Preprocessor;
```
mojavelinux: I don't think this line is needed. There are no special methods on the Preprocessor class.

Mogztter: True but I need it to declare the documentation `@namespace` and `@module` (otherwise I can use `@link`)

mojavelinux: Gotcha. That make sense. :+1:

PR #332, opened 2017-05-31
Map more of the public API

mojavelinux:


src/asciidoctor-core-api.js
```diff
@@ -948,15 +958,45 @@ Title.$$proto.hasSubtitle = function () {
...
+ *
  * @memberof Inline
- * @extends AbstractNode
+ * @returns {string} - returns the converted String text for this Inline node, or undefined if not applicable for this node.
  */
-Inline.$$proto.convert = function () {
-  return this.$convert();
+Inline.$$proto.getText = function () {
+  var text = this.$text();
+  return text == Opal.nil ? undefined : text
```
Mogztter: `975:45  error  Missing semicolon  semi` 😉

Mogztter: For consistency, I think we should use `===` instead of `==`

mojavelinux: Doh! Hubert doesn't like using semi-colons, so I always get tripped up when switching between projects ;)

mojavelinux: Yes, absolutely. That's a typo on my part.

Mogztter: Maybe one day we will remove them but for now I feel safer to have them ⛑

mojavelinux: :)

src/asciidoctor-core-api.js
```diff
@@ -948,15 +958,45 @@ Title.$$proto.hasSubtitle = function () {
...
 var Inline = Opal.Asciidoctor.Inline;
 
 /**
+ * Get the converted String text of this Inline node, if applicable.
+ *
  * @memberof Inline
- * @extends AbstractNode
+ * @returns {string} - returns the converted String text for this Inline node, or undefined if not applicable for this node.
  */
-Inline.$$proto.convert = function () {
```
Mogztter: Is this intended ? Because we are using the method `convert` in the [smiley-inline-macro.js#L15](https://github.com/asciidoctor/asciidoctor.js/blob/c28dfa60de30bf43f4413bd208918fc976b6bdfd/spec/share/extensions/smiley-inline-macro.js#L15).

mojavelinux: Oops, I was moving lines around and took away too many.

The reason I had initially deleted it was because I thought AbstractNode provide convert. As it turns out, it does not, though it probably should. It's something I've noted down to look at in core.

mojavelinux:
:+1:

Mogztter:
Thanks :smile:

PR #336, opened 2017-06-11
Resolves #331, API: docinfo processor extension

### LINKED ISSUE #331: Map the API needed to write a docinfo processor extension
Map the APIs needed to write a docinfo processor extension.

```js
asciidoctor.Extensions.register(function() {
  var self = this;
  this.docinfoProcessor(function() {
    self.atLocation('footer');
    self.process(function(doc) {
      return 'moar footer';
    });
  });
});
```

Mogztter:


mojavelinux:
Solid!

PR #341, opened 2017-06-13
Release on npm for each tag

### LINKED ISSUE #340: Cannot install with npm from git
I tried to install asciidoctor.js from master to benefit from the latest commits.
However, I ended up having only 3 files on my disk:

```
$ npm install https://github.com/asciidoctor/asciidoctor.js.git
$ ls node_modules/asciidoctor.js
LICENSE      README.adoc  package.json
```

I understood (and learned about) `package.json#files`:

```
  "files": [
    "dist/asciidoctor.js",
    "dist/extensions",
    "dist/css",
    "LICENSE",
    "README.adoc"
  ]
```

The problem being with git, the `dist` folder is not versioned.

Can I suggest `master` to be deployed on npm with a special tag?
Cf. https://docs.npmjs.com/cli/publish

```
$ npm publish --tag dev

# on a different machine

$ npm install asciidoctor.js@dev
```

I am happy to create a PR to automate the publishing on npm from TravisCI.

oncletom:
On a git tag push, a new version will be published on npm under the npm dist-tag `latest` (default behaviour).

On a git push on `master`, a new version will be published in npm as version `<package version>+git.<git-tag>` under the npm dist-tag `dev`.

----

Things to do in `.travis.yml` file:

- [x] make sure the `condition` value is correct (it is currently set to release if build is done with asciidoctor 1.5.5)
- [x] define `NPM_USER_EMAIL` variable in [Travis CI GUI][travis-gui] with the value of: `npm whoami`
- [x] define `NPM_USER_API_KEY` variable in [Travis CI GUI][travis-gui] with the value of:`cat ~/.npmrc | grep ':_authToken=' | awk -F'=' '{print $2}'`

fix #340

[travis-gui]: https://travis-ci.org/asciidoctor/asciidoctor.js/settings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,7 @@ sudo: false
 before_deploy:
 - npm run build
 - npm run dist
-- npm config set git-tag-version false
-- if [[ "$TRAVIS_BRANCH" == master ]]; then npm version "$(npm view . version)+git.$(git log -n1 --pretty=format:%h)"; fi
+- if [[ "$TRAVIS_BRANCH" == master ]]; then npm --no-git-tag-version version "$(npm view . version)+git.$(git log -n1 --pretty=format:%h)"; fi
 - if [[ "$TRAVIS_BRANCH" == master ]]; then npm config set tag dev; fi
 
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,4 @@ deploy:
     tags: true
     repo: asciidoctor/asciidoctor.js
     node: node
+    condition: "$ASCIIDOCTOR_CORE_VERSION -ne 'master'"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,47 @@
 language: node_js
 env:
-  - ASCIIDOCTOR_CORE_VERSION=v1.5.5
-  - ASCIIDOCTOR_CORE_VERSION=master
+- ASCIIDOCTOR_CORE_VERSION=v1.5.5
+- ASCIIDOCTOR_CORE_VERSION=master
 node_js:
-  - "4"
-  - "6"
-  - "node"
-
+- '4'
+- '6'
+- node
 jdk:
-  - oraclejdk8
-
+- oraclejdk8
 before_script:
-  - npm install
-
+- npm install
 script:
- - npm run lint
- - npm run package
- - npm run docs
- - SKIP_BUILD=1 npm run examples
-
+- npm run lint
+- npm run package
+- npm run docs
+- SKIP_BUILD=1 npm run examples
 sudo: false
+
+before_deploy:
+- npm run build
+- npm run dist
+- npm config set git-tag-version false
+- if [[ "$TRAVIS_BRANCH" == master ]]; then npm version "$(npm view . version)+git.$(git log -n1 --pretty=format:%h)"; fi
+- if [[ "$TRAVIS_BRANCH" == master ]]; then npm config set tag dev; fi
+
+deploy:
+- provider: npm
+  skip_cleanup: true
+  email: ascii@doctor.org
+  api_key:
+    secure: OOM/FNDuhWGeZgD5EKzn/Kuv3Nf/uU1rVLoYyJ5/SKo8UeJ95acZyDisG2Y9ZOF1VeFEpmXXfgMxAs5ZKO7GwrGhah9Mvnbex5UQZBiR6Mw2Emq+Zgo+J4qFsTevU280ShQFACw6j5AQM9pFBeg+EcqrD5Q4iloJxG5g0qkUsZQ=
+  on:
+    tags: true
+    repo: asciidoctor/asciidoctor.js
+    condition: "$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"
+    node: node
+- provider: npm
+  skip_cleanup: true
+  email: ascii@doctor.org
+  api_key:
+    secure: OOM/FNDuhWGeZgD5EKzn/Kuv3Nf/uU1rVLoYyJ5/SKo8UeJ95acZyDisG2Y9ZOF1VeFEpmXXfgMxAs5ZKO7GwrGhah9Mvnbex5UQZBiR6Mw2Emq+Zgo+J4qFsTevU280ShQFACw6j5AQM9pFBeg+EcqrD5Q4iloJxG5g0qkUsZQ=
+  on:
+    branch: master
+    repo: asciidoctor/asciidoctor.js
+    condition: "$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"
+    node: node
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,8 @@ before_deploy:
 deploy:
 - provider: npm
   skip_cleanup: true
-  email: ascii@doctor.org
-  api_key:
-    secure: OOM/FNDuhWGeZgD5EKzn/Kuv3Nf/uU1rVLoYyJ5/SKo8UeJ95acZyDisG2Y9ZOF1VeFEpmXXfgMxAs5ZKO7GwrGhah9Mvnbex5UQZBiR6Mw2Emq+Zgo+J4qFsTevU280ShQFACw6j5AQM9pFBeg+EcqrD5Q4iloJxG5g0qkUsZQ=
+  email: $NPM_USER_EMAIL
+  api_key: $NPM_USER_API_KEY
   on:
     tags: true
     repo: asciidoctor/asciidoctor.js
@@ -34,9 +33,8 @@ deploy:
     node: node
 - provider: npm
   skip_cleanup: true
-  email: ascii@doctor.org
-  api_key:
-    secure: OOM/FNDuhWGeZgD5EKzn/Kuv3Nf/uU1rVLoYyJ5/SKo8UeJ95acZyDisG2Y9ZOF1VeFEpmXXfgMxAs5ZKO7GwrGhah9Mvnbex5UQZBiR6Mw2Emq+Zgo+J4qFsTevU280ShQFACw6j5AQM9pFBeg+EcqrD5Q4iloJxG5g0qkUsZQ=
+  email: $NPM_USER_EMAIL
+  api_key: $NPM_USER_API_KEY
   on:
     branch: master
     repo: asciidoctor/asciidoctor.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,4 +27,3 @@ deploy:
     tags: true
     repo: asciidoctor/asciidoctor.js
     node: node
-    condition: "$ASCIIDOCTOR_CORE_VERSION -ne 'master'"
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ script:
 sudo: false
 
 before_deploy:
-- npm run build
 - npm run dist
 - if [[ "$TRAVIS_BRANCH" == master ]]; then npm --no-git-tag-version version "$(npm view . version)+git.$(git log -n1 --pretty=format:%h)"; fi
 - if [[ "$TRAVIS_BRANCH" == master ]]; then npm config set tag dev; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,6 @@ sudo: false
 
 before_deploy:
 - npm run dist
-- if [[ "$TRAVIS_BRANCH" == master ]]; then npm --no-git-tag-version version "$(npm view . version)+git.$(git log -n1 --pretty=format:%h)"; fi
-- if [[ "$TRAVIS_BRANCH" == master ]]; then npm config set tag dev; fi
 
 deploy:
 - provider: npm
@@ -28,14 +26,4 @@ deploy:
   on:
     tags: true
     repo: asciidoctor/asciidoctor.js
-    condition: "$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"
-    node: node
-- provider: npm
-  skip_cleanup: true
-  email: $NPM_USER_EMAIL
-  api_key: $NPM_USER_API_KEY
-  on:
-    branch: master
-    repo: asciidoctor/asciidoctor.js
-    condition: "$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"
     node: node
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ node_js:
 - node
 jdk:
 - oraclejdk8
-before_script:
-- npm install
 script:
 - npm run lint
 - npm run package
--- a/package.json
+++ b/package.json
@@ -35,6 +35,9 @@
     "type": "git",
     "url": "https://github.com/asciidoctor/asciidoctor.js.git"
   },
+  "publishConfig": {
+    "access": "public"
+  },
   "keywords": [
     "asciidoc",
     "asciidoctor",


.travis.yml
```diff
@@ -1,22 +1,47 @@
...
- - SKIP_BUILD=1 npm run examples
-
+- npm run lint
+- npm run package
+- npm run docs
+- SKIP_BUILD=1 npm run examples
 sudo: false
+
+before_deploy:
+- npm run build
```
oncletom: I was not sure if it was necessary to execute this line prior to a `npm run dist`.

.travis.yml
```diff
@@ -1,22 +1,47 @@
...
+- npm run dist
+- npm config set git-tag-version false
```
oncletom: We prevent npm to commit a new git tag if we use `npm version ...`.

.travis.yml
```diff
@@ -1,22 +1,47 @@
...
+- if [[ "$TRAVIS_BRANCH" == master ]]; then npm version "$(npm view . version)+git.$(git log -n1 --pretty=format:%h)"; fi
```
oncletom: We alter the package version if we are pushing from master.

.travis.yml
```diff
@@ -1,22 +1,47 @@
...
+- if [[ "$TRAVIS_BRANCH" == master ]]; then npm config set tag dev; fi
```
oncletom: We alter the default `dist-tag` from `latest` to `dev` if we are pushing from master.

mojavelinux:
I'm 100% in support of this effort.

One thing I'd like to challenge you on is to have the CI tool also create the tag. To ensure the cleanest release, a tag shouldn't really be something that is created locally and pushed. Otherwise, it's still coupled to a developer machine. Ideally, the release script should take master, validate it, then tag and release it. Would you agree?

mojavelinux:
Should we use a dedicated account to release from or should we make it the current maintainer (in this case @Mogztter)?

oncletom:
> One thing I'd like to challenge you on is to have the CI tool also create the tag.

@mojavelinux at the moment of this PR, the npm release is done because some pushed a new `git tag`. The artefact pushed to the npm registry will be done via CI. The maintainer still has control over the semver version they want to apply to the new release.

Is that what you had in mind?

mojavelinux:
I think that's a great first step. My point is that the next step would be to have CI also handle creating the tag. This would be triggered by a special keyword in the commit on master...or by triggering a Travis CI build with a special parameter. It's something I've been working on discussing in the AsciiDoc Jekyll plugin project. See https://github.com/asciidoctor/jekyll-asciidoc/issues/45#issuecomment-302927331

.travis.yml
```diff
@@ -1,22 +1,47 @@
...
+
+deploy:
+- provider: npm
+  skip_cleanup: true
+  email: ascii@doctor.org
+  api_key:
+    secure: OOM/FNDuhWGeZgD5EKzn/Kuv3Nf/uU1rVLoYyJ5/SKo8UeJ95acZyDisG2Y9ZOF1VeFEpmXXfgMxAs5ZKO7GwrGhah9Mvnbex5UQZBiR6Mw2Emq+Zgo+J4qFsTevU280ShQFACw6j5AQM9pFBeg+EcqrD5Q4iloJxG5g0qkUsZQ=
```
ldez: Prefer using aTravis env var in the GUI instead of encrypted values.

.travis.yml
```diff
@@ -1,22 +1,47 @@
...
+  on:
+    tags: true
+    repo: asciidoctor/asciidoctor.js
+    condition: "$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"
+    node: node
+- provider: npm
+  skip_cleanup: true
+  email: ascii@doctor.org
```
ldez: Prefer using a Travis env var in GUI.

.travis.yml
```diff
@@ -1,22 +1,47 @@
...
+- npm run build
+- npm run dist
+- npm config set git-tag-version false
+- if [[ "$TRAVIS_BRANCH" == master ]]; then npm version "$(npm view . version)+git.$(git log -n1 --pretty=format:%h)"; fi
+- if [[ "$TRAVIS_BRANCH" == master ]]; then npm config set tag dev; fi
+
+deploy:
+- provider: npm
+  skip_cleanup: true
+  email: ascii@doctor.org
```
ldez: Prefer using a Travis env var in GUI.

.travis.yml
```diff
@@ -1,22 +1,47 @@
...
+  on:
+    tags: true
+    repo: asciidoctor/asciidoctor.js
+    condition: "$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"
+    node: node
+- provider: npm
+  skip_cleanup: true
+  email: ascii@doctor.org
+  api_key:
+    secure: OOM/FNDuhWGeZgD5EKzn/Kuv3Nf/uU1rVLoYyJ5/SKo8UeJ95acZyDisG2Y9ZOF1VeFEpmXXfgMxAs5ZKO7GwrGhah9Mvnbex5UQZBiR6Mw2Emq+Zgo+J4qFsTevU280ShQFACw6j5AQM9pFBeg+EcqrD5Q4iloJxG5g0qkUsZQ=
```
ldez: Prefer using aTravis env var in the GUI instead of encrypted values.

.travis.yml
```diff
@@ -1,22 +1,47 @@
...
+- '6'
+- node
 jdk:
-  - oraclejdk8
-
+- oraclejdk8
 before_script:
-  - npm install
-
+- npm install
```
ldez: It's useless, Travis do this automatically.

.travis.yml
oncletom: Why do you recommend doing in Travis GUI?

oncletom:
@ldez I admit it is a bit overkill to publish master on npm — installing asciidoctor.js as an npm dependency from git master is not trivial. Can you think of another way of proceeding?

.travis.yml
ldez: because we can made changes without commit anything 😃

ldez: It's more easily to manage GitHub API from Travis UI:
- Currently it's your GitHub API, you can revoke this key.
- don't need to use `travis encrypt`

.travis.yml
Mogztter: No it's not necessary since `npm run dist` runs `build`: https://github.com/asciidoctor/asciidoctor.js/blob/07ce5da718e0234f83e6f16362f0fcc6735dd3c2/npm/builder.js#L159

.travis.yml
oncletom: @ldez super! I have proceeded to the changes 🙂

.travis.yml
```diff
@@ -1,22 +1,41 @@
...
+
+deploy:
+- provider: npm
+  skip_cleanup: true
+  email: $NPM_USER_EMAIL
+  api_key: $NPM_USER_API_KEY
+  on:
+    tags: true
+    repo: asciidoctor/asciidoctor.js
+    condition: "$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"
```
Mogztter: I think it's better to check the opposite, `ASCIIDOCTOR_CORE_VERSION` should not be equal to `master`. If it's not possible, `ASCIIDOCTOR_CORE_VERSION` should start with `v`.

.travis.yml
```diff
@@ -1,22 +1,41 @@
...
- - npm run docs
- - SKIP_BUILD=1 npm run examples
-
+- npm run lint
+- npm run package
+- npm run docs
+- SKIP_BUILD=1 npm run examples
 sudo: false
+
+before_deploy:
```
Mogztter: I think everyone agrees that publishing release on tags with Travis is a good idea.
So I think we should move the "publish (master) to npm on every commit" to another pull request.

.travis.yml
```diff
@@ -1,22 +1,41 @@
...
+  email: $NPM_USER_EMAIL
+  api_key: $NPM_USER_API_KEY
+  on:
+    tags: true
+    repo: asciidoctor/asciidoctor.js
+    condition: "$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"
+    node: node
+- provider: npm
+  skip_cleanup: true
+  email: $NPM_USER_EMAIL
```
Mogztter: Do I need to configure this variables on my Travis account ?

.travis.yml
Mogztter: Since we are building Asciidoctor.js with Node version 4, 6 and latest, do we need to add a condition to publish only once every builds are completed ?

.travis.yml
oncletom: Yes please :) Have a look at the PR description, it contains pointers on what to do to achieve this.

.travis.yml
oncletom: Sure, I can create a separate PR to pursue the conversation and avoid blocking the publishing on a new git tag 👍

.travis.yml
oncletom: We should — this is done with the other deploy and not this one, my bad.

Also, should we check `$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'` or `$ASCIIDOCTOR_CORE_VERSION = 'master'`? I understood from your previous comment the later would not be correct.

.travis.yml
ldez: ~~define NPM_USER_API_KEY variable in Travis CI GUI with the value of:travis encrypt $(cat ~/.npmrc | grep ':_authToken=' | awk -F'=' '{print $2}')~~
define NPM_USER_API_KEY variable in Travis CI GUI with the NPM token.

.travis.yml
Mogztter: If we don't want to publish on `master` then we can use the condition `"$ASCIIDOCTOR_CORE_VERSION = 'v1.5.5'"`.

However `v1.5.5` is "static" so every time a new version comes out we need to change it. In order to avoid that, could we use the condition "$ASCIIDOCTOR_CORE_VERSION -ne 'master'" or `"$ASCIIDOCTOR_CORE_VERSION =~ ^v.*"` ?
As far as I understand we can use Bash condition ? https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A)

.travis.yml
Mogztter: Thanks, that would be perfect 👍

.travis.yml
Mogztter: @ldez Where do you find this token ? Can I use the value on my `.npmrc` ?

> When you log into npm via the Command Line Interface (CLI), we create a unique identifier that we give to your computer, which is stored in your .npmrc file.

https://www.npmjs.com/settings/tokens

.travis.yml
Mogztter: Actually I also want to release preview version from Asciidoctor master.
For instance, Travis should be able to release version `v1.5.6-preview.2` based on Asciidoctor master. So I think we should remove this condition and release on tag.

.travis.yml
ldez: @Mogztter yes in the `~/.npmrc` file:
```
//registry.npmjs.org/:_authToken=aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa
```
It's `aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa` in this case.

Mogztter:
I've removed the deploy on master, added `NPM_USER_API_KEY` on the Travis build and removed the condition.

@oncletom @ldez Could you please do a review ? Thanks :wink:

ldez:
I think you need to add:

```json
  "publishConfig": {
    "access": "public"
  },
```
to the `package.json`

> npm assumes that scoped packages are private by default.

https://docs.travis-ci.com/user/deployment/npm/#Troubleshooting-“npm-ERR!-You-need-a-paid-account-to-perform-this-action.”

Mogztter:
@ldez Done, thanks! :+1:

.travis.yml
oncletom: @Mogztter gotcha — if I understand well, `v1.5.6-preview.2` can be deconstructed as `
$ASCIIDOCTOR_CORE_VERSION-preview.2`, correct? If so, we would use this variable to construct the package version (we can't publish the same package version twice).

But I guess this is a discussion to have when I reopen the PR about publishing asciidoctor.js@master to npm :-)

oncletom:
The `publishConfig` is only required to publish _scoped packages_ (aka `@org/package-name`). It won't hurt if it stays but it is not mandatory in our case.

All is good, I guess it can be merged when you want — we should keep an eye on the next tagged release to see if the npm publishing works as expected 🙂

.travis.yml
Mogztter: @oncletom Not exactly, `v1.5.6-preview.2` will be based on Asciidoctor core master. In other words, Asciidoctor.js is built and released with `$ASCIIDOCTOR_CORE_VERSION = master`.

You've added `condition: "$ASCIIDOCTOR_CORE_VERSION -ne 'master'"` but I think we need to remove this condition in order to release preview version (based on Asciidoctor core master) ?

Mogztter:
Thanks @oncletom !
I will try to release a new preview version :wink:

Mogztter:
Actually I don't think this is gonna work because I'm using the following build command to release a new version:

```
npm run release 1.5.6-preview.2
```

Basically this command, do the following:

* Replace version in `package.json`
* `npm run dist`
*  `git add` and `git commit` "Release version ..."
* `git tag`
* `npm publish`
* Remove `dist` folder
* `git add` and `git commit` "Next iteration"

I'm not sure how we can "hook" Travis in this process :thinking:

oncletom:
@Mogztter remove `npm publish`, `npm run dist` and add `git push && git push --tags`. Travis should kick in the release when it receives a new git tag :-)

oncletom:
It could also be simplified as `npm version 1.5.6-preview.2 && git push --all`. It does 99% of what you describe for you :-)

ldez:
`git push --all` seems overkill and with possible side effects.

`npm version prerelease; git push; git push --tags`

ldez:
`npm version prerelease` do : `1.5.6-preview.1` -> `1.5.6-preview.2`

Mogztter:
Yes but I need to generate the `dist` folder before creating a new tag because `master` doesn't versioned the `dist` folder.

I don't know if it's possible but Travis should perform a release on a special commit message and run `npm run release ${VERSION_FROM_COMMIT_MESSAGE}` and then publish to npm.

ldez:
I have missed something?
- npm version -> tag
- Travis build on the tag (generate the `dist`)
- Travis publish on NPM

Mogztter:
Yes, in your case the tag doesn't contain the `dist` folder.
So what is published on npm is not available on a tag.

oncletom:
@Mogztter it will because what will be uploaded to npm is the _build artefact_ after running `npm run dist`. So the npm package will have the dist folder whereas it is not versioned on git :-)

Mogztter:
The npm package will be correct but I also want the `dist` folder to be versioned for each release.

When a user reports an issue (with a stack trace), I want to be able to read the source code from the tag to find the root cause. Otherwise I need to install the npm package locally.

mojavelinux:
> Use commit message as condition is a very bad idea and seems unnecessary.

I tend to prefer that style, but it's definitely subjective. The real goal is to automate the release pipeline. As long as we do that, I'm not too concerned about how you get there.

mojavelinux:
> So the npm package will have the dist folder whereas it is not versioned on git :-)

That makes sense, but it's different than how we decided to set things up, as @Mogztter mentioned. Having the dist files only in the npm package might be okay if this were only an npm package. But Asciidoctor.js is also used outside of npm...and git provides a more universal distribution mechanism for those uses.

It used to be that we needed the dist folder for Bower. However, Bower has sort of faded away. But other tools might still need these distributed files.

mojavelinux:
@Mogztter What you might be able to do is have Travis recreate the tag after you push it. The first tag, which the tool pushes, lacks the dist folder. But then the build can delete it and make it again, this time with the dist folder. The condition on the tag build could be whether a certain file or folder (like the dist folder) is present. If it's there, the build is skipped.

That seems to give us everything we want. It allows the tool to initiate the release (instead of via a commit message), it allows us to use the Travis publisher, and it gives us the dist folder in the tag.

mojavelinux:
Btw, great work everyone! This is really coming together!

.travis.yml
mojavelinux: The best way to think about this is that sometimes the Asciidoctor.js preview version is based on the next version of Asciidoctor core. Then, there's a period when they come back into alignment, but then Asciidoctor.js jumps ahead again.

Stable releases of Asciidoctor.js should always be based on a released version of core.

I know that we have to rethink core versions so this becomes much less esoteric. But we have to discuss that in core first.

ldez:
@Mogztter Publish `dist` content (or a tar/zip) in the Github release, can do the job.
example:
- https://github.com/ldez/synergy-binaries/releases
- https://github.com/ldez/prm/releases

This is very classic way.

I can do that.
WDYT?

oncletom:
GitHub releases is a nice way to get workable artefacts, and it can be easily automated with Travis CI too — https://docs.travis-ci.com/user/deployment/releases/

The major benefit is to have artefacts and git tag linked together, without polluting much the git history. The good thing is we can install npm tarballs from URLs so we could publish early releases here, release them more often, and skip releasing master on npm.

mojavelinux:
I could probably get behind the idea of tarball / zip on GitHub releases. This would be kind of a "web developer friendly" package that doesn't require use of npm...kind of like what Boostrap does.

We should probably put out a call to see if anyone else is using the dist out of the tags and what the use case is. That way, we don't throw anyone for a loop.

ldez:
Tonight, I will do a PR to add GH Releases.

PR #345, opened 2017-06-21
:rocket: Add GitHub Release.

ldez:
Create a tar with the `dist` content and publish in GH release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,52 @@
 language: node_js
+
 env:
 - ASCIIDOCTOR_CORE_VERSION=v1.5.5
 - ASCIIDOCTOR_CORE_VERSION=master
+
 node_js:
 - '4'
 - '6'
 - node
+
 jdk:
 - oraclejdk8
+
 script:
 - npm run lint
 - npm run package
 - npm run docs
 - SKIP_BUILD=1 npm run examples
+
 sudo: false
 
 before_deploy:
-- npm run dist
+- >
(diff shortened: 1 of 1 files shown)


mojavelinux:
Nice!

However, I think we should create a zip instead of (or in addition to) a tar. As much as I hate to say it (and I do), zip is more accessible for the cross-platform web community than tar.

ldez:
See https://github.com/ldez/ajsdeploy/releases

.travis.yml
```diff
@@ -1,29 +1,52 @@
...
+
 sudo: false
 
 before_deploy:
-- npm run dist
+- >
+  if ! [ "$BEFORE_DEPLOY_RUN" ]; then
+    export BEFORE_DEPLOY_RUN=1;
+    npm run dist;
+    mkdir bin
```
Mogztter: Should we run this command in `npm run dist` ?
The main advantage is that we can reproduce the task on a local environment. Wdyt ?

I can take care of this one, in another pull request.

Mogztter:
LGTM

ldez:
@Mogztter you need to add `GITHUB_TOKEN` in Travis GUI 😉

Mogztter:
@ldez Done!

.travis.yml
ldez: 1. it's possible.
2. I don't known if we have real need to reproduce this specific GH step in local.
3. If you want, you commit in my PR

PR #357, opened 2017-07-21
Travis is now taking care of the release

Mogztter:
//cc @oncletom @ldez 

Thanks for mentioning `npm version` ! :tada: 
For the record, `npm version` is doing the following:

* Update the version in `package.json`
* Commit the change
* Tag

PR #360, opened 2017-07-24
Refactor es2015

Mogztter:
Based on https://github.com/asciidoctor/asciidoctor.js/pull/357

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,12 +1,21 @@
 version: '{build}'
+
 skip_tags: true
 skip_commits:
   message: /\[ci skip\]/
+
 clone_depth: 10
+
+# Test against the latest version of this Node.js version
 environment:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -25,6 +25,7 @@ before_deploy:
 - >
   if ! [ "$BEFORE_DEPLOY_RUN" ]; then
     export BEFORE_DEPLOY_RUN=1;
+    export export ASCIIDOCTOR_CORE_VERSION="${ASCIIDOCTOR_CORE_VERSION_PUBLISH}";
```
iliabylich: Double export works, but it doesn't make sense. Is it intentional?

Mogztter: Oups!
No it's not, thanks for your review :+1:

PR #362, opened 2017-07-24
Create an API to get Asciidoctor version

Mogztter:


mojavelinux:
Perhaps we should provide two methods, one to get the version of Asciidoctor.js and another for core? Granted they track each other, but there are sometimes subtle differences in the numbers.

* getVersion() - Asciidoctor.js
* getCoreVersion() - Asciidoctor

wdyt?

Mogztter:
Yes I was thinking about that but I'm not sure how can I get the Asciidoctor.js version (ie. the version defined in `package.json`) in a cross compatible way...

Mogztter:
Since I'm using a [template file](https://github.com/asciidoctor/asciidoctor.js/blob/master/src/template-asciidoctor.js) I could read the version on the `builder.js` script and use it in the template ?

```js
var package = require('./package.json');
console.log(package .version);
```

mojavelinux:
I definitely think it's something that should be baked into the compiled JavaScript.

PR #368, opened 2017-08-13
resolves #367 replace . with [\S\s] in regexp

### LINKED ISSUE #367: Processor fails to convert document containing a certain sequence of characters
Asciidoctor.js 1.5.6-preview.3 fails to convert the following document:

```
const asciidoctor = require('asciidoctor')()
asciidoctor.convert('* a\n* \u2028b')
```

The line separator character (`\u2028`) is causing Asciidoctor.js to go into an infinite loop. Asciidoctor Ruby converts this document without issue.

mojavelinux:
Fixes an infinite loop when a sibling list item contains the line separator character.

### DIFF
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -358,6 +358,10 @@ Builder.prototype.replaceUnsupportedFeatures = function (callback) {
   let data = fs.readFileSync(path, 'utf8');
   log.debug('Replace (g)sub! with (g)sub');
   data = data.replace(/\$send\(([^,]+), '(g?sub)!'/g, '$1 = $send\($1, \'$2\'');
+  // replace dot wildcard with negated line feed in single-line match expressions
+  data = data.replace(/Opal\.const_set\([^']+'[^']+Rx', *\/.+\/\);$/gm, function (m) {
+    return m.replace(/\.([*+])/g, '[^\\n]$1');
+  });
   fs.writeFileSync(path, data, 'utf8');
   callback();
 };
--- a/spec/share/asciidoctor.spec.js
+++ b/spec/share/asciidoctor.spec.js
@@ -376,7 +376,48 @@ var shareSpec = function (testOptions, asciidoctor) {
         var html = asciidoctor.convert('= Document\n\nThis is a simple document.\n\n== Section\n\nCAUTION: This is important!', options);
         expect(html).toContain('<i class="fa icon-caution" title="Caution"></i>');
       });
+    });
 
+    describe('Wildcard character match', function () {
+      it('should replace wildcard with negated line feed', function () {
+        expect(asciidoctor.$$const.UnorderedListRx.source).not.toContain('.*');
+        expect(asciidoctor.$$const.UnorderedListRx.source).toContain('[^\\n]*');
+      });
+
+      it('should match list item in an undered list item that has a trailing line separator', function () {
+        var html = asciidoctor.convert('* a\n* b\u2028\n* c');
+        expect(html).toContain('<li>\n<p>b</p>\n</li>');
+      });
+
+      it('should match line separator in text of list item in an unordered list', function () {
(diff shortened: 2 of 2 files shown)


mojavelinux:
I hate to have to solve it this way, but I'm not sure what else we can do without a change to core. It's really annoying that the JavaScript `.` does not match the same as in Ruby.

Mogztter:
I think this fix could be integrated in Opal. If MRI and Opal behave differently then it should be fixed in Opal: https://github.com/opal/opal/blob/8f182584aa6e2a21214ac3e43cae1c0c8efbbc4d/lib/opal/nodes/literal.rb#L177

Do you have a simple test case with a condition that's true with MRI but false with Opal ?

Mogztter:
And in the meantime we should definitely merge this fix 😉 

Did you find out why it's working with preview-2 but not with preview-3 ? Did we change the RegExp in core to use dots ? Or is it a regression in Opal/Asciidoctor.js ?

mojavelinux:
There's a test in there that demonstrates what it's fixing. However, I'd like to update it to add all the scenarios.

The detailed explanation for why this started failing can be found in the issue #367.

I'm not sure this is something we want Opal to fix, because it's one of many differences in the regexp engine. The more I think about it, the more I realize that it needs to be fixed in our regexp layer. And I actually have another way of fixing it. (One new idea I had is to make a special AnyListRx for Opal that matches like the other ListRx expressions...in other words, solve the other side of the mismatch).

mojavelinux:
We should hold off on the merge until we've settled on the strategy for fixing. I just wanted to get the PR out there so that we have something to discuss.

Mogztter:
Build is failing on version 1.5.5 because, the list is not encapsulated in a `<div class="paragraph">`.
I'm planning to drop 1.5.5 support on master: https://github.com/asciidoctor/asciidoctor.js/pull/376

> We should hold off on the merge until we've settled on the strategy for fixing. I just wanted to get the PR out there so that we have something to discuss.

The fix looks fine to me

> However, I'd like to update it to add all the scenarios.

Could you please elaborate this point ? Can I help on this task ?

Mogztter:
I've rebased your work on top of master, now the build is green :Tara:

mojavelinux:
Great!

I still need to add the additional scenarios. I've been extremely busy these last few weeks though, so I've been slow to get to it. But what I'm working on is wrapping up this week.

mojavelinux:
Well, that month flew by. I'm back looking at this today.

mojavelinux:
PR updated. I'll comment in issue w/ details.

mojavelinux:
@Mogztter wdyt?

Mogztter:
@mojavelinux It looks good to me, I have just one remark on the regular expression (ie. should we use `^\n` or `\S\s`): https://github.com/asciidoctor/asciidoctor.js/issues/367#issuecomment-333819404 

As far as I understand you are leaning toward `^\n` but you also mentioned that we are dealing with single line expressions so we could use `\S\s` to remove the negation ?

mojavelinux:
I'm a bit uncertain about what the right choice is. Technically, `\S\s` would be perfectly safe as we are dealing with single-line expressions. The reason I chose `[^\n]` is because that more closely aligns with the intention of the `.` character. If we have the right tests, it shouldn't matter which one we choose because we know we can prove it works.

Mogztter:
> The reason I chose [^\n] is because that more closely aligns with the intention of the . character. 

You are right:

> /./ - Any character except a newline.

https://ruby-doc.org/core-2.1.1/Regexp.html#class-Regexp-label-Character+Classes

I was trying to avoid negation but in this case it make perfect sense, so :+1:

mojavelinux:
Let's go for it. Merge at will.

Mogztter:
Done 🍻 
Thanks Dan 😃

PR #371, opened 2017-08-16
Adds AbstractBlock.getCaptionedTitle function

Mogztter:


mojavelinux:
I agree we need to do this. That method (and it's original) makes me cringe though because it's one of the areas of AsciiDoc where the terminology is woefully incorrect. A "captioned title" is really a "caption". What we currently use caption for is actually the prefix (which can be further broken down to reference signifier and reference numeral). But for now, we're kind of stuck with it.

Mogztter:
If I understand you correctly, the `caption` method should return the prefix and the title or an object ?

```ruby

class Caption
  attr_reader :title
  attr_reader :reference # signifier or numeral
end
```

@mojavelinux Do you think that Asciidoctor.js should initiate this change or do we stick with a one to one mapping ? (I'm leaning toward the second option because I think Asciidoctor core should lead the way).

mojavelinux:
I wasn't suggesting a change. I was just letting you know what's in my head. The caption stuff is really messy...but we have to address it in core.

PR #372, opened 2017-08-19
Reduce statically-registered extensions scope in tests

Mogztter:
Attempt to fix tests on master: https://github.com/asciidoctor/asciidoctor/commit/3759c0009df3111efb6a811179bdb5ded8db2426

The following declaration does not activate anymore (with https://github.com/asciidoctor/asciidoctor/commit/3759c0009df3111efb6a811179bdb5ded8db2426):
```js
// Extension declaration
module.exports = function (registry) {
  registry.treeProcessor(function () {
    var self = this;
    self.process(function (doc) {
      doc.getBlocks()[0] = self.createBlock(doc, 'paragraph', 'Made with icon:heart[]');
      return doc;
    });
  });
};

// Usage
var registry = asciidoctor.Extensions.create();
require('../share/extensions/love-tree-processor.js')(registry);
var opts = {};
opts[asciidoctorVersionGreaterThan('1.5.5') ? 'extension_registry' : 'extensions_registry'] = registry;

// tree processor is not activated!
var result = asciidoctor.convert(fs.readFileSync(path.resolve(__dirname + '/love-tree-processor-ex.adoc')), opts);
```

With the following declaration, the extension is activated:
```js
// Extension declaration
module.exports = function (asciidoctor) {
  return asciidoctor.Extensions.create(function () {
    this.treeProcessor(function () {
      var self = this;
      self.process(function (doc) {
        doc.getBlocks()[0] = self.createBlock(doc, 'paragraph', 'Made with icon:heart[]');
        return doc;
      });
    });
  });
};

// Usage
var registry = require('../share/extensions/love-tree-processor.js')(asciidoctor);
var opts = {};
opts[asciidoctorVersionGreaterThan('1.5.5') ? 'extension_registry' : 'extensions_registry'] = registry;

// tree processor is activated \o/
var result = asciidoctor.convert(fs.readFileSync(path.resolve(__dirname + '/love-tree-processor-ex.adoc')), opts);
```
commit: https://github.com/asciidoctor/asciidoctor.js/pull/372/commits/a20731275bb688e558df59b973316b19351e0d7f

@mojavelinux Is this intended ? I didn't find a test in Asciidoctor core with the first declaration but it should work right ? `Asciidoctor.Extensions.create()` (without function) should return a new `Registry` and we should be able to register an extension on this registry... ?

### DIFF
--- a/spec/node/asciidoctor.spec.js
+++ b/spec/node/asciidoctor.spec.js
@@ -11,19 +11,15 @@ var config = {
 var asciidoctor = require('../../build/asciidoctor.js')(config);
 function asciidoctorVersionGreaterThan (version) {
   var currentVersion = asciidoctor.getCoreVersion();
-  var currentVersionNumeric = parseInt(currentVersion.replace('.dev', '').replace(/\./g, ''));
+  // ignore the fourth number, keep only major, minor and patch numbers
+  var currentVersionNumeric = parseInt(currentVersion.replace('.dev', '').replace(/\./g, '').substring(0, 3));
   var versionNumeric = version.replace(/\./g, '');
   return currentVersionNumeric > versionNumeric; 
 }
+
 var Opal = require('opal-runtime').Opal; // for testing purpose only
 require('asciidoctor-docbook.js');
 require('asciidoctor-template.js');
-require('../share/extensions/smiley-inline-macro.js');
-require('../share/extensions/shout-block.js');
-if (asciidoctorVersionGreaterThan('1.5.5')) {
-  require('../share/extensions/foo-include.js');
-}
-require('../share/extensions/lorem-block-macro.js');
 var packageJson = require('../../package.json');
 
 var testOptions = {
@@ -257,10 +253,15 @@ describe('Node.js', function () {
     });
 
     it('should be able to process smiley extension', function () {
-      var result = asciidoctor.convert(fs.readFileSync(path.resolve(__dirname + '/smiley-inline-macro-ex.adoc')));
-      expect(result).toContain('<strong>:D</strong>');
-      expect(result).toContain('<strong>;)</strong>');
-      expect(result).toContain('<strong>:)</strong>');
+      try {
+        require('../share/extensions/smiley-inline-macro.js');
+        var result = asciidoctor.convert(fs.readFileSync(path.resolve(__dirname + '/smiley-inline-macro-ex.adoc')));
+        expect(result).toContain('<strong>:D</strong>');
+        expect(result).toContain('<strong>;)</strong>');
+        expect(result).toContain('<strong>:)</strong>');
+      } finally {
+        asciidoctor.Extensions.unregisterAll();
+      }
     });
 
     it('should be able to process love tree processor extension', function () {
@@ -290,14 +291,24 @@ describe('Node.js', function () {
     });
 
     it('should be able to process custom block', function () {
-      var result = asciidoctor.convert(fs.readFileSync(path.resolve(__dirname + '/shout-block-ex.adoc')));
-      expect(result).toContain('<p>SAY IT LOUD.\nSAY IT PROUD.</p>');
+      try {
+        require('../share/extensions/shout-block.js');
+        var result = asciidoctor.convert(fs.readFileSync(path.resolve(__dirname + '/shout-block-ex.adoc')));
+        expect(result).toContain('<p>SAY IT LOUD.\nSAY IT PROUD.</p>');
+      } finally {
+        asciidoctor.Extensions.unregisterAll();
+      }
     });
 
     it('should be able to process custom include processor when target does match', function () {
       if (asciidoctorVersionGreaterThan('1.5.5')) {
-        var result = asciidoctor.convert(fs.readFileSync(path.resolve(__dirname + '/foo-include-ex.adoc')));
-        expect(result).toContain('foo\nfoo');
+        try {
+          require('../share/extensions/foo-include.js');
+          var result = asciidoctor.convert(fs.readFileSync(path.resolve(__dirname + '/foo-include-ex.adoc')));
+          expect(result).toContain('foo\nfoo');
+        } finally {
+          asciidoctor.Extensions.unregisterAll();
+        }
       }
     });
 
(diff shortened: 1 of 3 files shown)


mojavelinux:
Yikes! That shouldn't be the case.

The change in core was to not activate extensions if none were registered. But the registrations should still get triggered. There seems to be a timing issue in the Asciidoctor.js code.

Here's the logic that we're using in core:

https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/extensions.rb#L661-L680

The question is, why is activate returning nil in the JavaScript environment?

mojavelinux:
>  Asciidoctor.Extensions.create() (without function) should return a new Registry and we should be able to register an extension on this registry... ?

Yes, this is always an option. But it appears that the global registration of extensions has broken. The question is, why. Were they being loaded after the registry was activated?

Play around with that return value in the activate method and see if that is causing the change. And if it is, then global extensions were getting loaded too late.

Mogztter:
Alright, thanks for you input.
I will debug this further.

mojavelinux:
Thanks. This is one of those cases where a change helped us realize that were were relying on behavior that perhaps wasn't right. We'll see. Thank goodness for these test runs against upstream. This is the system working!

Mogztter:
> Thank goodness for these test runs against upstream. This is the system working!

Yes the CI is doing a fine job 👍 

> Yes, this is always an option.

I tried in Asciidoctor core and I have the same behavior:

*working.rb*
```ruby
registry = Asciidoctor::Extensions.create do
  tree_processor SampleTreeProcessor
end

doc = document_from_string %(= Document Title\n\ncontent), :extension_registry => registry
refute_nil doc.extensions
```

*failing.rb*
```ruby
registry = Asciidoctor::Extensions.create
registry.tree_processor SampleTreeProcessor

doc = document_from_string %(= Document Title\n\ncontent), :extension_registry => registry
refute_nil doc.extensions
```

I think the problem is that when a new `Registry`is created (without `block`), the `groups` will be empty.

In the `add_document_processor`, we are adding the processor to the corresponding attribute (ie. in this case we are adding a `TreeProcessor` to the `@tree_processor_extensions` attribute) but the `@groups` remains unchanged.

In other word, we have a `Registry` with a `TreeProcessor` registered and an empty `groups`. Therefore the following code return immediately in the `activate` method:

```ruby
def activate document
  return if (ext_groups = Extensions.groups.values + @groups.values).empty?
  # ...
end
```

That make sense because we do not have `groups` to activate but we should still return `self` because we do have registered extensions.

@mojavelinux Tell me if I'm not clear 😀

mojavelinux:
Uh oh. It looks like I messed up the logic. Thanks for doing this research and describing the scenario in detail. That will definitely help me get this corrected. I'll report back once I've given it some more thought.

mojavelinux:
Okay, I see the problem. I had not anticipated this usage:

```ruby
registry = Asciidoctor::Extensions.create
registry.tree_processor SampleTreeProcessor
```

By design, you should always register an extension via a group. Otherwise, there's no way to unregister it. However, since this method is reachable from the public API, I'll update the code and add a test to continue to support this usage. See https://github.com/asciidoctor/asciidoctor/pull/2379

The way the logic works now, if you pass a registry, it will always be activated. Now, the extensions on the document will only be nil if you don't pass a registry and you have no global extensions registered.

wdyt?

mojavelinux:
I wonder if we could some how detect this an register the extension with the default group.

```ruby
registry.tree_processor SampleTreeProcessor
```

Just an idea for the future.

Mogztter:
> By (the original) design, you should always register an extension via a group.

I think the data structure is quite good:

* A `Group` can contains one or more extensions
* A `Registry` can contains one or more `Group`

and allows to do many things:

* Users can define multiple `Registry`
* Users can define "global" extensions
* Users can define the `Registry` they want to use for a specific conversion
* Users can register one or more extensions into a `Group`
* Users can unregister one or more `Group` (or all `Group`) 

@mojavelinux Although, I find it odd that you can call `register` method on `Asciidoctor::Extensions` but not on a `Asciidoctor::Extensions::Registry`.

*ok.rb*
```ruby
# OK
Asciidoctor::Extensions.register :sample do
  tree_processor SampleTreeProcessor
end
```

*ko.rb*
```ruby
# KO
registry.register :sample do
  tree_processor SampleTreeProcessor
end
```

The `register` method also take care of generating a group name if none is provided.
In my opinion, the two code snippets below should be equivalent:

```ruby
registry.register do
  tree_processor SampleTreeProcessor
end

# shortcut of the above code
registry.tree_processor SampleTreeProcessor
```

Also I do think that the `register` method should return a `Group` object so you can add more extensions to the (same) group.

```ruby
group = registry.register do
  tree_processor SampleTreeProcessor
end

group.register do
  block_macro GistBlockMacro
end
# the group now contains `SampleTreeProcessor` and `GistBlockMacro`
```

Continuing on the same idea, the `Group` object should have a `name`, so even if you didn't explicitly defined a `name`, you can still unregister the group:

```ruby
group = registry.register do
  tree_processor SampleTreeProcessor
end

Asciidoctor::Extensions.unregister group.name
```

Though it may be a bit odd, if we want to use the shortcut syntax:

```ruby
# The following code will create two groups with one extension on each group
registry.tree_processor SampleTreeProcessor
registry.block_macro GistBlockMacro
```
```ruby
# The following code will create one group with two extensions
group = registry.tree_processor SampleTreeProcessor
group.block_macro GistBlockMacro # If we want to register another extension to the group using the shortcut syntax, we need to define registration methods on the Group object
```

It's a bit tricky since we can use "registration" methods outside of a "Group".

If we want to stick to a DSL, I think we should not allow to instantiate "object" but only use functions. Having both a DSL and a more "object oriented" API will be challenging 😉

mojavelinux:
>  I find it odd that you can call register method on Asciidoctor::Extensions but not on a Asciidoctor::Extensions::Registry.

This seems like a natural next step in the evolution of extensions. Do you want to file an issue upstream? I like the proposals that you suggested.

We could have the unregister method accept a group in addition to a name.

mojavelinux:
I see you've already filed this issue here: https://github.com/asciidoctor/asciidoctor/issues/2383

PR #377, opened 2017-08-31
How to compile and use a Ruby extension with Asciidoctor.js

Mogztter:


manual.adoc
```diff
@@ -211,5 +211,112 @@ In the example above, we've created two registries:
...
+The first block extension is registered in the `registryA` and will transform the text to uppercase.
 The other one is registered in the `registryB` and will replace `.` by `{nbsp}!!!`.
+
+=== How to compile a Ruby extension to JavaScript (Node.js)
+
+Let's say you want to use the following Asciidoctor extension (written in Ruby) in your JavaScript based application.
+
+.lib/tel-inline-macro.rb
+```ruby
+require 'asciidoctor/extensions' unless RUBY_ENGINE == 'opal' # <1>
```
mojavelinux: As of Asciidoctor 1.5.6, it's no longer necessary to require the extensions library explicitly. It now autoloads.

manual.adoc
```diff
@@ -211,5 +211,112 @@ In the example above, we've created two registries:
...
+
+include ::Asciidoctor
```
mojavelinux: For demo purposes, I don't think we should use this include. Let's just refer to all classes using their absolute name.

(I'm considering taking this out of the extensions lab too).

manual.adoc
```diff
@@ -211,5 +211,112 @@ In the example above, we've created two registries:
...
+Extensions.register do
+  inline_macro do
+    named :tel
+    parse_content_as :text
+    process do |parent, target, attrs|
+      if (text = attrs['text']).empty?
+        text = target
+      end
+      target = %(tel:#{target})
+      (create_anchor parent, text, type: :link, target: target).render
```
mojavelinux: As of 1.5.6, it's no longer necessary to invoke the render method before returning. This future proofs the extension.

manual.adoc
```diff
@@ -211,5 +211,112 @@ In the example above, we've created two registries:
...
+```
+<1> The extension file is in the `lib` directory
+<2> The extension file is named `tel-inline-macro.rb`
+<3> A file named `tel-inline-macro.js` will be written
+
+Then run it:
+
+ node compile.js
+
+The above command will create a file named `tel-inline-macro.js` containing the extension code in JavaScript.
```
mojavelinux: When mentioning a path in the docs, I prefer: `[.path]_tel-inline-macro.js_` as it's more semantic and reduces the amount of monospace in the document.

manual.adoc
```diff
@@ -211,5 +211,112 @@ In the example above, we've created two registries:
...
+Next we create a Node script to convert this document:
+
+.convert.js
+[source,javascript]
+----
+const fs = require('fs');
+const asciidoctor = require('asciidoctor.js')();
+require('./tel-inline-macro.js'); // <1>
+
+const html = asciidoctor.convert(fs.readFileSync('document.adoc', 'utf-8')); // <2>
```
mojavelinux: Perhaps use:

```
const html = asciidoctor.convertFile('document.adoc', { to_file: false })
```

manual.adoc
Mogztter: That's good because this part was confusing :+1:

manual.adoc
Mogztter: I agree.
Is there any difference/advantage using `::Asciidoctor` (top level constant) ?

manual.adoc
Mogztter: Removed

manual.adoc
Mogztter: I'm a huge fan of single backtick but you are right `[.path]_path_` is more semantic :wink:

manual.adoc
Mogztter: Good idea, `require('fs')` is now longer needed.

rahmanusta:
👍

PR #383, opened 2017-09-26
Adds yarn.lock

Mogztter:
* Adds yarn.lock
 * Upgrade Reveal.js to 3.5.0 to be able to use node >= 4

Related: https://github.com/asciidoctor/asciidoctor-reveal.js/issues/151


lib/asciidoctor/js/opal_ext/file.rb
```diff
@@ -139,6 +141,64 @@ def self.read(path)
...
+  # * URLs starting with file://, http:// or https://
+  # * Windows path starting with a drive letter
+  def self.expand_path(path, basedir = nil)
+    if path.start_with?('file://') || path.start_with?('http://') || path.start_with?('https://')
+      return path
+    end
+    if WindowsRootRx.match? (path)
+      return path
+    end
+    # FIXME: The code is duplicated, how to call "super" ?
```
Mogztter: @mojavelinux I tried to use your "technique" but didn't succeed 🤕 

```js
const thisExtension = this,
    Html5Converter = Opal.klass(asciidoctor.$$const.Converter, null, 'Html5Converter', () => {})

Opal.alias(Html5Converter, 'super_inline_anchor', 'inline_anchor')
```

Mogztter:
Tests are failing on Windows because `require('./asciidoctor.js')` does not work anymore:

```
LoadError: can't find file: "asciidoctor" in ["./C:\\projects\\asciidoctor-js/C:\\projects\\asciidoctor-js\\opal", "./C:\\projects\\asciidoctor-js/C:\\projects\\asciidoctor-js\\stdlib", "./C:\\projects\\asciidoctor-js/C:\\projects\\asciidoctor-js", "./C:\\projects\\asciidoctor-js/C:\\projects\\asciidoctor-js\\node_modules\\bestikk-opal-compiler\\node_modules\\opal-compiler\\src\\stdlib", "./C:\\projects\\asciidoctor-js/C:\\projects\\asciidoctor-js\\lib", "./C:\\projects\\asciidoctor-js/C:\\projects\\asciidoctor-js\\build\\asciidoctor\\lib"]
    at C:\projects\asciidoctor-js\build\asciidoctor.js:21633:15
    at C:\projects\asciidoctor-js\build\asciidoctor.js:21634:3
    at Object.<anonymous> (C:\projects\asciidoctor-js\build\browser.spec.all.js:415:46)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at C:\projects\asciidoctor-js\node_modules\jasmine\lib\jasmine.js:93:5
```

I'm pretty sure we already experienced this issue but I can't remember what was the root cause. I don't know if it can be related to the recent changes in `File.expand_path`: https://github.com/opal/opal/commit/838d0700d78c4f06f1513f555937ac30c495ddf2

I need to do some experimentation on a local Windows machine.

Mogztter:
Opal builder is using `File.expand_path` so I need to patch opal.js otherwise the build won't work on Windows 😭 

Next error is `include file not found: c:/dev/asciidoctor.js/c:/dev/asciidoctor.js/spec/share/chapter-01/c:/dev/asciidoctor.js/c:/dev/asciidoctor.js/spec/share/resources/tip-versions.adoc` when using relative include path...

Windows paths are giving me headaches, seriously 🤢

mojavelinux:
I'll do my best to get to this asap (as well as the other issue holding up the release).

mojavelinux:
On an unrelated note, it looks like the build is calling the release task every time it runs:

> Unable to find the remote name of the original repository asciidoctor/asciidoctor.js

Mogztter:
> On an unrelated note, it looks like the build is calling the release task every time it runs:
> Unable to find the remote name of the original repository asciidoctor/asciidoctor.js

I think that this log is printed by the tests suite. I've added tests to test the release task: https://github.com/asciidoctor/asciidoctor.js/blob/master/spec/build/builder.spec.js

But I admit it's a bit confusing 😕

mojavelinux:
Aha! That makes sense. Then nothing to worry about.

mojavelinux:
This logic in Opal is definitely seems wrong. It's not considering the ALT_SEPARATOR char, which is the char that is set to backslash when running on Windows. Within checking for that separator, it won't parse windows paths correctly.

lib/asciidoctor/js/opal_ext/file.rb
mojavelinux: You need to first alias expand_path to super_expand_path. Then you can call the alias method.

mojavelinux:
As far as I can tell, this version of Opal is just broken on Windows. It's messing up paths left and right. I can't even get asciidoctor.js compiled because it's not requiring / appending the source files correctly. This seems to be the same error that AppVeyor is reporting.

Is Opal tested on Windows?

mojavelinux:
> Is Opal tested on Windows?

Answering my question, it does appear to be: https://github.com/opal/opal/blob/master/appveyor.yml

I'm not sure how it's passing the path tests then. expand_path looks wrong to me because it doesn't consider either the Windows root or the ALT_SEPARATOR.

mojavelinux:
For now, I would recommend against upgrading Opal because there is clearly a lot to sort out before we can use this version.

I would separate out the switch to Yarn into a different PR.

package.json
```diff
@@ -80,12 +80,15 @@
...
     "karma-requirejs": "1.1.0",
-    "phantomjs-prebuilt": "^2.1.7",
+    "phantomjs-prebuilt": "^2.1.15",
     "requirejs": "2.1.0",
     "sinon": "2.3.8",
     "watchify": "^3.7.0",
     "webpack": "^1.13.3"
+  },
+  "resolutions": {
+    "reveal.js": "3.5.0"
```
mojavelinux: Why are we coupling Asciidoctor.js core to reveal.js? This seems like it doesn't belong.

mojavelinux:
>  I need to patch opal.js otherwise the build won't work on Windows :sob:

Can you share that patch?

Mogztter:
> I would separate out the switch to Yarn into a different PR.

Sure 👌 

> Can you share that patch?

https://github.com/opal/opal/pull/1702

Tests are missing on this pull request because I need to install Ruby development tools on Windows to be able to run the Opal tests suite.

mojavelinux:
It looks like you are headed in the right direction. We definitely need a working expand_path in Opal on Windows. :+1:

package.json
Mogztter: I'm using `asciidoctor-reveal.js` in the tests suite to check that the integration between Asciidoctor.js and Asciidoctor Reveal.js is working.

Asciidoctor.js is used in a variety of environments and I want to make sure that I don't unintentionally break compatibility. I see it as a "smoke test" but arguably we could have a dedicated project to do that :nerd_face: 

Anyway, Asciidoctor Reveal.js is using Reveal.js 3.3.0 and this version does not allow to use a recent version of Node: `"node": "~4.1.1"` (see https://github.com/hakimel/reveal.js/blob/539e774d31f91676bcc3f75e28168921cd27d819/package.json#L22).
As opposed to `npm`, `yarn` strictly enforces this requirement, so in order to use `yarn`, I had to force the version. In fact, version 3.5.0 now allows to use any version of Node >= 4: `"node": ">=4.0.0"`.

Issue #379, opened 2017-09-04
Include paths resolved differently between v1.5.5-4 and v1.5.6-preview.3


oncletom:
Hello,

I have tried to upgrade from `asciidoctor.js@1.5.5-4` to `asciidoctor.js@v1.5.6-preview.3` but I stumbled on a problem related to include path resolution. I have not made any change to my build, I simply bumped asciidoctor dependency.

I includes files as following:

```adoc
include::../resources/tip-versions.adoc[]
include::../resources/tip-examples.adoc[]
```

On the generated HTML, I then see:

```
Unresolved directive in index.adoc - include::../resources/tip-versions.adoc[] Unresolved directive in index.adoc - include::../resources/tip-examples.adoc[]
```

And in my terminal shell, I see these warnings:

```
asciidoctor: WARNING: index.adoc: line 31: include file not found: /Users/oncletom/workspace/nodebook/chapter-06/resources/tip-versions.adoc
asciidoctor: WARNING: index.adoc: line 32: include file not found: /Users/oncletom/workspace/nodebook/chapter-06/resources/tip-examples.adoc
```

If I revert back to `asciidoctor.js@1.5.5-4` or `asciidoctor.js@1.5.5-5`, it is okay again.

Is it an expected change to do? I seem not to have seen anything in the releases changelog referring to this.

Thanks a lot for your help :-)

Mogztter:
Strange... Can I reproduce this issue if I checkout [oncletom/nodebook](https://github.com/oncletom/nodebook) `master` branch and upgrade Asciidoctor.js to `1.5.6-preview.3` ?

oncletom:
@Mogztter yes exactly 😸  I have tried with `1.5.6-preview.2` and had the same issue.

Mogztter:
Good catch :fish: 
I think I found the root cause. We are now using the "default" method to read file as per: https://github.com/asciidoctor/asciidoctor/commit/ed67065104be760ff66cab8b83a60b8a02742e66#diff-9f52e64185d6eaeea012017fc43a9fa0

Basically it means that we are no longer using a special case for Node (since Opal Node implementation has all the necessary I/O methods).

Unfortunately the `File.expand_path` does not behave exactly the same in MRI and Node:

```ruby
::File.expand_path 'chapter-01/index.adoc'
# "/home/guillaume/workspace/opensource/nodebook/chapter-01/index.adoc"
```


```js
[...]

Issue #382, opened 2017-09-23
File.expand_path produces unexpected result with file protocol


Mogztter:
Elia made some changes to make `File.expand_path` compliant to the Ruby spec: https://github.com/opal/opal/commit/838d0700d78c4f06f1513f555937ac30c495ddf2

Unfortunately it produces regressions on Asciidoctor.js because we are using a `File.expand_path` on a path starting with `file://`, more specifically:

https://github.com/asciidoctor/asciidoctor/blob/d6608c90384ce425df68972c423ed0780371a9f5/lib/asciidoctor/document.rb#L325

**before.js**
```js
Opal.const_get_qualified('::', 'File').$expand_path("file:///path/to/asciidoctor.js"); 
// file:///path/to/asciidoctor.js
```

**after.js**
```js
Opal.const_get_qualified('::', 'File').$expand_path("file:///path/to/asciidoctor.js"); 
// ./file:/path/to/asciidoctor.js
```

**mri.rb**
```rb
p File.expand_path('file:///path/to/asciidoctor.js')
# /path/to/file:/path/to/asciidoctor.js
```

I'm a bit surprised with MRI, I was expecting `file:///` to be considered an absolute path... :neutral_face: 

As a workaround we could define a path without the `file` protocol in the tests but then we will need to fix this issue : https://github.com/asciidoctor/asciidoctor/issues/2375 otherwise local path without protocol cannot be read when running in a browser environment. 

https://github.com/asciidoctor/asciidoctor.js/blob/54bc65cb74c922d8b714c2f16a64f52857091d1e/spec/browser/asciidoctor.spec.js#L5

If we don't want to fix https://github.com/asciidoctor/asciidoctor/issues/2375 right away, we could monkey-patch `::File.expand_path` in Asciidoctor.js to handle path starting with `file:///` ?

@mojavelinux I would like to have your opinion on this issue (short term and long term fix), thanks :stuck_out_tongue_winking_eye:

mojavelinux:
This is being addressed upstream. In the long term, we will support a URI as a base dir. In the short term, we'll have an exception for the xmlhttprequest IO module.

Mogztter:
Resolved in 10c1ca2

PR #384, opened 2017-10-02
Upgrade to the latest version of Opal

Mogztter:
* [x] Upgrade to Opal `0.11.1.dev` https://github.com/opal/opal/commit/fc20415
* [x] Wait for https://github.com/opal/opal/issues/1755 to be fixed
* [x] Synchronize with core to define how we should read files where paths are URI
* [x] Wait for the Asciidoctor 1.5.6.2 release

Ref: #382
Resolved: #379

### DIFF
--- a/lib/asciidoctor/js/opal_ext/file.rb
+++ b/lib/asciidoctor/js/opal_ext/file.rb
@@ -54,7 +54,6 @@ class File
         var chomped  = #{lines.chomp},
             trailing = lines.length != chomped.length,
             splitted = chomped.split(separator);
-
         for (var i = 0, length = splitted.length; i < length; i++) {
           self.lineno += 1;
           if (i < length - 1 || trailing) {
@@ -76,25 +75,27 @@ class File
     File.readlines(@path)
   end
 
-  def self.readlines(path, separator = $/)
-    content = File.read(path)
-    content.split(separator)
-  end
+  class << self
 
-  # TODO use XMLHttpRequest HEAD request unless in local file mode
-  def self.file?(path)
-    true
-  end
+    def readlines(path, separator = $/)
+      content = File.read(path)
+      content.split(separator)
+    end
 
-  def self.readable?(path)
-    true
-  end
+    # TODO use XMLHttpRequest HEAD request unless in local file mode
+    def file?(path)
+      true
+    end
 
-  def self.read(path)
-    # REMIND will be overriden by a specific implementation
-    ''
-  end
+    def readable?(path)
+      true
+    end
 
+    def read(path)
+      # REMIND will be overriden by a specific implementation
+      ''
+    end
+  end
 end
 
 class IO
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
   },
   "homepage": "https://github.com/asciidoctor/asciidoctor.js",
   "dependencies": {
-    "opal-runtime": "0.11.0-integration8"
+    "opal-runtime": "1.0.2"
   },
   "devDependencies": {
     "asciidoctor-docbook.js": "1.5.6-preview.2",
@@ -71,7 +71,6 @@
     "bestikk-fs": "0.1.0",
     "bestikk-jdk-ea": "0.1.1",
     "bestikk-log": "0.1.0",
-    "bestikk-opal-compiler": "0.3.1-integration7",
     "bestikk-uglify": "0.2.1",
     "browserify": "15.2.0",
     "cross-env": "5.1.3",
@@ -85,6 +84,7 @@
     "karma-chrome-launcher": "2.2.0",
     "karma-jasmine": "1.1.1",
     "karma-requirejs": "1.1.0",
+    "opal-compiler": "1.0.2",
     "puppeteer": "1.0.0",
     "requirejs": "2.3.5",
     "sinon": "4.2.2",
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ addons:
   chrome: stable
 
 env:
-- ASCIIDOCTOR_CORE_VERSION=asciidoctor/asciidoctor#v1.5.6.x
+- ASCIIDOCTOR_CORE_VERSION=v1.5.6.2
 - ASCIIDOCTOR_CORE_VERSION=master
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ addons:
   chrome: stable
 
 env:
-- ASCIIDOCTOR_CORE_VERSION=v1.5.6.1
+- ASCIIDOCTOR_CORE_VERSION=asciidoctor/asciidoctor#v1.5.6.x
 - ASCIIDOCTOR_CORE_VERSION=master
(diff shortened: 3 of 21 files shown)


Mogztter:
https://github.com/opal/opal/pull/1702/ should fix the build on AppVeyor

Mogztter:
We need to synchronize with core to define how we should read files in a "browser context" where paths are URI.

mojavelinux:
Understood. It's on my list to follow up.

Mogztter:
@mojavelinux I've monkey-patched Asciidoctor to remove the special case for `xmlhttprequest` in the include processor (as proposed in https://github.com/asciidoctor/asciidoctor/issues/2455).
I think that using the same behavior (with a few exceptions) will be easier to understand for users.

With this change, one important difference is that `allow-uri-read` attribute needs to be defined, otherwise  URI will be resolved as links. Since this is consistent with Asciidoctor (core) we could just emphasize this change and add a link to the documentation: http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#include-files

I've also patched the `is_root` method: Paths starting with `file://`, `http://` or `https://` are considered absolute paths.

The next step will be to able to read content from URI on Node.js. Currently we are using `fs.readFileSync` but this function does not handle URI. So based on the path we should use `fs.readFileSync`, `http.request` or `https.request`.

Mogztter:
Opal 0.11.0 is out! https://github.com/opal/opal/tree/v0.11.0 :tada: 
I will update my pull request to use the latest release :smile:

lib/asciidoctor/js/path_resolver.rb
```diff
@@ -0,0 +1,15 @@
...
+module Asciidoctor
+class PathResolver
+  SLASH = '/'
+  BACKSLASH = '\\'
+  WindowsRootRx = /^[a-zA-Z]:(?:\\|\/)/
+  WebRootRx = /^(file|https?):\/\//
+
+  def is_root? path
+    (path.start_with? SLASH) ||
+      (::JAVASCRIPT_IO_MODULE == 'xmlhttprequest' && (WebRootRx.match? path)) ||
```
Mogztter: Remove when Asciidoctor 1.5.6.2 will be released

mojavelinux: Really I think we should wait on this PR until 1.5.6.2 is available, which could be this week.

Mogztter: Ok

lib/asciidoctor/js/opal_ext/file.rb
```diff
@@ -76,25 +76,87 @@ def readlines
...
 
-  def self.readlines(path, separator = $/)
-    content = File.read(path)
-    content.split(separator)
-  end
+  class << self
+
+    def read(path)
+      case JAVASCRIPT_IO_MODULE
+      when 'node'
```
mojavelinux: We should remove the 'node' case here as this would reintroduce what we removed in master.

lib/asciidoctor/js/opal_ext/file.rb
```diff
@@ -76,25 +76,87 @@ def readlines
...
+            });
+            xhr.overrideMimeType('text/plain');
+            xhr.send();
+          }
+          catch (e) {
+            status = 0;
+          }
+          // assume that no data in local file mode means it doesn't exist
+          if (status == 404 || (status == 0 && data == '')) {
+            throw #{IOError.new `'No such file or directory: ' + path`};
```
mojavelinux: Let's make sure we align this with changes in master so we don't revert the reporting of the reason.

lib/asciidoctor/js/opal_ext/file.rb
mojavelinux: As make sure the case in the constructor doesn't get reintroduced.

lib/asciidoctor/js/opal_ext/file.rb
```diff
@@ -76,25 +76,87 @@ def readlines
...
+    def read(path)
+      # REMIND will be overriden by a specific implementation
+      ''
+    end
 
+    alias :super_expand_path :expand_path
+
+    # Monkey-patch to keep path unchanged on absolute path:
+    # * URLs starting with file://, http:// or https://
+    def expand_path(path, basedir = nil)
```
mojavelinux: I don't think we need this override anymore now that we have taken a different approach.

lib/asciidoctor/js/opal_ext/file.rb
Mogztter: Bad rebase on this file, sorry 😞

lib/asciidoctor/js/opal_ext/file.rb
Mogztter: see above, I messed up 😉

lib/asciidoctor/js/opal_ext/file.rb
Mogztter: I think you're right, I will try to remove it.

Mogztter:
Build is failing because we are building against master (should work) but also against the `1.5.6.1` release.

mojavelinux:
I think we should start building against the 1.5.6.x branch instead of the 1.5.6.1 release.

Mogztter:
>  I think we should start building against the 1.5.6.x branch instead of the 1.5.6.1 release.

@mojavelinux  https://github.com/asciidoctor/asciidoctor/commits/v1.5.6.x branch is not up to date ?

mojavelinux:
It should be. All it's missing right now is https://github.com/asciidoctor/asciidoctor/pull/2613, but I'm ready to merge that in.

mojavelinux:
All the changes are now merged into 1.5.6.x. It's ready for final testing!

mojavelinux:
Are you ready for the 1.5.6.2 release?

mojavelinux:
I think we should target the extension class enhancement for 1.5.7. As soon as we wrap up Asciidoctor.js 1.5.6, I'll start working on getting 1.5.7 core finalized.

Mogztter:
We are now building against the 1.5.6.2 release and master.
@mojavelinux Do you want to do a final review before merging ?

package.json
```diff
@@ -60,7 +60,7 @@
   },
   "homepage": "https://github.com/asciidoctor/asciidoctor.js",
   "dependencies": {
-    "opal-runtime": "0.11.0-integration8"
+    "opal-runtime": "1.0.2"
```
mojavelinux: This makes me very happy to see ;)

mojavelinux:
I vote to proceed!

package.json
Mogztter: It's not really a 1.0.0 even though Opal is considering to release a 1.0.0 after 0.11. But the version was confusing so I've opted to use semantic versioning (with a compatibility matrix table).

mojavelinux: But I like that it's actually based on a commit in the upstream history. That's progress.

Mogztter: Indeed no need to use an integration branch anymore 🎉

PR #393, opened 2017-10-24
Add an include macro support matrix

Mogztter:


Mogztter:
I should probably provide a "generic" tests suite to test all the cases :stuck_out_tongue_closed_eyes:

Mogztter:
@mojavelinux I've added MRI (Asciidoctor Ruby 1.5.6.1) to the [include macro support table](https://github.com/Mogztter/asciidoctor.js/blob/8276c81f48fa67dcfb892bef890abef25c7bb02b/include-support.adoc).
I've also added a bash script to run each test case with the `asciidoctor` command line (available in the `$PATH`).

The goal here is to have at least a consistent behavior between Node.js and MRI. 

As mentioned, if it is a security concern (ie. someone can "fake" the protocol in a file URI) then I don't mind if this feature is only available when the safe mode is set to unsafe as long as it's possible to use protocol in an include path.

I was a bit surprised that `include::https://raw.com/master/file.adoc[]` is working but not `include::file.adoc[]` with a `base_dir` defined to `https://raw.com/master`.

I've opened an issue to be able to include a content from a URI in Node.js: https://github.com/asciidoctor/asciidoctor.js/issues/394

mojavelinux:
> I was a bit surprised that include::https://raw.com/master/file.adoc[] is working but not include::file.adoc[] with a base_dir defined to https://raw.com/master.

I may have broken that when I tweaked the logic recently.

mojavelinux:
A big thumbs up to this document and test.

Only one request. The correct terminology is "include directive", not "include macro". The include mechanism is not a macro. It is a preprocessor directive.

mojavelinux:
:+1:

Issue #314, opened 2017-04-28
Script requires charset to be set to utf-8


mojavelinux:
Since the addition of the extended regular expression character classes, the Asciidoctor.js script now requires that the charset to be set to utf-8. When loading the script in an HTML page, either the charset must be declared in the meta tag:

```html
<meta charset="utf-8">
```

or it must be declared in the script tag:

```html
<script src="asciidoctor.js" charset="utf-8"></script>
```

I'm not sure if this is a good thing or not. The reason we define the character classes using Unicode characters is to save bytes. However, it might be safer if we keep the script charset independent and use escape codes instead. We should decide before releasing 1.5.6.

mojavelinux:
Here's the error if the charset is not properly set:

```
Uncaught SyntaxError: Invalid regular expression: /^([A-Za-z...
```

Mogztter:
> I'm not sure if this is a good thing or not.

Using utf-8 charset is a good thing I guess.

> However, it might be safer if we keep the script charset independent and use escape codes instead.

True... 

> The reason we define the character classes using Unicode characters is to save bytes

In this case, I think we can avoid using Unicode characters because a few more bytes is not a big deal (at least in Asciidoctor.js) compared to a runtime error if this new requirement is not met.

mojavelinux:
I'd definitely like to get this corrected before 1.5.6. Now that we proved it works, I think we should add a utility task to the build that generates the ranges used in the rx.rb file. That way, we can updated it when needed.

Mogztter:
It's already fixed by https://github.com/asciidoctor/asciidoctor.js/issues/43 right ?

As far as I understand, babel-plugin-utf-8-regex is nothing more than a mapping table between UTF-8 RegExp and Unicode ?

> Now that we proved it works, I think we should add a utility task to the build that generates the ranges used in the rx.rb file.

Could you please explain how we can do it ?

mojavelinux:
Even though this can be fixed by setting the charset in the page, it's going to be perceived as an error in Asciidoctor.js. Therefore, I still think it's worth fixing.

I tried to limit the size of the rx.rb by using the non-latin unicode glyphs directly. The downside is that this requires the file be read as UTF-8. What we need to do instead is use the escape sequences (like XRegExp does). The file ends up being larger, but at least we avoid parsing errors.

For example, instead of:

```
À-Ö
```

we use:

```
\u00c0-\u00d6
```

[...]

PR #395, opened 2017-12-10
Replace with escape code

Mogztter:
Ref: https://github.com/asciidoctor/asciidoctor.js/issues/314

At first I thought that the "Letter" category contained every other "Letter" categories but that's not true. For instance `\xDF` is defined in `P_Ll` (lowercase letter) but not in `P_L`... I don't know if this is intended or not...
As a consequence there's a lot of duplications. 

@mojavelinux Do you think it's worth to deduplicate ? I'm asking because it will be harder to resynchronize with https://github.com/slevithan/xregexp/blob/master/src/addons/unicode-categories.js (though I don't think this file will be update too often)


mojavelinux:
Great! I'll have a look asap.

mojavelinux:
> At first I thought that the "Letter" category contained every other "Letter" categories but that's not true.

This confused me as well.

> As a consequence there's a lot of duplications.

This was my main concern. I really would like to make it as small as possible. I think the key here is to keep the script that generates it handy. That way, when we need to update for whatever reason, we just refine the script. And we can continue to make improvements to what it outputs as we find things we missed. In other words, we really need to automate this.

mojavelinux:
Here's the script I used to condense the character references.

```rb
require 'js_regex'
require 'strscan'

CHAR_RX = %r/\\u[A-F0-9]{4}|[A-Za-z0-9]/

js_rx = JsRegex.new /\p{L}/

raw = js_rx.source.gsub /\[|\]/, ''

s = StringScanner.new raw

def normalize_char char
  if char.start_with? '\\u00'
    %(\\x#{char.slice 4, char.length})
  else
    char
  end
end

chars = []

until s.eos?
  chars << (normalize_char s.scan CHAR_RX)
  if (s.peek 1) == '-'
    chars << s.get_byte
    chars << (normalize_char s.scan CHAR_RX)
  end
end

puts chars.join
```

mojavelinux:
Here's the script I used to check for duplicates and gaps in the ranges.

```rb
require 'js_regex'
require 'strscan'

CHAR_RX = %r/\\u[A-F0-9]{4}|[A-Za-z]/

js_rx = JsRegex.new /\p{L}/

raw = js_rx.source.gsub /\[|\]/, ''

s = StringScanner.new raw

def normalize_char char
  if char.start_with? '\\u00'
    %(\\x#{char.slice 4, char.length})
  else
    char
  end
end

def resolve_glyph char
  if char.start_with? '\\u', '\\x'
    [((char.slice 2, char.length).to_i 16)].pack 'U'
  else
    char
  end
end

def resolve_char glyph
  if glyph.ascii_only?
    glyph
  else
    hex = sprintf('%X', glyph.unpack('U')[0])
    if hex.length < 3
      %(\\x#{hex})
    else
      %(\\u#{hex.rjust 4, '0'})
    end
  end
end

def next_glyph glyph
  [((sprintf '%X', glyph.unpack('U')[0] + 1).to_i 16)].pack 'U'
end

glyphs = []
until s.eos?
  glyphs << (curr_glyph = resolve_glyph s.scan CHAR_RX)
  if (s.peek 1) == '-'
    s.get_byte
    stop_glyph = resolve_glyph s.scan CHAR_RX
    until (curr_glyph = next_glyph curr_glyph) == stop_glyph
      glyphs << curr_glyph
    end
    glyphs << stop_glyph
  end
end

glyphs = glyphs.sort.uniq

chars = []

glyph = nil

while (glyph = glyphs.shift)
  chars << (resolve_char glyph)
  if (next_glyph glyph) == glyphs[0]
    glyph = glyphs.shift
    cnt = 0
    until (next_glyph glyph) != glyphs[0]
      cnt += 1
      glyph = glyphs.shift
    end
    chars << '-' if cnt > 0
    chars << (resolve_char glyph)
  end
end

puts chars.join
```

mojavelinux:
I was using the js_regex gem. Unfortunately, I couldn't use it for all the ranges since it's missing Nl and Pc. But those were small enough to inspect manually. Perhaps we could contribute those ranges upstream so we can make use of this gem, which is actually quiet useful.

mojavelinux:
Btw, here are the properties that Ruby recognizes w/ definitions. https://ruby-doc.org/core-2.4.1/Regexp.html#class-Regexp-label-Character+Properties

mojavelinux:
I have an idea for how to shave off 1/3 the size of this string. If we use the 4-digit hex codes, then we can remove the repeating "\u" from the string and restore it using a gsub. Essentially, we can zip it.

mojavelinux:
Here's a sample of that string:

```
0AA00B500BA00C0-00D600D8-00F600F8-02C1
```

Here's how that replacement might work:

```rb
gsub(/([A-F0-9]{4})(?:-([A-F0-9]{4}))?/) { '\\u' + $1 + ($2 ? %(-\\u#$2) : '') }
```

or

```rb
gsub(/([A-F0-9]{4})(?:-([A-F0-9]{4}))?/) { %(\\u#$1#{$2 && "-\\u#$2"}) }
```

mojavelinux:
I'll update the PR with this proposal so we can see how much it improves the size. This may be the perfect compromise to having the codes in ascii yet still keeping the string reasonable in length.

mojavelinux:
(of course, we'll need to add the 'A-Za-z' range manually, but that's pretty easy to see).

mojavelinux:
Here's the script I used to create the packed ranges:

```rb
require 'js_regex'
require 'strscan'

CHAR_RX = %r/\\u[A-F0-9]{4}|\\x[A-F0-9]{2}|[A-Za-z0-9]/

js_rx = JsRegex.new Regexp.new %(\\p{#{ARGV[0] || 'L'}})

raw = js_rx.source.gsub /\[|\]/, ''

s = StringScanner.new raw

def normalize_char char
  if char.start_with? '\\u'
    char.slice 2, char.length
  elsif char.start_with? '\\x'
    '00' + (char.slice 2, char.length)
  else
    char
  end
end

chars = []

until s.eos?
  chars << (normalize_char s.scan CHAR_RX)
  if (s.peek 1) == '-'
    chars << s.get_byte
    chars << (normalize_char s.scan CHAR_RX)
  end
end

puts chars.join
```

mojavelinux:
In the end, the change adds 1.1K, which is not horrible.

Mogztter:
> In the end, the change adds 1.1K, which is not horrible.

Not bad!

> I have an idea for how to shave off 1/3 the size of this string. If we use the 4-digit hex codes, then we can remove the repeating "\u" from the string and restore it using a gsub. Essentially, we can zip it.

I'm not sure about this one, do you think it's worth it ? we are saving 2Ko, which is a good thing, but Opal is more than 700Ko 😝 
However the function `unpack_hex_range` is pretty simple so why not 🤔

mojavelinux:
I actually find it easier to read without all the `\u`s in there.

mojavelinux:
\o/

Btw, those changes I submitted got merged into the js_regex gem and released, so we can use that to recreate these ranges in the future. We may want to note that somewhere.

PR #398, opened 2017-12-19
✨ Add a stub to support URI

Mogztter:
Ref: https://github.com/asciidoctor/asciidoctor/pull/2514


mojavelinux:
This will serve as a stop-gap in the interim so I can merge https://github.com/asciidoctor/asciidoctor/pull/2514. Then we can take it the rest of the way.

Mogztter:
Merged! We can proceed with asciidoctor/asciidoctor#2514.

PR #402, opened 2017-12-25
add getSourceLines and getSource to Block API

mojavelinux:


mojavelinux:
I decided to use the API getSourceLines() instead of getLines() since it's a more accurate name. I may add that alias to core as well.

mojavelinux:
It's also consistent with the method name on Document.

spec/node/asciidoctor.spec.js
```diff
@@ -141,6 +141,8 @@ describe('Node.js', function () {
       expect(blocks[1].getBlocks()[1].getId()).toBe('blockid');
       expect(blocks[1].getBlocks()[1].getStyle()).toBe('quote');
       expect(blocks[1].getBlocks()[1].getAttribute('attribution')).toBe('Abraham Lincoln');
+      expect(blocks[1].getBlocks()[1].getSourceLines()).toEqual(['This is a quote.', 'It has a title, id, and attribution.'])
```
Mogztter: Missing semicolon 😉 
At some point we may switch to https://standardjs.com/ and remove semicolons but in the meantime...

spec/node/asciidoctor.spec.js
```diff
@@ -141,6 +141,8 @@ describe('Node.js', function () {
...
+      expect(blocks[1].getBlocks()[1].getSource()).toBe('This is a quote.\nIt has a title, id, and attribution.')
```
Mogztter: Missing semicolon

spec/node/asciidoctor.spec.js
Mogztter: Missing semicolon

mojavelinux:
Perhaps you could add the linter as part of the test task. I always forgot to run it because it's not linked.

Mogztter:
> Perhaps you could add the linter as part of the test task. I always forgot to run it because it's not linked.

Yes that's an excellent idea. It happened to me too!

mojavelinux:
:+1:

PR #407, opened 2018-01-02
👷 Resolves #405, skip the rebuild phase by default

### LINKED ISSUE #405: Build task should (by default) skip the compilation phase
The compilation phase is relatively slow and the result is often the same (because the code doesn't change between each build).

* If the file already exists and the "rebuild" tag is absent, skip the compilation phase
* If the file doesn't exists, run the compilation phase (even if the "rebuild" tag is absent)
* If the "rebuild" tag is present, run the compilation phase (even if the file already exists)

Mogztter:
The compilation from Asciidoctor core (rebuild) is skipped if, the file `build/asciidoctor-lib.js` already exists and the "rebuild" tag is absent. Otherwise, the rebuild task is run.

Also adds a clean task to clean the `build` directory.

Resolves #405

### DIFF
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -53,12 +53,7 @@ Builder.prototype.build = function (callback) {
   const start = process.hrtime();
 
   async.series([
-    callback => builder.clean(callback), // clean
-    callback => builder.downloadDependencies(callback), // download dependencies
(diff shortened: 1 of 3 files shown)


mojavelinux:
Very nice. This should save a lot of CPU cycles ;)

PR #411, opened 2018-01-17
convert README to Markdown before publishing package

mojavelinux:
- move AsciiDoc README to the side
- convert AsciiDoc README to Markdown
- allow pack to include README.md in package
- restore AsciiDoc README after publishing

### DIFF
--- a/npm/prepublish.js
+++ b/npm/prepublish.js
@@ -1,0 +1,48 @@
+'use strict';
+
+const fs = require('fs');
+
+const RefMacroRx = new RegExp('(image:)?(?:(https?:[^\\[]+)|\{([a-z0-9_-]+)\})\\[(|.*?[^\\\\])\\]', 'g');
+
+function markdownify (asciidoc) {
+  const attrs = asciidoc
+    .split('\n\n')[0]
+    .split('\n')
+    .filter((line) => line.charAt(0) === ':')
+    .reduce((accum, line) => {
+      let [, name, value] = line.match(/^:([^:]+):(?: (.+)|)$/);
+      if (value && ~value.indexOf('{')) value = value.replace(/\{([^}]+)\}/, (_, refname) => accum[refname]);
+      accum[name] = value || '';
+      return accum;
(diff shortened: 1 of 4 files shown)


mojavelinux:
This will restore the README on npmjs. It will also improve the appearance of the README.

We're still using some Markdown formatting in the README, but that just makes it a little easier to convert.

mojavelinux:
While we continue to lobby for AsciiDoc support on npm, we at least want to have something to display. You can see an example here: https://www.npmjs.com/package/bespoke-overview

npm/prepublish.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+const fs = require('fs');
+
+const RefMacroRx = new RegExp('(image:)?(?:(https?:[^\\[]+)|\{([a-z0-9_-]+)\})\\[(|.*?[^\\\\])\\]', 'g');
+
+function markdownify (asciidoc) {
```
Mogztter: It's clever, I think we should share this code as part of a library 😉

mojavelinux: I agree. That would avoid a lot of copy-paste between projects. My only hesitation is that this kind of a quick hack, so putting it on npm might give people the impression that it's a general purpose converter. I suppose we could name it accordingly.

This converter is more comprehensive than the one I used for bespoke-overview since this README is more advanced.

Mogztter: > My only hesitation is that this kind of a quick hack, so putting it on npm might give people the impression that it's a general purpose converter. I suppose we could name it accordingly.

I had thought about it but as you said we can name it accordingly and state in the README that this converter is *not* a general purpose converter. But still I think it can be very useful :+1:

Mogztter:
Looking forward to see it in action!

Mogztter:
The README is (finally) looking gorgeous on npmjs 🎉 🎉 🎉 
https://www.npmjs.com/package/asciidoctor.js

Thanks @mojavelinux 😄

PR #425, opened 2018-02-12
Provide a dedicated JavaScript file for each runtime

### LINKED ISSUE #396: Don't include opal-runtime in npm package
I'd really prefer if opal-runtime wasn't bundled inside the npm package. This is counter to how npm works and it adds unnecessary bloat to the package. Since the package is already including a generated file, I think we can configure the build to exclude opal-runtime for the file that gets included in the npm package.

I can understand bundling opal-runtime inside the distribution file hosted on CDNs, since they don't benefit from the dependency system that node / npm offers.

Mogztter:
Currently we provide a single UMD file with all the specific implementations.
In this pull request, we want to generate a dedicated JavaScript file for each runtime:

- node
- nashorn
- electron
- browser
- umd (for backward compatiblity)
- phantomjs (meh)
- spidermonkey (meh)

Each specific implementation has moved to a dedicated directory.

The idea is to compile a dedicated JavaScript file for each runtime in order to create a dedicated version of Asciidoctor.js. More specifically the browser version of Asciidoctor.js will not contain `require('fs')` and the Node version of Asciidoctor.js will not include `opal-runtime`.

Resolves #396

### DIFF
--- a/lib/asciidoctor/js/opal_ext.rb
+++ b/lib/asciidoctor/js/opal_ext.rb
@@ -1,106 +1,3 @@
-%x(
-  var isNode = typeof process === 'object' && typeof process.versions === 'object' && process.browser != true,
-      isElectron = typeof navigator === 'object' && typeof navigator.userAgent === 'string' && typeof navigator.userAgent.indexOf('Electron') !== -1,
-      isBrowser = typeof window === 'object',
-      isNashorn = typeof Java === 'object' && Java.type,
-      isRhino = typeof java === 'object',
-      isPhantomJS = typeof window === 'object' && typeof window.phantom === 'object',
-      isWebWorker = typeof importScripts === 'function',
-      isSpiderMonkey = typeof JSRuntime === 'object',
-      platform,
-      engine,
-      framework,
-      ioModule;
-
-  // Load common modules
-  Opal.load("pathname");
-  Opal.load("base64");
-
-  if (typeof moduleConfig === 'object' && typeof moduleConfig.runtime === 'object') {
-    var runtime = moduleConfig.runtime;
-    platform = runtime.platform;
-    engine = runtime.engine;
-    framework = runtime.framework;
-    ioModule = runtime.ioModule;
-  }
-
-  if (typeof platform === 'undefined') {
-    // Try to automatically detect the JavaScript platform, engine and framework
-    if (isNode) {
-      platform = platform || 'node';
-      engine = engine || 'v8';
-      if (isElectron) {
-        framework = framework || 'electron';
-      }
-    }
-    else if (isNashorn) {
-      platform = platform || 'java';
-      engine = engine || 'nashorn';
-    }
-    else if (isRhino) {
-      platform = platform || 'java';
-      engine = engine || 'rhino';
-    }
-    else if (isSpiderMonkey) {
-      platform = platform || 'standalone';
-      framework = framework || 'spidermonkey';
-    }
-    else if (isBrowser) {
-      platform = platform || 'browser';
-      if (isPhantomJS) {
-        framework = framework || 'phantomjs';
-      }
-    }
-    // NOTE: WebWorker are not limited to browser
-    if (isWebWorker) {
-      framework = framework || 'webworker';
-    }
-  }
-
-  if (typeof platform === 'undefined') {
-    throw new Error('Unable to automatically detect the JavaScript platform, please configure Asciidoctor.js: `Asciidoctor({runtime: {platform: \'node\'}})`');
-  }
-
-  // Optional information
-  if (typeof framework === 'undefined') {
-    framework = '';
-  }
-  if (typeof engine === 'undefined') {
-    engine = '';
-  }
-
-  // IO Module
-  if (typeof ioModule !== 'undefined') {
-    if (ioModule != 'spidermonkey'
-         && ioModule != 'phantomjs'
-         && ioModule != 'node'
-         && ioModule != 'java_nio'
-         && ioModule != 'xmlhttprequest') {
-      throw new Error('Invalid IO module, `config.ioModule` must be one of: spidermonkey, phantomjs, node, java_nio or xmlhttprequest');
-    }
-  } else {
-    if (framework === 'spidermonkey') {
-      ioModule = 'spidermonkey';
-    } else if (framework === 'phantomjs') {
-      ioModule = 'phantomjs';
-    } else if (platform === 'node') {
-      ioModule = 'node';
-    } else if (engine === 'nashorn') {
-      ioModule = 'java_nio'
-    } else if (platform === 'browser' || typeof XmlHTTPRequest !== 'undefined') {
-      ioModule = 'xmlhttprequest'
-    } else {
-      throw new Error('Unable to automatically detect the IO module, please configure Asciidoctor.js: `Asciidoctor({runtime: {ioModule: \'node\'}})`');
-    }
-  }
-
-)
-
-JAVASCRIPT_IO_MODULE = %x(ioModule)
-JAVASCRIPT_PLATFORM = %x(platform)
-JAVASCRIPT_ENGINE = %x(engine)
-JAVASCRIPT_FRAMEWORK = %x(framework)
-
 require 'asciidoctor/js/opal_ext/file'
 require 'asciidoctor/js/opal_ext/match_data'
 require 'asciidoctor/js/opal_ext/kernel'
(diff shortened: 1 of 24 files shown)


lib/asciidoctor/js/opal_ext.rb
```diff
@@ -1,121 +1,8 @@
...
-  require 'asciidoctor/js/opal_ext/nashorn/dir'
-end
-if JAVASCRIPT_FRAMEWORK == 'electron'
-  require 'asciidoctor/js/opal_ext/electron/io'
-end
-if JAVASCRIPT_PLATFORM == 'node'
-  require 'asciidoctor/js/opal_ext/node/file'
-  `Opal.load("nodejs")`
-end
+require 'asciidoctor/js/opal_ext/env' # dirty hack, the require will be replaced by a specific implementation
```
Mogztter: The goal is to transpile Asciidoctor only once and then to transpile each specific implementation.
So we end up with: 

* `asciidoctor-core.js` which contains `asciidoctor` and `asciidoctor/js`
* `opal-ext-browser.js` which contains `asciidoctor/js/opal_ext/browser`
* `opal-ext-node.js` which contains `asciidoctor/js/opal_ext/node`
* `opal-ext-umd.js` which contains `asciidoctor/js/opal_ext/umd`
* `opal-ext-nashorn.js` which contains `asciidoctor/js/opal_ext/nashorn`
* ...

The tricky part is to require the specific implementation (at the right time).
So my idea was to replace a "fake" require and concat the two files. For instance if we want to generate Asciidoctor.js for the browser environment:

1. Replace `require 'asciidoctor/js/opal_ext/env'` by `require 'asciidoctor/js/opal_ext/browser`
1. Concat `opal-ext-browser.js` with `asciidoctor-core.js`
1. Enjoy 😋 

But I'm not satisfied with this solution... @mojavelinux Do you have a better idea ?

mojavelinux: I agree we don't want this to be a runtime require. What we want is something very similar to browserify or webpack. Instead of a dedicated name, though, it might be better to use some sort of symbol. I've seen this done in Node using something like a `~` prefix or a magic name like `__asciidoctor_env__`.

Mogztter: I'm now using a "magic" code comment instead of a require on an empty file.

> I agree we don't want this to be a runtime require. 

Actually we need a runtime require because Opal is generating a module for each runtime, so we have to require the module. We could also use an auto requirable module... but for now I think it's OK.

Mogztter:
I think we are good to go 😋

lib/asciidoctor/js/opal_ext.rb
mojavelinux: :+1:

mojavelinux:
Excellent. I'm checking it out.

Am I correct in saying this will be targeted for 1.5.6?

src/template-asciidoctor-node.js
```diff
@@ -0,0 +1,24 @@
+Opal = require('opal-runtime').Opal;
```
mojavelinux: shouldn't this line include a var?

```
var Opal = require('opal-runtime').Opal;
```

mojavelinux:
Don't we need to include the new dist files in the dist property in package.json?

src/template-asciidoctor-node.js
Mogztter: Yes, nice catch!
I didn't know but it's actually a "feature" in JavaScript:

> Globals created with var (those created in the program outside of any function) cannot be deleted.
Implied globals created without var (regardless if created inside functions) can be deleted.

https://stackoverflow.com/a/6888838

mojavelinux:
Overall, I'd say this is a lot more sustainable. I'm sure there will be some adjustments as we get used to the new system, but having things in separate files just seems a lot easier to work with.

src/template-asciidoctor-node.js
mojavelinux: I know, it's very weird. In the future, we'll want to switch to `const`, which is even more restrictive..but that's kind of pointless right now given that Opal doesn't use `const` anywhere (because browser support still isn't universal).

Mogztter:
> Excellent. I'm checking it out.
Am I correct in saying this will be targeted for 1.5.6?

You are :100: correct!

> Don't we need to include the new dist files in the dist property in package.json?

I think we should but the documentation is unclear:

> The optional "files" field is an array of file patterns that describes the entries to be included when your package is installed as a dependency. If the files array is omitted, everything except automatically-excluded files will be included in your publish.

I don't understand if the files will be included in the published package but not available when using the package as a dependency (in a Node environment) or if the files won't be published ?
Anyway I think it's safer to include the whole `dist` folder:

> If you name a folder in the array, then it will also include the files inside that folder (unless they would be ignored by another rule in this section.).

https://docs.npmjs.com/files/package.json#files

> Overall, I'd say this is a lot more sustainable. I'm sure there will be some adjustments as we get used to the new system, but having things in separate files just seems a lot easier to work with.

:+1:

src/template-asciidoctor-node.js
Mogztter: I switched to `const`. 
I think it should be possible to use flags in Opal to enable ECMAScript features. For instance if I want to target "ECMAScript 6 environment" then I should use `let` and `const` instead of `var` (where possible).

mojavelinux: I agree 1000%.

mojavelinux:
> I don't understand if the files will be included in the published package but not available when using the package as a dependency (in a Node environment) or if the files won't be published ?

Aha! I discovered this recently.

There are two ways to get files into an npm package. The first way is to specify nothing and npm (specifically `npm pack`) will happily just gobble up everything and include it. You can tell it to exclude certain files by adding an .npmignore file to the project root (just like .gitignore). However, some filenames and filename prefixes are reserved and can never be excluded (such as READMEs and CHANGELOGs). But the game of exclusion quickly becomes tedious. That's where the files property comes in.

If you specify the `files` property, npm (specifically `npm pack`) will only include the files you list (except for those reserved filenames and filename prefixes). In other words, the files property allows you to be explicit about what npm puts into the package.

I now prefer using the `files` property because it is easier to understand and prevents surprises when you add new files to the repository.

mojavelinux:
>  I think it's safer to include the whole dist folder:

If all those files are retrieved by installing the npm package, I would say yes. If there are any files that are not meant to be retrieved that way, then we should exclude those.

Mogztter:
> If all those files are retrieved by installing the npm package, I would say yes. If there are any files that are not meant to be retrieved that way, then we should exclude those.

That's the case because people are now using `npm` or `yarn` to get the package and then use/copy the files.
The package will definitely be bigger because the Asciidoctor.js code is "duplicated" in each file.

mojavelinux:
> The package will definitely be bigger because the Asciidoctor.js code is "duplicated" in each file.

I'm not too concerned about that. What matters is what we load at runtime. In a Docker image, I can remove the files I don't need to reduce the size post installation.

mojavelinux:
> We should probably remove phantomjs and spidermonkey as they add a few megabytes to the package. 

:+1:

oncletom:
> The first way is to specify nothing and npm (specifically npm pack) will happily just gobble up everything and include it. You can tell it to exclude certain files by adding an .npmignore file to the project root (just like .gitignore)

Actually, npm will gobble up everything… which is not encompassed by the `.gitignore` already. The `.npmignore` file is here to let us ignore different things at _versioning_ and _packaging_ time.

mojavelinux:
I left out that .npmignore extends from .gitignore. That's an important point.

Mogztter:
I've removed `phantomjs` and `spidermonkey` and I've also merged `node` and `electron`.
Since we define `dist/node/asciidoctor.js` as the entry point, the Node "flavor" should work in the following environments:

* Node
* Electron
* Browser

I think it's safer since the `browser` hint is just an hint (and not part of the "standard" `package.json` definition).

mojavelinux:
> Since we define dist/node/asciidoctor.js as the entry point, the Node "flavor" should work in the following environments:

I think you lost me on this one. Isn't the whole point to use a different file for the browser and for node? (Or does browser have another meaning in this context?)

Mogztter:
> I think you lost me on this one.

Maybe I'm over-thinking it :see_no_evil: 

> Isn't the whole point to use a different file for the browser and for node? (Or does browser have another meaning in this context?)

Yes it is!
But let's say that I'm using a web bundler to create a single web page application. And of course, I want to use Asciidoctor.js.

So I install the dependency:

```sh
$ npm install asciidoctor.js
```
Then I import the dependency:

```js
import * as Asciidoctor from 'asciidoctor.js';
```

So far, so good!
But the problem is that this web bundler does not recognize the browser hint. So we are effectively using `dist/node/asciidoctor.js` in a browser environment.

Maybe, what we should do is just throw an exception (or log a warning message) if the file `dist/node/asciidoctor.js` is used in a browser environment ?

mojavelinux:
> But the problem is that this web bundler does not recognize the browser hint.

I think we're trying to fix things that are outside the scope of this project. If the web bundler doesn't recognize the hint, then other measures have to be taken or that tool needs to be fixed.

> Maybe, what we should do is just throw an exception (or log a warning message) if the file dist/node/asciidoctor.js is used in a browser environment ?

That's going to lead to trouble. We can't imagine or prepare for every scenario. We have to trust the application developer knows how to use the library correctly.

We should optimize require for node since this is, after all, a node module (it is what users will expect first). So `require('asciidoctor.js')` should load `dist/node/asciidoctor.js`. When targeting a browser environment, either the bundler needs to recognize the browser hint, or the application developer needs to use that path to which the browser hint points `i.e., require('asciidoctor.js/dist/browser/asciidoctor.js')`.

Mogztter:
> I think we're trying to fix things that are outside the scope of this project. If the web bundler doesn't recognize the hint, then other measures have to be taken or that tool needs to be fixed.

You are absolutely right, it's not a good idea. I will revert my changes.
Though we definitely need a documentation page about integration.

> That's going to lead to trouble. We can't imagine or prepare for every scenario. We have to trust the application developer knows how to use the library correctly.

Agreed

Mogztter:
@mojavelinux Should be OK now :wink:

mojavelinux:
Just tested it and c'est top !

PR #432, opened 2018-03-11
AbstractNode#isAttribute should return false if delegate returns Opal.nil

mojavelinux:


Mogztter:
Yesterday I was reading this part of the code and I was like... hmmm `isAttribute` should return a boolean right.... :thinking:

src/asciidoctor-core-api.js
```diff
@@ -359,10 +359,7 @@ AbstractNode.$$proto.getAttribute = function (name, defaultValue, inherit) {
  */
 AbstractNode.$$proto.isAttribute = function (name, expectedValue, inherit) {
   var value = this['$attr?'](name, expectedValue, inherit);
-  if (value === Opal.nil) {
-    return undefined;
-  }
-  return value;
+  return value === Opal.nil ? false : value;
```
Mogztter: Is this behavior consistent with the Asciidoctor core API ? We return a boolean or a value (which could be an instance instance of another type) ?

mojavelinux: Technically, core only returns truthy or falsy values, but I think the public-facing API here can be more precise. We should coerce a truthy value to true too. I'll update.

mojavelinux: Actually, all we need is a test. It's already returning true and false in all instances as it stands.

Mogztter: Perfect then 👌

PR #435, opened 2018-03-12
resolves #431 add AbstractBlock#hasAttribute method

### LINKED ISSUE #431: Add AbstractNode#hasAttribute method to API as a simple attribute check
Unlike `isAttribute`, the `hasAttribute` method would simply check whether the attribute is set (it exists) or not. It may be a more intuitive API for those not looking for the extra value comparison functionality provided by `isAttribute`.

```js
expect(node.hasAttribute('start')).toBe(true)
```

mojavelinux:


mojavelinux:
Updated!

Mogztter:
Thanks... and merged!

PR #437, opened 2018-03-13
remove used file patches for nodejs

mojavelinux:
- logic is now replaced by Opal's built-in nodejs module


mojavelinux:
We'll still get a warning about the binary flag until the opal-runtime is updated. If you want, go ahead and add that to this PR.

Mogztter:
> We'll still get a warning about the binary flag until the opal-runtime is updated. If you want, go ahead and add that to this PR.

I will upgrade the opal-runtime in https://github.com/asciidoctor/asciidoctor.js/pull/384

mojavelinux:
Oops. I realize the title should have said "unused" not "used". Well, either way, we removed them :smiling_imp:

PR #439, opened 2018-03-14
allow ASCIIDOCTOR_CORE_VERSION to be ref in any fork

mojavelinux:


npm/builder.js
```diff
@@ -27,10 +27,16 @@ function Builder () {
     'node_modules/opal-runtime/src/opal.js',
     'build/asciidoctor-lib.js'
   ];
-  if (process.env.ASCIIDOCTOR_CORE_VERSION) {
-    this.asciidoctorCoreVersion = process.env.ASCIIDOCTOR_CORE_VERSION;
+  var asciidoctorRef = process.env.ASCIIDOCTOR_CORE_VERSION;
+  if (asciidoctorRef && ~asciidoctorRef.indexOf('/')) {
```
Mogztter: In my opinion, the tilde operator is uncommon, could you use `includes` instead ? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

npm/builder.js
```diff
@@ -27,10 +27,16 @@ function Builder () {
     'node_modules/opal-runtime/src/opal.js',
     'build/asciidoctor-lib.js'
   ];
-  if (process.env.ASCIIDOCTOR_CORE_VERSION) {
-    this.asciidoctorCoreVersion = process.env.ASCIIDOCTOR_CORE_VERSION;
+  var asciidoctorRef = process.env.ASCIIDOCTOR_CORE_VERSION;
```
Mogztter: You can use `const` and `let` because we are using Node 6+

Mogztter: I think we should add an example in a comment to mention that `ASCIIDOCTOR_CORE_VERSION` can be a reference to a fork (user/project#branch) or a tag (v1.5.6.1).

Mogztter: You can use `const` and `let` because we are using Node 6+

npm/builder.js
Mogztter: In my opinion, the tilde operator is uncommon, could you use `includes` instead ? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

npm/builder.js
Mogztter: I think we should add an example in a comment to mention that `ASCIIDOCTOR_CORE_VERSION` can be a reference to a fork (user/project#branch) or a tag (v1.5.6.1).

npm/builder.js
```diff
@@ -27,10 +27,16 @@ function Builder () {
...
+  if (asciidoctorRef && ~asciidoctorRef.indexOf('/')) {
+    var segments = asciidoctorRef.split('/');
+    this.asciidoctorCoreUser = segments[0];
+    segments = segments[1].split('#');
+    this.asciidoctorCoreRepo = segments[0];
```
Mogztter: `asciidoctorCoreRepo` should always be `asciidoctor` right ?

npm/builder.js
mojavelinux: Yep.

npm/builder.js
mojavelinux: :+1:

npm/builder.js
mojavelinux: If we're going to allow the form user/repo#branch, we should do it properly and not make any assumptions.

mojavelinux:
Updated!

PR #440, opened 2018-03-15
♻ Use modules to compose the build

Mogztter:
I think it's better to have one module per task instead of a single file with all the tasks.
@mojavelinux What do you think ?


mojavelinux:
I agree 1000%. Discrete tasks are essential.

PR #442, opened 2018-03-19
add Processor#createImageBlock method to public extension API

### LINKED ISSUE #364: Map Processor.create_image_block method in the API
https://github.com/asciidoctor/asciidoctor/blob/fba0af1262269a133579fe9ef1b04fb87c76c477/lib/asciidoctor/extensions.rb#L150-L152

mojavelinux:
This method is of particular importance, not only because it cuts down the number of method arguments needed to create an image block, but it enforces that the target attribute is set, automatically computes the alt text from the image name, and stores the generated alt text in the default-alt attribute. It's a lot of bang for the buck.

Resolves #364

### DIFF
--- a/spec/node/asciidoctor.spec.js
+++ b/spec/node/asciidoctor.spec.js
@@ -397,6 +397,21 @@ Content 2`;
       expect(result).toContain('we have liftoff.');
     });
 
+    it('should be able to create an image block from a processor extension', () => {
+      const registry = asciidoctor.Extensions.create(function () {
+        this.blockMacro(function () {
+          this.named('img');
+          this.process((parent, target) => {
+            return this.createImageBlock(parent, { target: target + '.png' });
+          });
+        });
+      });
+      const opts = {};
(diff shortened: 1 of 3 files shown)


spec/node/asciidoctor.spec.js
```diff
@@ -397,6 +397,21 @@ Content 2`;
...
+      const registry = asciidoctor.Extensions.create(function () {
+        this.blockMacro(function () {
+          this.named('img');
+          this.process((parent, target) => {
+            return this.createImageBlock(parent, { target: target + '.png' });
+          });
+        });
+      });
+      const opts = {};
+      opts[asciidoctorVersionGreaterThan('1.5.5') ? 'extension_registry' : 'extensions_registry'] = registry;
```
Mogztter: `master` should only be built against Asciidoctor core 1.5.6+ and `master`. So we should remove this condition and use `extensions_registry`.

But I think it's better to remove this condition globally (in a dedicated pull request)

PR #444, opened 2018-03-19
🐛 skip the rebuild phase by default

Mogztter:
* Use a constant for the target file


mojavelinux:
:+1: 

I'll be honest, when I want it to rebuild asciidoctor.js, I always just nuke the build directory. So in my mind the build should be optimized for quick turnaround and let the full build just be triggered by a clean surface.

PR #446, opened 2018-03-20
fix toHash logic

mojavelinux:
- properly detect Opal hash
- consolidate impl to a single line
- use better variable name in fromHash

### DIFF
--- a/src/asciidoctor-core-api.js
+++ b/src/asciidoctor-core-api.js
@@ -3,10 +3,7 @@
  * @private
  */
 var toHash = function (object) {
-  if (object && !object.smap) {
-    return Opal.hash(object);
-  }
-  return object;
+  return object && !('$$smap' in object) ? Opal.hash(object) : object;
 };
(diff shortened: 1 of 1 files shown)


mojavelinux:
The existing implementation was not properly detecting an Opal hash. As a result, it was always converting to a hash, hence performing redundant logic.

Mogztter:
> The existing implementation was not properly detecting an Opal hash. As a result, it was always converting to a hash, hence performing redundant logic.

Nice catch! I'm wondering if it's possible to provide a test case... :thinking:

mojavelinux:
I thought about that, but I can't see a way unless we exposed the function.

mojavelinux:
In a way, it's already tested in that the tests do use both cases. What isn't tested is redundant work, but I don't actually think that needs to be tested. As long as the behavior is correct, it's "correct".

Mogztter:
I agree!

PR #448, opened 2018-03-20
map public properties consistently

mojavelinux:
- remove $ prefix or unnecessary method call

### DIFF
--- a/src/asciidoctor-core-api.js
+++ b/src/asciidoctor-core-api.js
@@ -317,7 +317,7 @@ Block.$$proto.getSource = function () {
  * @returns {Array} - returns the String {Array} of source lines for this block.
  */
 Block.$$proto.getSourceLines = function () {
-  return this.$lines();
+  return this.lines;
 };
 
 // AbstractNode API
@@ -386,7 +386,7 @@ AbstractNode.$$proto.removeAttribute = function (name) {
  * @returns {Document} - returns the {@link Document} object to which this node belongs.
  */
 AbstractNode.$$proto.getDocument = function () {
-  return this.$document();
+  return this.document;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


mojavelinux:
@Mogztter As it turns out, some of these mapping had fallen out of date. It probably calls for some more tests since some of these couldn't have been working before.

src/asciidoctor-core-api.js
```diff
@@ -323,7 +323,7 @@ Block.$$proto.getSource = function () {
  * @returns {Array} - returns the String {Array} of source lines for this block.
  */
 Block.$$proto.getSourceLines = function () {
-  return this.$lines();
+  return this.lines;
```
Mogztter: I think both are working if the getter/setter is not overridden because we do have a test case on this function: https://github.com/asciidoctor/asciidoctor.js/blob/eac4c2e30b90f3a16661ebb4f87c2502e2c9b357/spec/node/asciidoctor.spec.js#L153

But it's better to use `this.lines` because `lines` is defined as an attribute accessor in Asciidoctor core.

Mogztter:
> It probably calls for some more tests since some of these couldn't have been working before.

Yes, are you working on it ? Otherwise I can take care of it.
We should probably add tests in this pull request to make sure that everything is working.

src/asciidoctor-core-api.js
mojavelinux: The biggest problem was the API was doing different things for different methods. I think we should use the shortest path to properties for best performance, which is the form `this.lines` as you mentioned.

mojavelinux:
Nope, I'm not working on it atm. I need to sign off shortly, so feel free to take the reigns.

Mogztter:
Sorry I didn't had time to work on it 😞

mojavelinux:
Tests on the way!

Mogztter:
Thanks Dan! 👍

PR #450, opened 2018-03-20
resolves #342 map API method to unregister specific extension groups

### LINKED ISSUE #342: Map the API needed to unregister a single extension
https://github.com/asciidoctor/asciidoctor/pull/2265

mojavelinux:
- map API for Extensions.unregister method
- map getGroups(), unregisterAll(), and unregister() API methods on Registry


mojavelinux:
I think I'm still missing getGroups() on Extensions. I'll add an additional commit for that.

It's kind of strange that Extensions and Registry are two separate interfaces. I think core may need some sort of common interface here.

mojavelinux:
Updated!

src/asciidoctor-extensions-api.js
```diff
@@ -87,12 +95,51 @@ Extensions.unregisterAll = function () {
...
+/**
+ * @memberof Extensions/Registry
+ */
+Registry.$$proto.getGroups = Extensions.getGroups;
+
+/**
+ * @memberof Extensions/Registry
+ */
+Registry.$$proto.unregisterAll = function () {
+  this.groups = Opal.hash2();
```
Mogztter: Can we use `Opal.hash` ? I've searched for `Opal.hash2` and found an old GitHub issue: https://github.com/opal/opal/issues/758 :wink:

src/asciidoctor-extensions-api.js
```diff
@@ -87,12 +95,51 @@ Extensions.unregisterAll = function () {
...
+ * NOTE Opal cannot delete an entry from a Hash that is indexed by symbol, so
+ * we have to resort to using low-level operations in this method.
+ *
+ * @memberof Extensions
+ */
+Extensions.unregister = function () {
+  var names = Array.isArray(arguments[0]) ? arguments[0] : arguments;
+  var groups = this.groups;
+  var groupNameIdx = {};
+  for (var i = 0, groupSymbolNames = groups.$$keys, l1 = groupSymbolNames.length; i < l1; i++) {
```
Mogztter: Does it really make a difference to cache the length of the array ?
In theory it should be faster but it introduces a new variable and in my opinion the code is a bit less readable.

I did search on the Internet but nothing conclusive... Also this array will most likely contains less than one hundred elements, so I think we are fine :rocket:

mojavelinux: You're correct the performance difference isn't measurable. This template is just hardwired in my head.

src/asciidoctor-extensions-api.js
mojavelinux: Technically, Opal.hash and Opal.hash2 are both still supported. The difference is that Opal.hash2 doesn't do any processing on the keys. But in this case, it's empty, so we could technically just use Opal.hash_init().

mojavelinux: Nope, Opal.hash() is the right thing to use.

spec/node/asciidoctor.spec.js
```diff
@@ -355,6 +355,121 @@ Content 2`;
...
+      } finally {
+        asciidoctor.Extensions.unregisterAll();
+      }
+    });
+
+    it('should be able to unregister multiple statically-registered extension groups', () => {
+      var extensions = asciidoctor.Extensions;
+      try {
+        extensions.register('test', function () {
+          this.blockMacro(function () {
```
Mogztter: I really like to see the extension API in action :heart_eyes:

PR #451, opened 2018-03-20
map API necessary to retrieve a converter instance from the converter factory

mojavelinux:


mojavelinux:
Admittedly, this is rather low-level, but essential when extending Asciidoctor

spec/share/asciidoctor.spec.js
```diff
@@ -529,6 +529,27 @@ var shareSpec = function (testOptions, asciidoctor) {
         expect(html).toContain('March');
       });
     });
+
+    describe('Converter', function () {
```
Mogztter: Before knowing about the subtleties of the arrow functions I was using them everywhere in the tests suite.
Mocha explicitly discouraged the usage of arrow functions: https://mochajs.org/#arrow-functions

I think we should follow this rule and always use `function` to be consistent.

src/asciidoctor-extensions-api.js
```diff
@@ -299,3 +299,65 @@ var DocinfoProcessor = Extensions.DocinfoProcessor;
...
+var Converter = Opal.const_get_qualified(Opal.Asciidoctor, 'Converter');
+
+// Alias
+Opal.Asciidoctor.Converter = Converter;
+
+/**
+ * Convert the specified node.
+ *
+ * @param {AbstractNode} node - the AbstractNode to convert
+ * @param {string} transform -  An optional String transform that hints at
```
Mogztter: (nitpicking) should we start with a lower case ? and I think there's a trailing space :nerd_face:

src/asciidoctor-extensions-api.js
```diff
@@ -299,3 +299,65 @@ var DocinfoProcessor = Extensions.DocinfoProcessor;
...
+/**
+ * @namespace
+ * @module Converter/Factory
+ */
+var ConverterFactory = Opal.Asciidoctor.Converter.Factory;
+
+/**
+ * Retrieves the singleton instance of the converter factory.
+ *
+ * @param {boolean} initialize - Instantiate the singleton if it has not yet
```
Mogztter: Uppercase or lowercase that's the question... :thinking: :laughing:

spec/share/asciidoctor.spec.js
mojavelinux: I strongly disagree on this one. The arrow functions make JavaScript 1,000 times easier to read. The function should only be used when you need `this`. In fact, it's a very useful way of suppressing the `this` context change.

That recommendation from mocha is out of date. They actually recommend the opposite now.

(of course, until we switch the code style, we shouldn't make any sweeping changes).

src/asciidoctor-extensions-api.js
mojavelinux: lowercase it is!

mojavelinux:
PR updated!

spec/share/asciidoctor.spec.js
Mogztter: I'm kinda relieved because I like the concise syntax of arrow functions :+1: 
Once bitten, twice shy :flushed:

Mogztter:
This looks solid! Thanks :+1:

spec/share/asciidoctor.spec.js
mojavelinux: Be bold! :muscle:

PR #453, opened 2018-03-20
resolves #452 allow class or class instance to be registered as an extension

### LINKED ISSUE #452: Allow a class or class instance to be passed to extension methods
Extension methods in the public API such as blockMacro should accept a class or class instance. Currently, an assumption is made that the processor is a function block. This cases methods such as blockMacro to not behave like their Ruby counterparts (e.g., `$block_macro`).

mojavelinux:


src/asciidoctor-extensions-api.js
```diff
@@ -96,86 +104,86 @@ var Registry = Extensions.Registry;
 /**
  * @memberof Extensions/Registry
  */
-Registry.$$proto.block = function (name, block) {
-  if (typeof name === 'function' && typeof block === 'undefined') {
-    return Opal.send(this, 'block', null, toBlock(name));
-  } else {
-    return Opal.send(this, 'block', [name], toBlock(block));
+Registry.$$proto.block = function (name, processor) {
+  if (arguments.length === 1) {
```
Mogztter: I'm wondering if we could do it in the `registerExtension` ?

src/asciidoctor-extensions-api.js
```diff
@@ -9,6 +9,14 @@ var toBlock = function (block) {
   return block;
 };
 
+var registerExtension = function (registry, type, processor, name) {
+  if (typeof processor === 'function') {
```
Mogztter: `processor` can be undefined if the user only used a "name". In this context "name" must be a "function", so we could do:

```js
if (typeof processor === 'undefined') {
  processor = name;
  name = null;
}
```

Arguably this is not a better solution :neutral_face: 

Also should we throw an exception in this case:

```js
this.block('not like this...'); // It ain't gonna work :|
```

src/asciidoctor-extensions-api.js
mojavelinux: I thought about that, but it isn't always needed. The purpose of the first iteration of this function was to allow all the logic to be tested. I think we should make improvements to it in a subsequent issue.

src/asciidoctor-extensions-api.js
mojavelinux: True. But I don't think it's the responsibility of the JavaScript API to validate the arguments. That should happen in the Ruby method.

The purpose of the API method is to prepare the arguments for the Ruby method. The only thing we have to do is coerce the function to a Ruby block (since Opal doesn't do this for us). Otherwise, we're just passing through the arguments as they were given.

mojavelinux: I've noticed that the Ruby side doesn't throw an intelligible error in this case. It probably should.

mojavelinux: Hmm, actually, it does. It just seems we aren't hitting it. I'll investigate.

mojavelinux: I'll adjust it so that we hit the correct error when only a string argument is passed in.

mojavelinux: Aha! I see. The error on the Ruby side only happens if the value is a symbol.

mojavelinux: As it turns out, this is the correct logic. The problem is that the exception in core is obtuse. I will address that for 1.5.7.

```
this.block('not like this...');
```

This is a valid call. What it means is to look for a class with the name "not like this...". Obviously, it will fail to resolve a class. But right now, core sends back the obtuse message "wrong constant name" instead of "Could not resolve class for name". That's what I'll focus on fixing.

mojavelinux:
I've submitted a patch upstream to ensure the user gets a sensible error message. I also simplify the logic to resolve a class from a string for Ruby >= 2 (including Opal). See https://github.com/asciidoctor/asciidoctor/issues/2622

PR #455, opened 2018-03-21
fix stylesheet path for node

mojavelinux:


mojavelinux:
I think we may have to do this for all environments, right?

Mogztter:
> I think we may have to do this for all environments, right?

In a Node.js environment we know for sure where the stylesheet is because we are using `require`. In other environments we don't know for sure, so I guess `css/` is a good default ?

mojavelinux:
I'm not sure how it affects other environments, which is why I needed to check with you. For example, how does it affect nashhorn? Does it try to read the file from disk?

I'm happy with just fixing for node for now as that's like the case that is going to use it. We may want to embed the stylesheet for other environments in a later release.

mojavelinux:
I'll look into the CI error.

mojavelinux:
Updated!

npm/module/nashorn.js
```diff
@@ -12,6 +12,10 @@ const nashornCheckConvert = (result, testName) => {
     log.error(`${testName} failed, AsciiDoc source is not converted`);
     process.stdout.write(result);
   }
+  if (result.indexOf('Asciidoctor default stylesheet') === -1) {
```
Mogztter: :+1:

PR #458, opened 2018-03-22
Move documentation to docs

Mogztter:


mojavelinux:
This is correct for right now. Once we make this an Antora documentation component, the docs will need to be moved. But since we aren't sure yet how we want to structure the modules, I don't think we should do that yet.

PR #459, opened 2018-03-23
fix failure when retrieving extension groups when none have been defined

mojavelinux:


mojavelinux:
Use this.$groups() instead of this.groups since the accessor method performs logic.

Mogztter:
Thanks for the fix 😉

PR #461, opened 2018-03-23
resolves #460 map AbstractNode#getParent in the public API

### LINKED ISSUE #460: Map AbstractNode#getParent in public API
Map the AbstractNode#getParent method in the public API. It should map to the raw property `parent`.

mojavelinux:


Mogztter:
👍

PR #466, opened 2018-03-25
✨ Allow console.log in development

Mogztter:
**Usage**

```
NODE_ENV=development yarn build:quick
```

I don't know if it's a good idea to add this environment variable by default on `build:quick` ?

cc: @mojavelinux


mojavelinux:
> I don't know if it's a good idea to add this environment variable by default on build:quick ?

I wonder if it's better to either remove the call to lint or put lint after the tests (so it doesn't block running the tests).

Mogztter:
> I wonder if it's better to either remove the call to lint or put lint after the tests (so it doesn't block running the tests).

I think it's important to run the linter but we could rename `build:quick` to `build:dev` and add the environment variable `NODE_ENV=development`.
So it's clear that you are using a "development" task and thus allow to use `console.log`. Wdyt ?

mojavelinux:
I agree the linter is important and often bites me when I don't run it. I think the main problem is that it runs before the tests. If we just reverse it, then I think it would be fine. It could even eliminate the need for the exception.

Mogztter:
> If we just reverse the order, then I think it would be fine. 

Oh right I see... :+1: 

> It could even eliminate the need for the exception

I will let the exception for now, the code is simple and it could be useful for debugging

PR #469, opened 2018-03-26
fix Extensions#unregister

mojavelinux:
- use more portable array flattening logic
- add test when names are passed as array

### DIFF
--- a/spec/node/asciidoctor.spec.js
+++ b/spec/node/asciidoctor.spec.js
@@ -450,7 +450,7 @@ Content 2`;
             });
           });
         });
-        const groups = extensions.getGroups();
+        let groups = extensions.getGroups();
         expect(groups).toBeDefined();
         expect(Object.keys(groups).length).toBe(3);
         expect(Object.keys(groups)).toEqual(['test', 'foo', 'bar']);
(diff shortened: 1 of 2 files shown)


spec/node/asciidoctor.spec.js
```diff
@@ -468,6 +471,38 @@ Content 2`;
...
+          });
+        });
+        let groups = extensions.getGroups();
+        expect(groups).toBeDefined();
+        expect(Object.keys(groups).length).toBe(2);
+        expect(Object.keys(groups)).toEqual(['foo', 'bar']);
+        extensions.unregister(['foo', 'bar']);
+        groups = extensions.getGroups();
+        expect(groups).toBeDefined();
+        expect(Object.keys(groups).length).toBe(0);
```
Mogztter: Tests are failing, I didn't debug but `Object.keys(groups).length` equals 1 not 0: https://travis-ci.org/asciidoctor/asciidoctor.js/jobs/358264467

mojavelinux:
Something happened with the builds that they never got triggered after I fixed the problem. I just forced pushed again. Let's see if that works because everything is green locally.

PR #471, opened 2018-04-08
Setup Antora

Mogztter:
* [x] Configure the site header (see: https://gitlab.com/antora/docs.antora.org/tree/master/supplemental-ui)
* [x] Configure the build 
* [x] Publish to Netlify


docs/modules/spec/pages/include-support-matrix.adoc
```diff
@@ -68,15 +74,15 @@ include uri not readable: `file:///root/path/to/file.adoc`.
 |:x: +
 `SecurityError: Jail is not an absolute path: https://raw.com/master`
 |:x: +
-include file not found: `{local_pwd}/https:/raw.com/master/file.adoc`.
+include file not found: `\{local_pwd\}/https:/raw.com/master/file.adoc`.
```
mojavelinux: You only need to escape the leading curly brace.

antora/supplemental-ui/partials/header-content.hbs
```diff
@@ -0,0 +1,29 @@
...
+        <a href="{{or site.url (or siteRootUrl siteRootPath)}}">Docs</a>
+      </div>
+      <button class="navbar-burger" data-target="topbar-nav">
+        <span></span>
+        <span></span>
+        <span></span>
+      </button>
+    </div>
+    <div id="topbar-nav" class="navbar-menu">
+      <div class="navbar-end">
```
Mogztter: I think we should add a link to the JS API documentation ?

docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,7 @@
+= Asciidoctor.js Documentation
+:uri-opal: https://opalrb.com
+
+Asciidoctor.js brings AsciiDoc to the JavaScript world!
+
+This project uses {uri-opal}[Opal] to transpile https://asciidoctor.org[Asciidoctor], a modern implementation of AsciiDoc, from Ruby to JavaScript to produce _asciidoctor.js_.
```
Mogztter: I'm not sure about this introduction.
I don't think we should start off by mentioning Opal. I'm not saying that we should hide the fact that we are using Opal but we should probably focus on what Asciidoctor.js is and does and less on how Asciidoctor.js is built.

docs/modules/processor/examples/test.js
```diff
@@ -0,0 +1,53 @@
+const asciidoctor = require('asciidoctor.js')();
+
+// tag::load-file[]
+var doc = asciidoctor.loadFile('sample.adoc');
+// end::load-file[]
+
+// tag::document-title[]
+console.log(doc.getDocumentTitle()); // The Documentation Chronicles: Based on True Events
```
Mogztter: @mojavelinux It would be awesome if the expected result was generated dynamically. We could even create a custom include directive to read the expected value from a test case.

**code**
```js
describe('Extract information from a document', () => {
  it('should get the document title', () => {
    const doc = asciidoctor.loadFile('sample.adoc');
    // tag::document-title[]
    expect(doc.getDocumentTitle()).toBe('The Documentation Chronicles: Based on True Events');
    // end::document-title[]
  });
});
```

**documentation**
```
include-js-expect::{examplesdir}/spec.js[tags=document-title]
```

**result**
```js
console.log(doc.getDocumentTitle()); // The Documentation Chronicles: Based on True Events
```

* `doc.getDocumentTitle()` is extracted from `expect(${code})`
* `The Documentation Chronicles: Based on True Events` is extracted from `toBe(${expectedValue})`
* Then we apply a template `console.log(${code}); // ${expectedValue}`

Mogztter: It should be possible with an include processor but I don't want to handle partial include (see https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/master/lib/whitespace-include-processor.rb)

Mogztter: @mojavelinux Do you think it could be a good idea to introduce an include content processor triggered once the content has been read ?

```rb
Extensions.register do
  include_content_processor do
    process do |doc, reader, source, attributes|
      reader.push_include source.gsub("\t", '    '), target, target, 1, attributes
      reader
    end

    def handles? target
      target.start_with? 'code/'
    end
  end
end
```

PR #474, opened 2018-04-15
Add logger impl to stdlib

Mogztter:
Temporary fix until https://github.com/opal/opal/pull/1802 is merged

### DIFF
--- a/spec/node/asciidoctor.spec.js
+++ b/spec/node/asciidoctor.spec.js
@@ -29,7 +29,7 @@ function fileExists (path) {
   try {
     fs.statSync(path);
     return true;
-  } catch(err) {
+  } catch (err) {
     return !(err && err.code === 'ENOENT');
   }
 }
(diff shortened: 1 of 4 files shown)


spec/node/asciidoctor.spec.js
```diff
@@ -48,6 +48,34 @@ describe('Node.js', () => {
...
+`;
+        const defaultLogger = asciidoctor.LoggerManager.getLogger();
+        const memoryLogger = asciidoctor.MemoryLogger.$new();
+        try {
+          asciidoctor.LoggerManager.setLogger(memoryLogger);
+          asciidoctor.convert(input);
+          const errorMessage = memoryLogger.getMessages()[0];
+          expect(errorMessage.severity.toString()).toBe('ERROR');
+          expect(errorMessage.message['text']).toBe('invalid part, must have at least one section (e.g., chapter, appendix, etc.)');
+          expect(errorMessage.message['source_location'].lineno).toBe(8);
```
Mogztter: Since the `Cursor` object will be available in the context, we should probably add a wrapper API around this object:

```js
const cursor = Asciidoctor.Cursor.$new();
cursor.getDirectory();
cursor.getFile();
cursor.getPath();
cursor.getLineNumber();
```

mojavelinux:
This is not the first time the Cursor object has made its way into the public API. It's also the return value of the getSourceLocation method on a block. So I agree it should be API wrapped.

PR #477, opened 2018-04-22
Resolves #464, convert attributes passed to process method of an extension to JavaScript object

### LINKED ISSUE #464: Convert attributes passed to process method of an extension to JavaScript object
For extensions to which it applies, convert the attributes argument of the process method from an Opal hash to a JavaScript object.

Currently, we see code such as this:

```js
this.process((parent, reader, attrs) => {
    let role = Opal.hash_get(attrs, 'role')
    let id = Opal.hash_get(attrs, 'id')
    ...
})
```

We'd like to reduce that to:

```js
this.process((parent, reader, attrs) => {
    let role = attrs.role
    let id = attrs.id
    ...
})
```

That may mean we have to proxy the function in certain cases and coerce its attributes.

Mogztter:
Resolves #464

Thanks @oncletom for the suggestion :+1:

### DIFF
--- a/spec/node/asciidoctor.spec.js
+++ b/spec/node/asciidoctor.spec.js
@@ -355,7 +355,7 @@ Content 2`;
             this.named('cloak');
             this.onContexts('paragraph', 'literal');
             this.process((parent, reader, attrs) => {
-              return this.createBlock(parent, 'paragraph', 'cloaked: ' + Opal.hash_get(attrs, 'cloaked-context'));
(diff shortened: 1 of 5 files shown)


Mogztter:
I'm pretty happy with this one but it's a breaking chance.
Should we add the IE polyfill ?

Mogztter:
I don't want to include the polyfill, I will just add a note on the documentation.

PR #486, opened 2018-05-06
Upgrade to Opal 0.11.1.dev@0dccf06

Mogztter:
Resolves #484

PR #489, opened 2018-05-06
resolves #482 add missing continue to redo

### LINKED ISSUE #482: Escaping pipe does not seem to work
Having an example table with escaped pipe ...
```
|===
|`-a\|-b`|Options cannot be used together
|===
```
I get the following picture in the browser:

asciidoctor-firefox

With `asciidoctor-pdf` it works just fine:
[with-asciidoctor.pdf](https://github.com/asciidoctor/asciidoctor.js/files/1976491/with-asciidoctor.pdf)

Firefox version: 59.0.2 (64 bit) - for Ubuntu Canonical
Asciidoctor.js Live Preview version: 1.5.6.100

mojavelinux:


mojavelinux:
This patch is really just to show you what's missing from Opal. I don't see any reason why this can't be resolved in Opal itself. The redo logic is already there. What's missing is that the loop doesn't get advanced after setting the redo status. In other words, the call to `continue` is missing.

iliabylich:
> I don't see any reason why this can't be resolved in Opal itself

It can be, take a look at the https://github.com/opal/opal/issues/1816#issuecomment-386923238 please.

> The redo logic is already there

Only for blocks, for the `while` loop it's broken (or incomplete, doesn't matter)

> In other words, the call to continue is missing

That's not what `redo` does. `redo` explicitly re-runs current iteration, but `continue` (or `next` in Ruby) simply "exits" from the current iteration and **potentially** starts a new one. For example if the loop has a condition that is not valid anymore (i.e. reduces to `false`) `next` will have zero effect.

Of course, it's possible that for your case `next` is fine, but from the first glance it seems to be wrong.

mojavelinux:
> Of course, it's possible that for your case next is fine

True, this is unique to this case.

> but from the first glance it seems to be wrong.

Actually, it's right. But the reason is, the while condition has no logic. It's always true.

mojavelinux:
> That's not what redo does. redo explicitly re-runs current iteration

I was thinking specifically about this case. You are correct that if the while condition had actually logic, it could cause different behavior. This works in this case since redo and next would essentially have the same behavior.

iliabylich:
Then I guess it's perfectly fine. `next` does exactly what `redo` is supposed to do.

mojavelinux:
Actually, I still think the logic in Opal is already correct, it's just missing the continue. Here's why:

```js
$b = false;
while ($b || $truthy(true)) {
  $b = false;
  // some code
  $b = true; continue;
  // skipped code
}
```

Because of `$b` the while condition is going to be short-circuited. But what Opal is not doing is adding that `continue` keyword after the `$b = true;` assignment. If it did, that would complete the redo logic.

...perhaps there's a case I'm not considering, but this looks like everything is already in place minus the `continue` keyword.

mojavelinux:
I posted more details to the linked issue.

Mogztter:
This issue will be fixed upstream: https://github.com/opal/opal/pull/1820

mojavelinux:
:1st_place_medal:

Mogztter:
https://github.com/asciidoctor/asciidoctor.js/pull/486 will resolve this issue

PR #492, opened 2018-05-07
Resolves #481, add a prepend function to prepend document extension (default is append)

### LINKED ISSUE #481: Add position option to extension DSL
Each extension can specify a position option, which determines whether the extension is added to the beginning or end of the list for that extension type. The DSL should support this option.

Current:

```js
asciidoctor.Extensions.register(function () {
  this.includeProcessor(function () {
    this.$option('position', '>>')
    this.handles((target) => target.startsWith('https://'))
    this.process((doc, reader, target, attrs) => {
      ...
    })
  })
})
```

Proposed:

```js
asciidoctor.Extensions.register(function () {
  this.includeProcessor(function () {
    this.position('>>')
    this.handles((target) => target.startsWith('https://'))
    this.process((doc, reader, target, attrs) => {
      ...
    })
  })
})
```

The `>>` refers to unshift. I'm open to introducing an alternate keyword here.

Mogztter:


mojavelinux:
I like it. I think we should add this alias to core too. Underneath it will still set the same option value. It's just a porcelain method.

mojavelinux:
The only think we'll have to be careful about is not conflicting w/ the built-in prepend keyword. That doesn't affect the JavaScript API, but would be a problem for the Ruby API.

Mogztter:
Sorry for the noise, my IDE automatically reformats the code :disappointed: 
Shall we move forward ?

mojavelinux:
Yep!

PR #500, opened 2018-05-10
🍩 Map getter and setter methods on Section

Mogztter:
Especially useful to build a "Table Of Contents": https://github.com/asciidoctor/asciidoctor-browser-extension/issues/86

### DIFF
--- a/src/asciidoctor-core-api.js
+++ b/src/asciidoctor-core-api.js
@@ -325,6 +325,114 @@ AbstractBlock.$$proto.getLineNumber = function () {
 };
 
 /**
+ * Check whether this block has any child Section objects.
+ * Only applies to Document and Section instances.
+ * @memberof AbstractBlock
+ * @returns {boolean} - true if this block has child Section objects, otherwise false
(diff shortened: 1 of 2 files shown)


mojavelinux:
:+1:

src/asciidoctor-core-api.js
```diff
@@ -324,6 +324,114 @@ AbstractBlock.$$proto.getLineNumber = function () {
...
+  this.special = value;
+};
+
+/**
+ * Get the state of the numbered attribute at this section (need to preserve for creating TOC).
+ * @memberof Section
+ * @returns {boolean}
+ */
+Section.$$proto.isNumbered = function () {
+  return this.numbered;
```
Mogztter: @mojavelinux `numbered` is an attribute accessor but the documentation is just "Get the state...". So I didn't add a setter. Is it OK ?

mojavelinux: The API doc was out of date. It should be read/write. I just updated it.

PR #506, opened 2018-05-16
Replace Jasmine + Karma with Mocha + Chai + Puppeteer

### LINKED ISSUE #222: Karma & jasmine are very difficult to debug
Why don't we use nodejs tape as Testing framework instead of Karma/Jasmine. That should simplify how we debug the tests case as configuring  Karma/Jasmine is not so easy ?

Mogztter:
* Use a plain HTML page with <script> to test in the browser
* Remove Webpack and Browerify (integration tests)
* Replace Jasmine with Mocha + Chai

Resolves #222 

As part of this pull request I've removed Webpack and Browerify. I think we should provide integration tests to make sure that the Asciidoctor.js module is valid but we could do that in a following pull request.

:warning: Node.js 8+ is required
:zap: The build is two times faster
:tada: One nice thing is that you can open `spec/browser/index.html` file in a browser and :boom: 

boom

You will need to add `--allow-file-access-from-files` when launching Chrome otherwise some tests will fail. As an alternative you can serve this file using an HTTP server:
```
npm run server
```
Then open http://localhost:8080/spec/browser/index.html


mojavelinux:
Impressive work!

You might consider adding dirty-chai, which allows you to write terminal assertions as methods like this:

```
expect(someting).to.be.true()
```

That's jives more with standard.js.

PR #511, opened 2018-06-17
Upgrade to Opal 0.11.1.dev@3c8d93e

Mogztter:
:boom: I've removed _asciidoctor-docbook.js_, _asciidoctor-template.js_ and _asciidoctor-reveal.js_ dependencies to avoid circular dependencies issues.

Sometimes Opal introduces a breaking change and in this case we need to compile everything with the same version of Opal. More specifically _asciidoctor-docbook.js_ compiled with version X of Opal will not be compatible with _asciidoctor.js_ compiled with version Y of Opal. 
Since _asciidoctor.js_ has a dependency on _asciidoctor-docbook.js_ (and vice-versa) the build will fail on both projects. So I can't upgrade the version of _asciidoctor.js_ in _asciidoctor-docbook.js_ because the build is failing on _asciidoctor.js_ and I can't upgrade the version of _asciidoctor-docbook.js_ in _asciidoctor.js_ because the build is failing on _asciidoctor-docbook.js_.

I think that each projects should provide unit/integration tests with Asciidoctor.js (and not the other way around).
Having said that, those tests were very useful to make sure that the main projects in the ecosystem are working with the new version of Asciidoctor.js before doing a release.

:warning: Now Opal classes are `function` so if we want to know if a variable is a class or a lambda/function, we need to use `variable.$$is_class` (and not `typeof 'function'`)

:warning: I'm using a workaround for this issue https://github.com/opal/opal/issues/1846


iliabylich:
Weird, not sure if it's related to your code, looks like it's actually works fine:
``` ruby
module Asciidoctor
  module Extensions
    class IncludeProcessor
    end
  end
end


%x{
  var includeProcessor = Opal.const_get_qualified(
    Opal.const_get_qualified(
      Opal.const_get_relative(Opal, 'Asciidoctor'),
      'Extensions'
    ),
    'IncludeProcessor'
  );

  var scope = Opal.klass(null, includeProcessor, 'LoremIncludeProcessor', function () {});

  Opal.def(scope, '$process', function (doc, reader, target, attrs) {
    console.log(this, doc, reader, target, attrs);
    // return reader.$push_include('Lorem ipsum', target, target, 1, attrs);
  });

}

p LoremIncludeProcessor
p LoremIncludeProcessor.new.process(1, 2, 3, 4)
```

it prints

```
LoremIncludeProcessor
$IncludeProcessor { '$$id': 62 } 1 2 3 4
nil
```

mojavelinux:
> I think that each projects should provide unit/integration tests with Asciidoctor.js (and not the other way around).

:+1:

PR #513, opened 2018-07-08
Getting started docs

Mogztter:


Mogztter:
@mojavelinux I think we should update the README to add a link on the Antora documentation website. What do you think ?

PR #514, opened 2018-07-08
Trigger a deployment on docs.asciidoctor.org

Mogztter:
`NETLIFY_WEBHOOK_URL` variable is defined on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,4 +60,8 @@ deploy:
 
 # Trigger a deploy on asciidoctor/docs.asciidoctor.org
 after_success:
-- curl -X POST -d '' "$NETLIFY_WEBHOOK_URL"
+- >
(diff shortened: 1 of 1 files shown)


Mogztter:
The webhook was triggered 4 times (once for each build in the matrix).

trigger

We should probably use an environment variable to trigger the webhook only once:

```yml
after_success:
->
  if ! [ "$AFTER_SUCCESS_RUN" ]; then
    export AFTER_SUCCESS_RUN=1;
    # ...
```

mojavelinux:
I think you can use build stages to accomplish a single after hook for all builds in the matrix.

https://docs.travis-ci.com/user/build-stages

Mogztter:
Superseded by https://github.com/asciidoctor/asciidoctor.js/pull/516

Thanks @mojavelinux and @oncletom for your feedbacks.

PR #515, opened 2018-07-09
Another way to trigger Netlify Webhook

oncletom:
refs #514

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,11 +57,10 @@ deploy:
     repo: asciidoctor/asciidoctor.js
     tags: true
     node: node
-
-# Trigger a deploy on asciidoctor/docs.asciidoctor.org
-after_success:
-- >
-  if ! [ "$AFTER_SUCCESS_RUN" ]; then
-    export AFTER_SUCCESS_RUN=1;
-    curl -X POST -d '' "$NETLIFY_WEBHOOK_URL"
-  fi
+- provider: script
+  script: curl -X POST -d '' "$NETLIFY_WEBHOOK_URL"
+  on:
+    condition: '$ASCIIDOCTOR_CORE_VERSION = $ASCIIDOCTOR_CORE_VERSION_PUBLISH'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -57,11 +57,10 @@ deploy:
...
-  if ! [ "$AFTER_SUCCESS_RUN" ]; then
-    export AFTER_SUCCESS_RUN=1;
-    curl -X POST -d '' "$NETLIFY_WEBHOOK_URL"
-  fi
+- provider: script
+  script: curl -X POST -d '' "$NETLIFY_WEBHOOK_URL"
+  on:
+    condition: '$ASCIIDOCTOR_CORE_VERSION = $ASCIIDOCTOR_CORE_VERSION_PUBLISH'
+    repo: asciidoctor/asciidoctor.js
+    tags: true
```
Mogztter: In this case we will only send a notification when we perform a release. I think the condition should be "send a notification when a successful build occurred on the master branch".

oncletom: @Mogztter so to clarify, you want to update the doc for each push on `master` although you release the npm package on a new tag, correct?

Mogztter: Yes exactly!
For now the documentation is only on the master branch but in the future we will also need to update the documentation when we push some updates on a branch. For instance if I edit the documentation on the `v2.0.x` branch, I want to notify https://asciidoctorjs-docs.netlify.com

I'm working on a [build stages](https://docs.travis-ci.com/user/build-stages/) solution, see #516

oncletom: Cool :-) Closing this PR then; I can see build stages to be more flexible and visually representative of what's happening as a sequence of events.

Mogztter: Indeed!
I really like the fact that if a stage does not validate all conditions then the stage is not display on the build page: https://travis-ci.org/asciidoctor/asciidoctor.js/builds/401705215

Anyway thanks for your suggestions :+1:

PR #516, opened 2018-07-09
Use Travis build stages

Mogztter:


Mogztter:
It seems to be working well!
Here's a build on the master branch: https://travis-ci.org/asciidoctor/asciidoctor.js/builds/401757561

As you can see the "Notify" stage is triggered.

Mogztter:
Yep, it's working great. The only downside I found is that (as far as I know) you can't configure an optional step (ie. a step that can fail) : https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail

PR #527, opened 2018-08-05
The "user manual" is now published on Netlify

Mogztter:
The documentation available on Netlify is now more complete than the file `docs/manual.adoc`


mojavelinux:
I'm excited about this, but I'm also very eager to get them promoted to docs.asciidoctor.org once that site is available. I'm hopeful work is going to resume on that effort next week.

PR #533, opened 2018-08-27
resolves #530 map AbstractBlock#append method in API

### LINKED ISSUE #530: Add AbstractBlock#append method to API
Map AbstractBlock#append to the underlying Ruby method to allow a block to be appended to a parent in the AST. See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/abstract_block.rb#L94-L110.

mojavelinux:


Mogztter:
And merged, thanks again!

PR #536, opened 2018-08-28
resolves #532 map ListItem#setText method in API

### LINKED ISSUE #532: Add ListItem#setText to API
Map the ListItem#setText method to the underlying text= method. This method allows a program to assign next text to a list item.

mojavelinux:


Mogztter:
Tests are failing because I'm using an external service (namely `raw.githubusercontent.com`).
I should probably use a local HTTP server to test the include directive with a remote URI :thinking:

mojavelinux:
Not sure. I don't have experience with that.

You could just increase the timeout.

Mogztter:
> Not sure. I don't have experience with that.

Ok.

> You could just increase the timeout.

:+1:

Mogztter:
Thanks @mojavelinux :+1:

PR #537, opened 2018-08-28
use simpler mechanism to convert from Opal hash to JavaScript object

mojavelinux:


mojavelinux:
Hmm, I just realized the IE 11 doesn't support Object.assign. Really? I suppose we could check ('assign' in Object) and then use the old school way as a fallback.

mojavelinux:
It turns out Object.assign is slower. But what would be faster is `for..in`.

mojavelinux:
Updated. There used to be a risk of doing `for..in`, but I think that's been resolved now, right?

(this would be the perfect use case for `for..of`, but IE 11 doesn't support that either.

mojavelinux:
This build seems back to normal.

mojavelinux:
I noticed you're already using Object.assign in this file. If you'd rather switch back to Object.assign, let me know.

Mogztter:
> There used to be a risk of doing for..in

I'm just aware that `for .. in` is slower than a basic array loop (because it does more). What do you mean by "risk" ?

> I noticed you're already using Object.assign in this file. If you'd rather switch back to Object.assign, let me know.

Yes you can use `Object.assign`.
I need to document how to integrate Asciidoctor.js with old/deprecated browsers:

* https://polyfill.io/ 
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill

mojavelinux:
Okay, let's go with Object.assign. I'll update.

mojavelinux:
I think this is ready to go now.

Mogztter:
The code looks nice 👍

PR #538, opened 2018-08-29
resolves #529 map AbstractBlock#applySubstitutions in API

### LINKED ISSUE #529: Add AbstractNode#applySubstitutions to API
Map the AbstractNode#applySubstitutions method to the underlying apply_subs method that's added to nodes by the Substitutors module. The apply_subs method provides access to performing the various substitutions. We could also consider mapping substituteAttributes since this is a very common substitution to perform from the API.

See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/substitutors.rb#L87-L93

Currently this is required to substitute attribute references in the target of a block macro since this capability is not currently available via the DSL. See the open issue: https://github.com/asciidoctor/asciidoctor/issues/2839

mojavelinux:


Mogztter:
Merged! :smile:

PR #542, opened 2018-09-05
Resolves #535, upgrade to Opal 0.11.99.dev@296930e and add a test

### LINKED ISSUE #535: Processor hangs when parsing overlapping passthrough in monospaced text
The processor in 1.5.7-rc.1 hangs processing the following document:

```
Link the system library `+libconfig++.so.9+` located at `+/usr/lib64/libconfig++.so.9+`.
```

Technically, that passthrough syntax doesn't produce the desired result. But it shouldn't cause Asciidoctor.js to hang.

Note that Asciidoctor Ruby 1.5.7 processes the same syntax without hanging. Somewhere, the cursor is not advancing in Asciidoctor.js.

Mogztter:
Resolves #535

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5134,18 +5134,18 @@
       }
     },
     "opal-compiler": {
-      "version": "1.0.7",
-      "resolved": "https://registry.npmjs.org/opal-compiler/-/opal-compiler-1.0.7.tgz",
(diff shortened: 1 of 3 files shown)


Mogztter:
It also partially resolves https://github.com/opal/opal/issues/1846 since I save and restore `Function` methods.

PR #543, opened 2018-09-08
when patching regexp, don't replace . that's preceded by a backslash

mojavelinux:


mojavelinux:
This should resolve our test failure.

PR #553, opened 2018-09-30
Demonstrate how to use a JSON formatter

Mogztter:
Once approved, we should add an example to the documentation.

### DIFF
--- a/src/asciidoctor-core-api.js
+++ b/src/asciidoctor-core-api.js
@@ -1465,7 +1465,61 @@ Cursor.prototype.getLineNumber = function () {
 };
 
 // Logger API (available in Asciidoctor 1.5.7+)
-// REMIND: we are using "skip_missing" because this API is only available starting with Asciidoctor 1.5.7
+
+function initializeLoggerFormatterClass (className, functions) {
+  var superclass = Opal.const_get_qualified(Opal.Logger, 'Formatter');
+  var scope = Opal.klass(Opal.Object, superclass, className, function () {});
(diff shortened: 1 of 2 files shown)


src/asciidoctor-core-api.js
```diff
@@ -1465,7 +1465,61 @@ Cursor.prototype.getLineNumber = function () {
 };
 
 // Logger API (available in Asciidoctor 1.5.7+)
-// REMIND: we are using "skip_missing" because this API is only available starting with Asciidoctor 1.5.7
+
+function initializeLoggerFormatterClass (className, functions) {
```
Mogztter: This code is mostly copied-pasted from `initializeProcessorClass`.
I think we should create a base function.

src/asciidoctor-core-api.js
```diff
@@ -1465,7 +1465,61 @@ Cursor.prototype.getLineNumber = function () {
...
+    if (functions.hasOwnProperty(key)) {
+      (function (key) {
+        var userFunction = functions[key];
+        if (key === 'postConstruct') {
+          postConstructFunction = userFunction;
+        } else if (key === 'initialize') {
+          initializeFunction = userFunction;
+        } else {
+          Opal.def(scope, '$' + key, function () {
+            for (var i = 0; i < arguments.length; i++) {
```
Mogztter: Here we tranform Opal.Hash to JSON for usability.
Though it's actually only really useful for the `message` argument in the `call` function.

PR #555, opened 2018-10-01
Write the Fedora package extension with the extension API

Mogztter:
@mojavelinux It was useful as I fixed two issues in the process :+1: 

The `$new` function fells a bit weird but I'm OK with it :wink: 

Reference: https://pagure.io/fedora-docs/docs-fp-o/blob/master/f/lib/extensions/package-inline-macro.js

### DIFF
--- a/spec/node/asciidoctor.spec.js
+++ b/spec/node/asciidoctor.spec.js
@@ -772,6 +772,29 @@ intro
           const result = asciidoctor.convert('include::whatever.adoc[]', opts);
           expect(result).to.contain('included content');
         });
+
+        it('should be able to create the Fedora package inline macro', () => {
+          const PackageInlineMacro = asciidoctor.Extensions.createInlineMacroProcessor('PackageInlineMacro', {
+            initialize: function (name, config) {
+              this.DEFAULT_PACKAGE_URL_FORMAT = 'https://apps.fedoraproject.org/packages/%s';
+              this.super(name, config);
+            },
+            process: function (parent, target) {
+              const format = parent.getDocument().getAttribute('url-package-url-format', this.DEFAULT_PACKAGE_URL_FORMAT);
+              const url = format.replace('%s', target);
+              const content = target;
+              const attributes = { window: '_blank' };
+              return this.createInline(parent, 'anchor', content, { type: 'link', target: url, attributes });
+            }
+          });
+          const registry = asciidoctor.Extensions.create();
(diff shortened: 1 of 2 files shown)


mojavelinux:
Oh, this is awesome!

I realized that it isn't necessary to create the processor instance when registering the extension. That's only required when the Antora context needs to be passed*. We'll have to think about how to handle that case, but since it isn't needed here, you can just pass the class itself.

\* In other words, we need some way to pass the Antora context to the processor when the extension is registered. One way is to create an instance of the processor when registering it.

Mogztter:
> Oh, this is awesome!

:smile: 

> I realized that it isn't necessary to create the processor instance when registering the extension. 

That's true, you can pass a class.

```js
const registry = asciidoctor.Extensions.create();
registry.inlineMacro('package', asciidoctor.Extensions.createInlineMacroProcessor(/* ... */));
```

> That's only required when the Antora context needs to be passed*. We'll have to think about how to handle that case, but since it isn't needed here, you can just pass the class itself.

Indeed, in this case, the name of the inline macro is defined when calling the `inlineMacro` function.

> In other words, we need some way to pass the Antora context to the processor when the extension is registered. One way is to create an instance of the processor when registering it.

Yes you can "override" the initialize function to pass the Antora context:

```js
const PackageInlineMacro = asciidoctor.Extensions.createInlineMacroProcessor('PackageInlineMacro', {
  initialize: function (name, config, antoraContext) {
    this.antoraContext = antoraContext;
    this.super(name, config);
  },
  process: function (parent, target) {
    /* ... */
  }
});

PackageInlineMacro.$new('package', {}, Antora.getContext())
```

mojavelinux:
> you can "override" the initialize function to pass the Antora context:

Indeed, that's what I've been doing. What I meant was some way to avoid the call to `$new`. Perhaps a createInstance method on the class that automatically passes the name and config, but accepts additional arguments.

Mogztter:
> Indeed, that's what I've been doing. What I meant was some way to avoid the call to $new. Perhaps a createInstance (or create) method on the class that automatically passes the name and config, but accepts additional arguments.

Why not, I think we should track this improvement in a new issue right ?

mojavelinux:
Indeed. Do you want me to file it? Or would you prefer to propose it?

Mogztter:
Thanks!
@mojavelinux Otherwise does this pull request look good to you ? Can I merge as is ?

mojavelinux:
Oh, absolutely. This looks extremely enticing!

PR #559, opened 2018-10-08
(wip) Test against GraalVM

Mogztter:


lib/asciidoctor/js/opal_ext/graalvm/file.rb
```diff
@@ -0,0 +1,7 @@
+class File
+
+  def self.read(path)
+    `IncludeResolver.read(path)`
```
Mogztter: Maybe I'm doing something wrong but the following code throws an exception:

```js
var Paths = Java.type('java.nio.file.Paths');
```

```
Exception in thread "main" TypeError: Java Interop is not available
```
Reference: https://github.com/graalvm/graaljs/blob/master/docs/user/JavaInterop.md#java-interoperability-features

To workaround this issue, I'm using a technique described in the GraalVM documentation: https://www.graalvm.org/docs/graalvm-as-a-platform/embed/#access-java-from-graal-languages

In my Java application I declare a class.
This class can be passed to the JavaScript context using `Polyglot.import` (see `spec/graalvm/AsciidoctorConvertWithGraalVM.java`)

npm/module/graalvm.js
```diff
@@ -0,0 +1,107 @@
...
+
+  // Run
+  log.debug(`running ${className}`);
+  const start = process.hrtime();
+  const result = child_process.execSync(`${javaBin} -classpath ./build ${className}`).toString('utf8');
+  log.debug(`running ${className} in ${process.hrtime(start)[0] }s`);
+  return result;
+};
+
+const graalvmRun = (name, jdkInstallDir) => {
```
Mogztter: Code should be shared with Nashorn.

spec/graalvm/asciidoctor-convert.js
```diff
@@ -0,0 +1,17 @@
...
+var data = '= asciidoctor.js, AsciiDoc in JavaScript\n' +
+'Doc Writer <docwriter@example.com>\n\n' +
+'Asciidoctor and Opal come together to bring\n' +
+'http://asciidoc.org[AsciiDoc] to the browser!.\n\n' +
+'== Technologies\n\n' +
+'* AsciiDoc\n' +
+'* Asciidoctor\n' +
+'* Opal\n\n' +
+'NOTE: That\'s all she wrote!!!\n\n' +
+'include::spec/fixtures/include.adoc[]';
```
Mogztter: include will be resolved

spec/graalvm/asciidoctor-convert.js
```diff
@@ -0,0 +1,17 @@
...
+
+var options = {safe: 'server', header_footer: true, attributes: {showtitle: true}};
```
Mogztter: The default stylesheet will be embed in the HTML5 page.

spec/graalvm/AsciidoctorConvertWithGraalVM.java
```diff
@@ -0,0 +1,37 @@
...
+    Context context = Context.newBuilder("js").allowIO(true).build();
+    context.getPolyglotBindings().putMember("IncludeResolver", new IncludeResolver());
+    context.eval("js", "var IncludeResolver = Polyglot.import('IncludeResolver');");
+    context.eval("js", "load('./build/asciidoctor-graalvm.js')");
+    context.eval("js", "load('./spec/graalvm/asciidoctor-convert.js')");
+  }
+
+  public static class IncludeResolver {
+    public String read(String path) throws IOException, URISyntaxException {
+      URL url = this.getClass().getClassLoader().getResource(path);
```
Mogztter: It's not so bad because the user can decide if the file should be resolved from the classpath

spec/graalvm/run.js
```diff
@@ -0,0 +1,9 @@
+const Mocha = require('mocha');
```
Mogztter: This file will be run using `build/graalvm/bin/node spec/graalvm/run.js`

Mogztter:
Green bar :tada:

mojavelinux:
Looks good to me.

I do anticipate being able to make the IncludeResolver pluggable in Asciidoctor core in the future, so just something to keep in mind. But at that time, we can just rewire the code for GraalVM to use pass it in, if we decide that works.

Mogztter:
> I do anticipate being able to make the IncludeResolver pluggable in Asciidoctor core in the future, so just something to keep in mind.

Yes that would be great :+1:

PR #566, opened 2018-10-21
Resolves #561, Document how to use Asciidoctor.js with GraalVM

### LINKED ISSUE #561: Document how to use GraalVM to run Asciidoctor.js
- Using the Graal Polyglot API to run Asciidoctor.js in a JVM-based application: https://www.graalvm.org/docs/graalvm-as-a-platform/embed/
- Using GraalVM `npm` and `node` script to run Asciidoctor.js: https://www.graalvm.org/docs/getting-started/#running-javascript

Mogztter:
Resolves #561


Mogztter:
I might create a sample repository to show how to embed Asciidoctor.js in a Java application using GraalVM: https://github.com/Mogztter/asciidoctor.js-integration-graalvm

I think it will be easier to understand because you can experiment with a working project based on Maven. What do you think ?

Mogztter:
I agree the documentation is too complicated but at the same time it's good to extract some informations from the tests suite (it's better than nothing).

If we take that approach one further we could even create a tiny Java library to wrap Asciidoctor.js.

And since it's technically possible to return JavaScript functions to Java values we could even create a Java API on top of Asciidoctor.js... AsciidoctorG (for GraalVM) is coming :wink:

mojavelinux:
I didn't mean to suggest that the documentation was too complicated, there's just perhaps an opportunity to show rather than tell in some cases. But experience will teach us where that line is.

> AsciidoctorG

I'm not sure whether to be excited or have a heart attack. :rofl: 

(Though I can say I've long been interested in an AsciidoctorJ impl based on Asciidoctor.js instead of Asciidoctor. And AsciidoctorJ is definitely moving in that direction.)

Mogztter:
> I didn't mean to suggest that the documentation was too complicated, 

Let's say that there's a lot to tell 😄 

> there's just perhaps an opportunity to show rather than tell in some cases. 

💯 agree

> I'm not sure whether to be excited or have a heart attack.

Come on, that's the future ! 🤣 

> (Though I can say I've long been interested in an AsciidoctorJ impl based on Asciidoctor.js instead of Asciidoctor. And AsciidoctorJ is definitely moving in that direction.)

Indeed but there's a big requirement, if the implementation is based on the Polyglot API, your Java application must run on GraalVM.
But I'm pleasantly surprised by how well Asciidoctor.js is running on GraalVM with little to no effort and the performance are wayyyyy better than Nashorn! (and it's not even a 1.0.0) 🚀

mojavelinux:
> I'm pleasantly surprised by how well Asciidoctor.js is running on GraalVM with little to no effort and the performance are wayyyyy better than Nashorn! 

I'm really excited to hear that. We really need JavaScript on the JVM to be going somewhere.

PR #587, opened 2018-11-03
 Resolves #570 introduce DSL method to mark document processor as preferred

### LINKED ISSUE #570: Map Registry#prefer in the API
Source code: https://github.com/asciidoctor/asciidoctor/commit/07bd1ea36aa2b52b8230237b4fcc2b099c1cdf15

Mogztter:


Mogztter:
> The prefer method was meant to replace prepend. Therefore, prepend should now be deprecated. It should also delegate to prefer.

Ok

> Are the new checks for arguments on the processor creation methods necessary to support the prefer chaining style? Or is that something different?

No it's actually a bug discovered when I was writing the test case for `prefer`.
The following code was not working:

```js
 this.prefer('tree_processor', asciidoctor.Extensions.newTreeProcessor({
  process: function (doc) {
    doc.append(this.createBlock(doc, 'paragraph', 'a', {}));
  }
}));
```

It was mandatory to pass a name as the first argument otherwise an exception was thrown at runtime.
The fix is on a separated commit and I should probably move it to another pull request.

mojavelinux:
:+1:

Mogztter:
@mojavelinux the `option` method (used by `prepend`) was defined on the `Processor` class but now the `prefer` is only defined on the `Registry` and on the `DocumentProcessorDsl`. As a consequence we cannot use `prefer` on a Block or Macro processor. I'm sure it's intended because I don't think it makes sense to use `prefer` on a block or macro processor right ?

mojavelinux:
Yes, that was indented. The option method has too broad of scope, which is another good reason why this method needed to be added. Now it has the right scope.

PR #588, opened 2018-11-03
Resolves #569, Map Processor#createListItem

### LINKED ISSUE #569: Map Processor#createListItem in the API
Source code: https://github.com/asciidoctor/asciidoctor/commit/9c1e394013f27c4cde15281ab093a291c44dbc38

Mogztter:


spec/node/asciidoctor.spec.js
```diff
@@ -814,6 +814,42 @@ intro
...
+        it('should be able to create a list item linked to a list', () => {
+          const extensions = asciidoctor.Extensions;
+          try {
+            extensions.register('test', function () {
+              this.block(function () {
+                this.named('test');
+                this.onContext('paragraph');
+                this.process((parent) => {
+                  const list = this.createList(parent, 'ulist');
+                  list.append(this.createListItem(list, 'foo'));
```
Mogztter: @mojavelinux It can be a little confusing because at first I did the following:

```js
const list = this.createList(parent, 'ulist')
this.createListItem(list, 'foo') // we create foo in "list"
```

But as far as I understand I also need to `append` the child ?
I was expecting that to be the default behavior because the list (parent) is known by the list item (child) at creation time.

Maybe we should automatically append the child or remove the parent from the API:


```js
const list = this.createList('ulist')
list.append(this.createListItem('foo'))
parent.append(list)
```

```js
const list = this.createList(parent, 'ulist')
this.createListItem(list, 'foo')
```

mojavelinux: > But as far as I understand I also need to append the child ?

The parent argument can be nil when creating the node. If you don't set it, it will be assigned when you append.

All the node creation methods take parent as the first argument, so it wouldn't make sense to just change this one. It's really there for consistency (and also reserved for future use).

You don't always want the node to be appended right away. It highly depends on what you're doing. Typically, the code needs to do follow-up work to prepare the node before attaching it to the tree.

Issue #347, opened 2017-06-23
How to detect external links in Asciidoc


salfuve:
Hi everyone, is it possible to get all the links contained inside of an asciidoc?I don't want to converse an asciidoc into HTML file, i want to iterate above the AST and get all the external links.¿Is this possible?
Thank you all in advance :)

mojavelinux:
If you enable the `catalog_assets` option, then convert the document, the collection of links will be available via the `:links` key of the catalog.

Here's how it's done in Ruby (as of Asciidoctor 1.5.6).

```ruby
doc = Asciidoctor.load_file 'sample.adoc', safe: :safe, catalog_assets: true
doc.convert
puts doc.catalog[:links]
```

This is not a particularly sophisticated mechanism. But that's where we are today.

Mogztter:
In JavaScript (with this pull request #349)

```js
var doc = asciidoctor.load('link:index.html[Docs]', {'safe': 'safe', 'catalog_assets': true});
doc.convert();
var links = doc.getCatalog().links; // [ 'index.html' ]
```

gaurav-nelson:
On a similar note, how to extract all internal references ? both xrefs to same document and xrefs to other documents in directory?

mojavelinux:
References to other documents are not currently cataloged.

gaurav-nelson:
Thanks for the clarification @mojavelinux

gaurav-nelson:
Thanks @mojavelinux What about xref's to the same document?

mojavelinux:
Those are available via document.catalog.xrefs.

mojavelinux:
Correction. document.catalog.refs.

We're also proposing to expose this as an API method like document.getRefs() or something equivalent.

Mogztter:
@gaurav-nelson @salfuve Just so you know a new API is coming in Asciidoctor.js 1.5.8:

It should look like:

```js
const doc = asciidoctor.load(input, {'safe': 'safe', 'catalog_assets': true})
doc.convert()

const links = doc.getLinks()
const ids = doc.getIds()
const refs = doc.getRefs()
const images = doc.getImages()
const footnotes = doc.getFootnotes()
const indexTerms = doc.getIndexTerms()
````

gaurav-nelson:
Thanks @Mogztter This will surely make link extraction easy.

Mogztter:
[...]

PR #603, opened 2018-11-19
Resolves #540, Map getIndexTerms, getRefs, getIds and getImages

### LINKED ISSUE #540: How to extract all anchors ?
Bu running:
```javascript
var stuff = asciidoctor.loadFile("test.adoc",{'safe': 'safe', 'catalog_assets': true});
stuff.convert();
console.log(util.inspect(stuff.getReferences()['ids']))
```
I am getting the following output:
```bash
$Hash {
  '$$smap':
   { 'devel-intro-about': 'About This Guide',
     'devel-intro-design': 'Design Overview',
     'devel-intro-workflow': 'Workflow',
     'devel-intro-prep': 'Preparation',
     'devel-intro-init': 'Initialization',
     'directory-structure': 'Directory Structure',
     'devel-intro-spec-file': 'APB Spec File',
     'devel-intro-actions': 'Actions',
     'devel-intro-build': 'Build',
     'devel-intro-deploy': 'Deploy' },
  '$$map': {},
  '$$keys':
   [ 'devel-intro-about',
     'devel-intro-design',
     'devel-intro-workflow',
     'devel-intro-prep',
     'devel-intro-init',
     'directory-structure',
     'devel-intro-spec-file',
     'devel-intro-actions',
     'devel-intro-build',
     'devel-intro-deploy' ] }
``` 

How can I extract just these keys? Or if there is a direct API call to get all anchors from an asciidoc file?

Mogztter:
@mojavelinux I believe that `links` is always empty right ? The key is defined but never used:

https://github.com/asciidoctor/asciidoctor/blob/76504b08d73bf6aaf35435e9345028ebc6fc89aa/lib/asciidoctor/document.rb#L283

`getFootnotes` was already mapped.

- [x] Map `getLinks` https://github.com/asciidoctor/asciidoctor.js/issues/347#issuecomment-310815041

### DIFF
--- a/spec/node/asciidoctor.spec.js
+++ b/spec/node/asciidoctor.spec.js
@@ -399,6 +399,127 @@ intro
       expect(blocks[3].getTitle()).to.equal('Got <span class="icon">[file pdf o]</span>?');
     });
 
+    it('should get links catalog', () => {
+      const input = `https://asciidoctor.org[Asciidoctor]
+
+link:index.html[Docs]
+
+devel@discuss.arquillian.org
+
+mailto:hello@opendevise.com[OpenDevise]
+
+irc://irc.freenode.org/#fedora
+
+http://discuss.asciidoctor.org[Discuss Asciidoctor^]`;
+      const doc = asciidoctor.load(input, {'catalog_assets': true});
+      doc.convert();  // available only once the document has been converted
+      const linksCatalog = doc.getLinks();
+      expect(linksCatalog).to.have.members([
+        'https://asciidoctor.org',
+        'index.html',
+        'mailto:devel@discuss.arquillian.org',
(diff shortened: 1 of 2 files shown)


Mogztter:
> @mojavelinux I believe that links is always empty right ? The key is defined but never used:

Actually it should work: https://github.com/asciidoctor/asciidoctor.js/issues/347#issuecomment-310815041.
I need to investigate.

Mogztter:
> Actually it should work: #347 (comment).
I need to investigate.

Actually some objects in the catalog are only captured or available if the document has been converted and/or if the option `catalog_assets` was defined on the processor.
I've added the `getLinks` function and added JS documentation.

mojavelinux:
Correct.

PR #613, opened 2018-12-03
Fix typos in code contribution document

LucianBuzzo:


CONTRIBUTING-CODE.adoc
```diff
@@ -53,7 +53,7 @@ To build Asciidoctor.js, simply run the npm `build` script from the root of the
  $ npm run build
 
 NOTE: The build task will make some minor code changes on Asciidoctor Ruby code.
-As you may know String are immutable in JavaScript, so we need to replace `gsub!` and `sub!` methods.
+As you may know, Strings are immutable in JavaScript, so we need to replace the `gsub!` and `sub!` methods.
```
Mogztter: I don't know about this one. 
I think it's String as in the String keyword: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String

Maybe we should use a monospaced word:

"As you may know, `String` objects are immutable..."

What do you think ?

Mogztter: You're right... 

> The MDN docs for javascript data structures has this to say

They are actually using both "strings" and "Strings": "For example and unlike to C, Strings are immutable."

I think it's safe to use "strings".

Mogztter:
Thanks again Lucian :smile:

PR #617, opened 2018-12-03
Resolves #614, Add a missing requirement

### LINKED ISSUE #614: Update the contributing code guide to add JDK8 as a requirement
Since the tests suite is running against Nashorn, we need to add JDK8 as a requirement.
As mentioned in #597 Nashorn is not actively maintained so we could also decide to run the tests if `java` and `jjs` are available in the `PATH` (and otherwise prints a warning/info message in the console).

Nashorn tests are also pretty slow so we might as well just remove them ?

@mojavelinux Thoughts ?

Mogztter:
Also describe how to build against a release and recommend to use `build:quick` to quickly iterate.


Mogztter:
Merging as an interim step, as mentioned in #614 we should deprecate Nashorn.

PR #620, opened 2018-12-04
Resolves #85, Explain how to create and register a custom converter

### LINKED ISSUE #85: How can i integrate backends or How can i create a new backend for Asciidoctorjs
How can i integrate backends or How can i create a new backend for Asciidoctorjs?

Mogztter:


Mogztter:
Thanks Olivier for your review. 
Yes this API is stable and shouldn't change between 1.5.9 and 2.0. 
After 2.0, if we decide to introduce a breaking change then we will do a "major revision bump" (3.0).

PR #623, opened 2018-12-06
Improve toHash function

Mogztter:


src/asciidoctor-core-api.js
```diff
@@ -3,7 +3,7 @@
  * @private
  */
 var toHash = function (object) {
-  return object && !('$$smap' in object) ? Opal.hash(object) : object;
+  return !object || (object && object.$$is_hash) ? object : Opal.hash2(Object.keys(object), object);
```
mojavelinux: I don't think the first negative check is needed. I'd simply write it as:

```js
object && !object.$$is_hash ? Opal.hash2(Object.keys(object), object) : object;
```

Mogztter: Yes you're right, I don't know why I wrote this complicated test... :thinking:

mojavelinux: :tada:

Mogztter:
@mojavelinux Can we merge ?

mojavelinux:
Yep!

PR #627, opened 2018-12-11
Run tests against Chrome and Firefox

Mogztter:
Currently Puppeteer for Firefox does not fire events when using the `file` protocol that's why I'm using an embedded HTTP server.

But now I'm wondering if we should run the tests suite against both the `http` and `file` protocol ?


Mogztter:
Damn it's not working on Travis... I need to investigate 🕵️

Mogztter:
Build is green but `oraclejdk8` is ignored by Travis on Xenial Ubuntu.
As a result the following warning is printed:

> Closure Compiler requires Java7 or higher
> Requirements are not satisfied (see previous errors), skipping minify task

By the default the version is Java 11.0.1. I need to check if Google Closure Compiler is working with Java 11 and update https://github.com/bestikk/bestikk-uglify. Or we could also use another solution to minify `asciidoctor.js`.

Mogztter:
> Build is green but oraclejdk8 is ignored by Travis on Xenial Ubuntu.

Fixed.
I've also updated Google Closure Compiler to use the latest version 😄 

I think we are good to go!

I've shamelessly "stolen" @oncletom work on Asciidoctor.js PDF to create an embedded HTTP server 😜

Mogztter:
We can now run the benchmark against Firefox but the include directive in the user manual does not work:

```
asciidoctor: ERROR: <stdin>: line 3203: include file not readable: file:///home/guillaume/workspace/opensource/asciidoctor/asciidoctor.js/build/benchmark/customers.csv
```

With Chrome I can allow `file` access with the `--allow-file-access-from-files` option but as far as I know it's not possible with Firefox. I think we should run the benchmark against the HTTP protocol until Firefox provides an option to allow file access.

PR #629, opened 2018-12-13
Resolves #628, Add a governance page

### LINKED ISSUE #628: Add a page on the project governance

Mogztter:


docs/modules/project/pages/governance.adoc
```diff
@@ -0,0 +1,38 @@
...
+* Anthonny Quérouil -- https://github.com/anthonny[@anthonny]
+
+Only the maintainers have the right to push on the {uri-gh-asciidoctorjs}[GitHub repository].
+
+
+=== npmjs
+
+Maintainers (write access)::
+* Dan Allen -- https://github.com/mojavelinux[@mojavelinux]
+* Olivier Bilodeau -- https://github.com/obilodeau[@obilodeau]
```
obilodeau: Do I really need npmjs write access? I think you guys are going to do fine without me. Unless it's somehow a group and related to asciidoctor-reveal.js I would drop myself out of that. Based on the security principle of "least privilege"

docs/modules/project/pages/governance.adoc
```diff
@@ -0,0 +1,38 @@
+= Asciidoctor.js Governance
```
obilodeau: Using governance confused me at first. To me, this is a board composition or a contact list, not rules defining the project's governance. That said, once I understood that Governance meant to refer to project leaders and core members I was less confused.

docs/modules/project/pages/governance.adoc
Mogztter: That's a good point, I think we should reduce the number of people with write access to just 2 (or maybe 3).
Especially given that releases are now performed by Travis using my token. So if a maintainer with write access to GitHub want to do a release, he can create a tag and push it to trigger a release. So 2 or 3 maintainers on npmjs is enough to recover access or perform a manual release.

docs/modules/project/pages/governance.adoc
Mogztter: I didn't know what else I should add because we don't have meetings and the organization is pretty flat. So it's really "who can do what" page :wink: 
If you think of a better title I'm all ears :ear:

PR #632, opened 2018-12-17
 Map Timings#printReport in the API

Mogztter:
I'm using a custom implementation.
The function takes a "Writable" object as output. It can be one of:

* WritableStream: https://nodejs.org/api/stream.html#stream_writable_streams
* `console`
* `process.stdout`
* Ruby Logger (ie. with a `add` method)


Mogztter:
I will extract the commit "Add create functions (alias for $new())" into a new pull request (so it's easier to review)

Mogztter:
> I will extract the commit "Add create functions (alias for $new())" into a new pull request (so it's easier to review)

Done: https://github.com/asciidoctor/asciidoctor.js/pull/652

Issue #630, opened 2018-12-14
fetch includes asynchronously (in-browser / client-side)


tigregalis:
I'd like to fetch documents and includes asynchronously (in-browser), using `fetch`.

By default, it uses synchronous XMLHttpRequest, which locks the browser and also give us a warning:

    Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help http://xhr.spec.whatwg.org/    asciidoctor.js:23436:8 

That's why I'd like to suggest a loadAsync / convertAsync API for Asciidoctor.js.

Perhaps a related suggestion is a way for the user to specify their own http client, e.g. if someone wishes to use fetch, axios, or node-fetch.

As a really convoluted work-around / proof-of-concept, I've managed to use two include processor extensions and Promises to progressively "load" the document. Right now this will only go down to one level of includes. There are a number of other issues of course, but primarily is that it's loading/parsing the document more than once.

```js
const asciidoctor = Asciidoctor();

const root = document.getElementById('root');

function promiseObject(object) {

  let promisedProperties = [];

  const objectKeys = Object.keys(object);

  objectKeys.forEach((key) => promisedProperties.push(object[key]));

  return Promise.all(promisedProperties)
    .then((resolvedValues) => {
      return resolvedValues.reduce((resolvedObject, property, index) => {
        resolvedObject[objectKeys[index]] = property;
        return resolvedObject;
      }, object);
    });

}

fetch('index.adoc').then(res => res.text()).then(content => {

  const includesMap = {};

  const registry1 = asciidoctor.Extensions.create();
  registry1.includeProcessor(function () {
    var self = this;
    self.handles(_ => true);
    self.process(function (doc, reader, target, attrs) {
      if (!includesMap[target]) includesMap[target] = fetch(target).then(res => res.text());
      return reader.pushInclude('pending...', target, target, 1, attrs);
    });
  });

  asciidoctor.load(content, {'safe': 'safe', extension_registry: registry1});

  promiseObject(includesMap).then(includes => {

    const registry2 = asciidoctor.Extensions.create();
    registry2.includeProcessor(function () {
      var self = this;
      self.handles(_ => true);
      self.process(function (doc, reader, target, attrs) {
        return reader.pushInclude(includes[target], target, target, 1, attrs);
      });
    });

    const doc = asciidoctor.load(content, {'safe': 'safe', extension_registry: registry2});
    const html = doc.convert();
[...]

PR #633, opened 2018-12-21
Introduce an async convert

Mogztter:
The idea is to introduce an experimental API to resolve unconditional includes (first level only). All other includes will be resolved synchronously by Asciidoctor core.

In the benchmark directory I've added a "best case scenario" with 20 unconditional includes.

#### Result

| Method        | Time (ms)
| ------------- |:-------------:|
| `convert`     | 780
| `convertAsync`       | 180 |

Ref: #630 

#### TODO

* [ ] Trigger preprocessor ?
* [ ] Handle or ignore include directive with attributes (leveloffset, lines, tags or indent)
* [ ] Handle or ignore target URI (remote or local include)
* [x] Ignore target that contains `{` (ie. contains a variable that need to be substituted)
* [ ] Trigger include processor or delegate to the processor (ie. if there's an include processor registered for the target, ignore the include directive)
  * ⚠️ we do not have a reader object! 
  * Use a special kind of include processor (that returns a content asynchronously) ?
* [ ] Use the Asciidoctor Logger to print an error when the include is not readable
* [x] Skip block comments


Mogztter:
This could work if we limit the scope but it has many limitations. 
Since we are resolving includes before the parsing the extension registry is not yet activated, attributes are not resolved, the Document and Reader object are not initialized...

Currently if one the following is true we will skip the include directive (ie. Asciidoctor core will resolve the file synchronously):

* If there's a Preprocessor extension
* If there's an Include Processor extension for the target
* If the target contains an attribute
* If the include directive contains leveloffset, tag, tag or lines
* ...

mojavelinux:
We desperately need to split out the logic that selects lines by tags and line numbers. I really don't like to see this logic repeated here (and I don't like that it is repeated in Antora either). This should be a top goal for Asciidoctor 2.0.x, and I think we should :hand: on adding it to Asciidoctor.js until that's available.

mojavelinux:
This really opens some fundamental questions about how includes are processed, and what limits there should be. This is definitely an area I want to focus on in the spec, because right now there are so many sharp edges.

mojavelinux:
One thing is for certain in my mind. This should be marked as highly experimental because it has important consequences we still need to understand ourselves.

Mogztter:
> We desperately need to split out the logic that selects lines by tags and line numbers. I really don't like to see this logic repeated here (and I don't like that it is repeated in Antora either). This should be a top goal for Asciidoctor 2.0.x, and I think we should ✋ on adding it to Asciidoctor.js until that's available.

👍 
I was just experimenting to see how far I can get 😄 

> On the other hand, I don't even think processing line numbers and tags is right here. We should be taking and caching the whole file. All we're really doing is populating a virtual file system so that those lines can be read later by the real include processor. So the filtering should be deferred until then. Isn't that correct?

Sure it's the plan.
I was poking around to see if there's a notable improvement in performance. 
Do you think we should use an include processor ? there can only be one at a time right ? and if we do that we will need to implement the tags/lines filtering.
We've talked about an "include resolver" extension, maybe we could also have an "include reader" (that returns the content for a given target) ?

> This really opens some fundamental questions about how includes are processed, and what limits there should be. This is definitely an area I want to focus on in the spec, because right now there are so many sharp edges.

Indeed, I read the code over and over again and I'm still not sure what happens if you include a remote file and then use a relative include inside this remote file 🤔 
And If you mix that with a browser environment where an URI can be in fact a "local resource" 💥 

> One thing is for certain in my mind. This should be marked as highly experimental because it has important consequences we still need to understand ourselves.

💯 
Potentially it can degraded performance and the current implementation is just a proof of concept so it will be a "use it at your own risk" API 💀

Mogztter:
I will extract the new Registry APIs (`hasPreprocessors`, `getPreprocessors`...) and the conversion from `Opal.nil` to `undefined` in two pull requests.

PR #634, opened 2018-12-21
Upgrade to Opal runtime 1.0.12

Mogztter:
* Ruby parser used by Opal was upgraded
* Opal is now using `unxhr` instead of `xmlhttprequest`
  * Add support for binary data in the OpenURI module
* Upgrade `glob` to 7.1.3
* Restore the `toString` function: https://github.com/Mogztter/opal-node-runtime/issues/4 (thanks @anthonny)

(not related but I've added `eslint` as a dev dependency to remove a warning)


Mogztter:
> Add support for binary data in the OpenURI module

It actually introduces a regression because Asciidoctor core is using the flag `rb` when reading UTF-8 encoded file. Currently reading an utf8 encoded data with the `b` (binary flag) is not decoded properly.

I found a "workaround" but not sure if it's the right way to fix this issue: https://github.com/opal/opal/pull/1921

mojavelinux:
We'll stop reading files this way in Asciidoctor 2.0 (and instead use `::IO.read file, mode: 'r:UTF-8', newline: :universal`). We were using the 'b' flag because Ruby 1.9 would mangle non-UTF-8 documents if you tried to read them without the binary flag on. But that's behind us now.

Mogztter:
> We were using the 'b' flag because Ruby 1.9 would mangle non-UTF-8 documents if you tried to read them without the binary flag on. 

What is the behavior of Ruby 2+ if you use `::IO.read file, mode: 'r:UTF-8', newline: :universal` on a non-UTF-8 document ? I'm asking because we want to make sure that the behavior is the same between MRI and Opal.

I will also prepare the ground because currently `IO.read` does not take argument and encoding is ignored in Opal.

I can't find documentation about `newline: universal`, I guess it should handle both Windows end-of-line (`\r\n`) and Unix end-of-line (`\n`). So if you use `each_line` it will just work right ?

mojavelinux:
> What is the behavior of Ruby 2+ if you use ::IO.read file, mode: 'r:UTF-8', newline: :universal on a non-UTF-8 document ? I'm asking because we want to make sure that the behavior is the same between MRI and Opal.

What I've decided to do is make the read mode a constant so it can be more easily controlled. See https://github.com/asciidoctor/asciidoctor/blob/4b56d8527a759d5972af46e65e5183dd5daf8283/lib/asciidoctor.rb#L205

I'm undecided at this point about the newline option. It's an option that gets inherited from String#encode (but only starting in Ruby 2.5). The idea is that it normalizes all newlines (\n, \r\n, and \r) to \n, so we can support documents that have Unix, Windows, and macOS newlines transparently. However, we already do newline scrubbing, so it seems redundant (though we'd still need it to support macOS newlines).

To answer your question, the 'b' flag is required when reading a non-text file. It creates an ASCII_8BIT string, which is really just a wrapped byte array in Ruby. When reading a text file, it's purpose is only to disable newline normalization (though that's now covered by the newline option starting in Ruby 2.5).

> So if you use each_line it will just work right ?

That's the idea of the newline option.

mojavelinux:
I tried to eliminate the encoding logic in Helpers.prepare_source_string and Helpers.prepare_source_array, but it turns out it's still needed. Ruby doesn't yet provide us with the portable built-in logic we need to support all the cases.

Mogztter:
Updated to Opal Runtime 1.0.13

lib/asciidoctor/js/asciidoctor_ext/browser/abstract_node.rb
```diff
@@ -1,5 +1,29 @@
 module Asciidoctor
 class AbstractNode
+  def read_contents target, opts = {}
+    doc = @document
+    if (Helpers.uriish? target) || ((start = opts[:start]) && (Helpers.uriish? start) && (target = doc.path_resolver.web_path target, start))
+      if (doc.path_resolver.descends_from? target, doc.base_dir) || (doc.attr? 'allow-uri-read')
```
Mogztter: In a browser environment we should allow to read URI if the path is descended from the base directory (even if the `allow-uri-read` attribute is not defined)

lib/asciidoctor/js/asciidoctor_ext/browser/abstract_node.rb
```diff
@@ -1,5 +1,29 @@
...
+        begin
+          if opts[:normalize]
+            (Helpers.prepare_source_string ::File.read(target).join LF)
```
Mogztter: I'm just using `File.read(target)` instead of `::OpenURI.open_uri(target, URI_READ_MODE) {|f| [f.content_type, f.read] }`

lib/asciidoctor/js/asciidoctor_ext/browser/abstract_node.rb
```diff
@@ -1,5 +1,29 @@
...
+          else
+            ::File.read(target)
```
Mogztter: Same here

lib/asciidoctor/js/asciidoctor_ext/browser/abstract_node.rb
mojavelinux: Agreed. I think we made a similar change in the Reader class.

Mogztter: Exactly I've copied the line from the Reader class :smile:

src/template-asciidoctor-umd.js
```diff
@@ -3,43 +3,7 @@ if (typeof Opal === 'undefined' && typeof module === 'object' && module.exports)
 }
 
 if (typeof Opal === 'undefined') {
-  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects#Fundamental_objects
-  var fundamentalObjects = [
```
Mogztter: Opal does not touch fundamental objects anymore :tada: :tada: :tada:

lib/asciidoctor/js/asciidoctor_ext/browser/open_uri.rb
```diff
@@ -0,0 +1,11 @@
+module OpenURI
+  def OpenURI.open_uri(name, *rest)
```
Mogztter: This part assumes that Asciidoctor core will only use `open_uri` when reading a text file from an URI.
Please note that Asciidoctor core uses `open_uri` to read images in `generate_data_uri_from_uri` but this method is overridden in Asciidoctor.js.

mojavelinux:
This upgrade also makes `ENV.key?` available, right?

src/template-asciidoctor-umd.js
mojavelinux: :tada: A day we had only dreamed of before!

Mogztter:
> This upgrade also makes ENV.key? available, right?

Yes and `File.expand_path`. I've also made some improvements to `File.absolute_path` to resolve `~`.

Mogztter:
This pull request contains a breaking change, so I should probably create a `1.5.x` branch before merging even though I don't think we are going to release a `1.5.10` version.

mojavelinux:
I would agree. Now's the time to split off and head down the 2.0.0 road.

PR #635, opened 2018-12-23
Convert Opal.nil to undefined

Mogztter:


src/asciidoctor-core-api.js
```diff
@@ -1486,7 +1486,8 @@ Document.prototype.getCompatMode = function () {
  * @memberof Document
  */
 Document.prototype.getSourcemap = function () {
-  return this.sourcemap
+  var sourcemap = this.sourcemap
+  return sourcemap === Opal.nil ? undefined : sourcemap
```
Mogztter: I'm not sure about this one, maybe we should return false ? The sourcemap option is either true or false.

mojavelinux: Yes, this should be a boolean property.

PR #636, opened 2018-12-23
Map Registry has/get extensions functions

Mogztter:


src/asciidoctor-extensions-api.js
```diff
@@ -257,6 +256,208 @@ Registry.prototype.docinfoProcessor = function (name, processor) {
...
+}
+
+/**
+ * Get any {{@link Extensions/BlockProcessor}} extensions are registered to handle the specified block name appearing on the specified context.
+ * @param name - the {string} block name
+ * @param context - the context of the block: paragraph, open... (optional)
+ * @memberof Extensions/Registry
+ * @returns the Extension proxy object for the {{@link Extensions/BlockProcessor}} that matches the block name and context or undefined if no match is found.
+ */
+Registry.prototype.getBlockFor = function (name, context) {
```
Mogztter: The function returns an extension or undefined if not found. 

In Asciidoctor core, the method `find_block_extension` returns the extension or nil if not found and `registered_for_block?` returns the extension or false if not found.

I believe it's pretty uncommon in JavaScript to return a boolean or the found object. I think it's better to return undefined/null or the found object. If we want to return a boolean value we could also introduce a `hasBlockFor` that returns true if found and otherwise false. (in my opinion it's unnecessary)

PR #642, opened 2018-12-29
Resolves #640, Implement generate_data_uri_from_uri in a Browser environment

### LINKED ISSUE #640: Embedding an image using option data-uri fails in a browser environment
**main.js**
```js
const options = { safe: 'safe', attributes: { 'data-uri': true, 'allow-uri-read': true } }
const html = asciidoctor.convert(`image::http://localhost:5000/image.jpg[]`, options)
```

Note: `allow-uri-read` is required since the image target is a URI.

The following exception is thrown:
```
"content_type: undefined method `content_type' for #<File:0x11a8>
    at $File.$$method_missing (http://localhost:5000/build/asciidoctor-browser.js:3936:56)
    at $File.method_missing_stub (http://localhost:5000/build/asciidoctor-browser.js:1325:35)
    at TMP_26 (http://localhost:5000/build/asciidoctor-browser.js:28305:26)
    at Object.Opal.yield1 (http://localhost:5000/build/asciidoctor-browser.js:1467:14)
    at $Block.$$open (http://localhost:5000/build/asciidoctor-browser.js:23967:21)
    at Opal.send (http://localhost:5000/build/asciidoctor-browser.js:1686:19)
    at $Block.$$generate_data_uri_from_uri (http://localhost:5000/build/asciidoctor-browser.js:28298:21)
    at $Block.$$image_uri (http://localhost:5000/build/asciidoctor-browser.js:28222:27)
    at $Html5Converter.$$image (http://localhost:5000/build/asciidoctor-browser.js:39465:68)
    at $Html5Converter.$$__send__ (http://localhost:5000/build/asciidoctor-browser.js:3731:21)"
```

The reason is that the module `open-uri` is not loaded in a browser environment: https://ruby-doc.org/stdlib-2.4.0/libdoc/open-uri/rdoc/OpenURI/Meta.html#method-i-content_type

Mogztter:


lib/asciidoctor/js/asciidoctor_ext/browser/abstract_node.rb
```diff
@@ -0,0 +1,61 @@
...
+              var byteCode = c & 0xff // byte at offset i
+              binary += String.fromCharCode(byteCode)
+            }
+            b64encoded = btoa(binary)
+            contentType = this.getResponseHeader('content-type')
+          }
+        })
+        xhr.send(null)
+        // try to detect the MIME Type from the file extension
+        if (!contentType) {
```
Mogztter: `content-type` is undefined when using the `file` protocol.
The above code will try to detect the MIME type from the file extension. It can return an incorrect value on edge cases but overall I think it provides a better user experience.

Mogztter:
This pull request is ready to go!

PR #644, opened 2018-12-29
Introduce an async convert

ggrossetie:
* The `Reader` class in Asciidoctor core has a new method `read_include_content`
  * If a VFS is defined, read the content of the include target from it. Otherwise use the file system or the network to read the content.
* The `convertAsync` function will fetch the content of all the unconditional includes and populate the VFS.
* If an include processor is defined for the target, the content will not be fetched
* If the target already exists in the VFS skip (cache)

Performance wise it about the same speed for local files but since the event loop is not blocked (or at least less blocked), the performance should be better on a running application.

I think we should be careful about how many asynchronous remote calls we can make at once. We may need to set a limit to avoid resource exhaustion.

 Using a cache greatly improve performance if a remote file is present more than once in the document. For reference `open-uri/cached` module used in Asciidoctor Ruby is not available in JavaScript.


lib/asciidoctor/js/opal_ext/reader.rb
```diff
@@ -0,0 +1,209 @@
...
+          return replace_next_line %(Unresolved directive in #{@path} - include::#{expanded_target}[#{attrlist}])
+        end
+      end
+      true
+    end
+  end
+
+  # If a VFS is defined, Asciidoctor will use it to resolve the include target.
+  # Otherwise use the file system or the network to read the file.
+  def read_include_content inc_path, target_type
```
Mogztter: @mojavelinux `preprocess_include_directive` method is almost identical, I've just replaced `File.open` with the new `read_include_content` method.

mojavelinux:
> I think we should be careful about how many asynchronous remote calls we can make at once. We may need to set a limit to avoid resource exhaustion.

It's annoying that Node.js doesn't provide this out of the box. I run into the same problem whenever I switch to async operations.

mojavelinux:
I'm not sure that overriding/importing large chunks of Asciidoctor core is the right idea here. But I'm also still at a loss for exactly how to handle this. We don't really have a true preprocessor that avoids side effects in Asciidoctor, and the include directive is inherently synchronous. This is topic I think we need to consider in the AsciiDoc Language project, then carry out whatever we decide after the architecture is worked out there.

cceneag (approved):
👍

cceneag:
👍

PR #645, opened 2019-01-05
Prepare Asciidoctor 2.0.0

Mogztter:
This pull request is compatible with: https://github.com/asciidoctor/asciidoctor/pull/2946

```
ASCIIDOCTOR_CORE_VERSION=mojavelinux/asciidoctor#issue-2764-no-unmaintained-rubies npm run build:quick
```

### DIFF
--- a/lib/asciidoctor/js.rb
+++ b/lib/asciidoctor/js.rb
@@ -1,4 +1,3 @@
 require 'asciidoctor/js/opal_ext'
 require 'asciidoctor/js/rx'
 require 'strscan'
-require 'logger'
--- a/lib/asciidoctor/js.rb
+++ b/lib/asciidoctor/js.rb
@@ -1,3 +1,4 @@
 require 'asciidoctor/js/opal_ext'
 require 'asciidoctor/js/rx'
 require 'strscan'
+require 'logger'
--- a/lib/asciidoctor/js/opal_ext/kernel.rb
+++ b/lib/asciidoctor/js/opal_ext/kernel.rb
@@ -1,0 +1,15 @@
+module Kernel
+  # basic implementation of open, enough to work with reading files over XMLHttpRequest
+  def open(path, *rest)
+    file = File.new(path, *rest)
+    if block_given?
+      yield file
+    else
+      file
+    end
+  end
+  def __dir__
+    # fake implementation
+    ''
+  end
+end
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ addons:
   chrome: stable
 
 env:
-- ASCIIDOCTOR_CORE_VERSION=v1.5.8
+- ASCIIDOCTOR_CORE_VERSION=mojavelinux/asciidoctor#issue-2764-no-unmaintained-rubies
 - ASCIIDOCTOR_CORE_VERSION=master
 
 node_js:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ addons:
   chrome: stable
 
 env:
-- ASCIIDOCTOR_CORE_VERSION=mojavelinux/asciidoctor#issue-2764-no-unmaintained-rubies
+- ASCIIDOCTOR_CORE_VERSION=v1.5.8
 - ASCIIDOCTOR_CORE_VERSION=master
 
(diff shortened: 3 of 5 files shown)


mojavelinux:
Actually, I might put back that `require 'logger'` statement since it doesn't follow the scenario of the other autoloads. We always require logging, so there's no benefit to using autoload here.

mojavelinux:
We no longer use the Kernel.open function, so although it doesn't hurt to keep it, that could be removed.

mojavelinux:
We no longer use the Kernel.open function, so although it doesn't hurt to keep it, that could be removed.

mojavelinux:
The only modifications that seem to be needed are:

* Mapping File.absolute_path to File.expand_path
* Mapping Kernel.__dir__

The only time either of these should be used by Asciidoctor.js is when defining the ROOT_DIR constant. I wonder if that constant should be defined by the asciidoctor/js require. Then, we could avoid this whole code path.

mojavelinux:
I'm mulling over the constant names FILE_READ_MODE and URI_READ_MODE since they become part of the integration API. Ruby calls these "open modes". However, they aren't just any open mode, but specifically for reads. Do you think these names are clear, or should we name them something like FILE_READ_OPEN_MODE? (I might be overthinking this).

Mogztter:
> The only time either of these should be used by Asciidoctor.js is when defining the ROOT_DIR constant. I wonder if that constant should be defined by the asciidoctor/js require. Then, we could avoid this whole code path.

Yes I think it's a good idea to define this constant in `asciidoctor/js` require.
We should also do the same with `STYLESHEETS_DATA_PATH` (instead of patching the code with a regexp).

> I'm mulling over the constant names FILE_READ_MODE and URI_READ_MODE since they become part of the integration API. Ruby calls these "open modes". However, they aren't just any open mode, but specifically for reads. Do you think these names are clear, or should we name them something like FILE_READ_OPEN_MODE? (I might be overthinking this).

In my opinion, `FILE_READ_MODE` and `URI_READ_MODE` are fine :grinning:

mojavelinux:
> Yes I think it's a good idea to define this constant in asciidoctor/js require.

I suppose we could add:

```
ROOT_DIR = ::File.dirname ::File.absolute_path __dir__ unless defined? ROOT_DIR
```

wdyt?

> In my opinion, FILE_READ_MODE and URI_READ_MODE are fine grinning

:+1:

mojavelinux:
For stylesheets, I think it's more correct to override the methods. The constants in that class should be considered internal. You don't have to satisfy them. You just have to satisfy the Stylesheets API. In fact, we could even stop loading it in Opal (so you have total control).

lib/asciidoctor/js.rb
```diff
@@ -1,3 +1,4 @@
 require 'asciidoctor/js/opal_ext'
 require 'asciidoctor/js/rx'
 require 'strscan'
+require 'logger'
```
mojavelinux: This is now loaded by default again.

lib/asciidoctor/js/opal_ext/kernel.rb
```diff
@@ -0,0 +1,15 @@
+module Kernel
+  # basic implementation of open, enough to work with reading files over XMLHttpRequest
+  def open(path, *rest)
```
mojavelinux: We'll no longer be using Kernel.open in Asciidoctor core starting in 2.0.

mojavelinux:
> Then I could override the method `primary_stylesheet_path` to resolve the path according to the JavaScript environment.

I'd even go up one level. How about `stylesheet_data_path`? That way, we can use that method to get any stylesheet we happen to bundle.

> I'm currently debugging the bom test failure. 

I may have made a mistake here. I'm looking into it.

mojavelinux:
I was using slice instead of byteslice. slice relies on how bytes are interpreted in the encoding. What we want is byteslice, which just rips out bytes 1-by-1.

mojavelinux:
PR updated.

Mogztter:
Build is green, should I move forward and merge this pull request before asciidoctor/asciidoctor#2946 ?

mojavelinux:
Yep! I'm about to merge that PR.

PR #648, opened 2019-01-09
Do not remove the build directory when calling 'npm run build'

Mogztter:
### Build command

`npm run build` does not remove the `build` directory anymore.

Also if a file already exists, the _build_ task will use this file (instead of generating it again).
For instance:

* If `build/asciidoctor.tar.gz` exists, the file will not be downloaded again from GitHub
* If `build/asciidoctor` exists, the file `build/asciidoctor.tar.gz` will not be extracted again
* If `build/asciidoctor-core.js` exists, Asciidoctor core will not be compiled again
* If `build/opal-ext-node.js` exists, `lib/asciidoctor/js/opal_ext/node.rb` will not be compiled again
* ...

### New clean commands

4 new commands to clean the workspace:

#### Clean all

`npm run clean:all` (alias `npm run clean`) will remove the `build` directory

#### Remove the compiled version of Asciidoctor core

`npm run clean:core` will remove the compiled version of Asciidoctor core (Ruby).
 
#### Remove the compiled "flavors" of Asciidoctor.js

`npm run clean:lib` will remove every `.js` files in the `build` directory except `asciidoctor-core.js`.

#### Remove all the compiled files

`npm run clean:js` will remove every `.js` files in the `build` directory.

#### Clean build

```sh
$ npm run clean && npm run build # long format: npm run clean:all && npm run build
```

Since the `build` directory will be removed, the *build* task will download Asciidoctor core again from GitHub.

#### Build Asciidoctor.js against a local (modified) version of Asciidoctor core

```sh
$ npm run clean:core && npm run build
```

The file `build/asciidoctor-core.js` will be removed but the directory `build/asciidoctor` (local copy of Asciidoctor core) will not. The _build_ task will use the local copy to generate again the file `build/asciidoctor-core.js`.

#### Add a new feature in Asciidoctor.js

```sh
$ npm run clean:lib && npm run build
```

The file  `build/asciidoctor-core.js` will not be removed. The goal is to save time as the compilation of Asciidoctor core can take up to a dozen of seconds. Other JavaScript files in the `build` directory will be remove. The _build_ task will use the sources of Asciidoctor.js (in the `src` and `lib` directory) to generate all the flavor of Asciidoctor.js (Node.js, Browser, GraalVM...).

If you've modified both the local copy of Asciidoctor core and the sources of Asciidoctor.js then:

```sh
$ npm run clean:js && npm run build
```

//cc @mojavelinux


Mogztter:
> The only thing I'm not sure about is clean:lib

Naming is hard 😰 
I was using `lib` to refer to the Asciidoctor.js library, maybe pkg (package) ?

> It's not clear to me what is being cleaned here. If a file under lib/ is changed, I would think core has to be transpiled again anyway. Perhaps you mean the dists?

Not necessarily.
For instance when we need to override a method in Asciidoctor core, we want to regenerate `asciidoctor.js` but we also want to reuse `asciidoctor-core.js` to save time (because Asciidoctor code didn't change and the transpilation takes some time).
The same thing is true when we add a new API to Asciidoctor.js, we just want to regenerate the Asciidoctor.js distribution without transpiling Asciidoctor core again.

If we decompose the `build` task:

1. Download Asciidoctor core (produced `build/asciidoctor.tar.gz` file)
1. Decompress Asciidoctor core (produced `build/asciidoctor` directory)
1. Transpile Asciidoctor core (produced `build/asciidoctor-core.js` file)
1. Transpile Asciidoctor.js (produced `build/opal-ext-*.js` files)
1. Generate the "distribution" (produced `asciidoctor-node.js`, `asciidoctor-browser.js` files...)

The idea is really to run the build task again at a specific step.

* `clean:all` or `clean`, starts again at step 1
* `clean:core`, starts again at step 3 (skip 4)
* `clean:lib`, starts again at step 4
* `clean:js`, starts again at step 3

We don't have task to start again at step 5. It could be useful when we just want to add/delete/modify an Asciidoctor.js API (ie. just concat the files again and run tests suite).

Does that make sense ?

mojavelinux:
In that case, I would call it `clean:patch` since that's what lib really is.

Mogztter:
`patch` is sad, maybe `api` ? or `ext` ? or `happy-little-patch` 🤣 

bob-ross-3

mojavelinux:
You are too funny. I think patch is a happy place, though. Just a bit more fluffy clouds. :paintbrush:

PR #658, opened 2019-01-26
resolves #661 issue a debug when the syntax highlighter is not found

### LINKED ISSUE #661: Issue a warning if a syntax highlighter is not found ?
Asciidoctor core now provides a pluggable system for syntax highlighters. Basically you can now register syntax highlighters or even provide a factory to resolve syntax highlighters.

Out of the box Asciidoctor.js provides 2 syntax highlighters:

* `highlight.js` (also known as `highlightjs`, though this name will probably be deprecated in the future)
* `html-pipeline`

Previously, if `coderay` or `pygment` were used the following warning message was issued:

```rb
optional gem '#{gem_name}' is not installed. Functionality disabled.
```

But with the new pluggable system `coderay` and `pygment` are no longer hard-wired into core so the warning message is gone.

So now If I'm using the following AsciiDoc document:

```
:syntax-highlighter: coderay

[source,js]
----
console.log('hello');
----
```

Asciidoctor.js will silently ignore the syntax highlighter and produce the following output:

```html
<div id="content">
  <div class="listingblock">
    <div class="content">
      <pre class="highlight"><code class="language-js" data-lang="js">console.log('hello');</code></pre>
    </div>
  </div>
</div>
```

Should we add a warning message if the requested syntax highlighter is not available ?

Mogztter:
Resolves #661

### DIFF
--- a/packages/core/lib/asciidoctor/js/asciidoctor_ext.rb
+++ b/packages/core/lib/asciidoctor/js/asciidoctor_ext.rb
@@ -2,6 +2,7 @@ require 'asciidoctor/js/asciidoctor_ext/stylesheet'
 require 'asciidoctor/js/asciidoctor_ext/document'
 require 'asciidoctor/js/asciidoctor_ext/substitutors'
 require 'asciidoctor/js/asciidoctor_ext/parser'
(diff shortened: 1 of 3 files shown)


Mogztter:
Let's go with a debug message.

PR #659, opened 2019-01-26
Create a pure browser bundle

Mogztter:
* Do not pollute the browser bundle with Node specific code.
* Explain how to minify `asciidoctor-browser.js` with Google Closure Compiler

```
1,8M janv. 26 15:49 asciidoctor-browser.js
632K janv. 26 15:49 asciidoctor-browser.min.js
```

### DIFF
--- a/docs/modules/setup/pages/browser-optimization.adoc
+++ b/docs/modules/setup/pages/browser-optimization.adoc
@@ -1,0 +1,25 @@
+= Asciidoctor.js for the browser
+
+Asciidoctor.js produces bundles that target different environments.
+The browser bundle is _not_ minified and includes the Opal runtime code.
+
+On this page, you'll learn how to optimize Asciidoctor.js for the browser.
+
+== Google Clojure Compiler
+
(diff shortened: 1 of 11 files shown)


lib/asciidoctor/js/asciidoctor_ext/node/abstract_node.rb
```diff
@@ -6,7 +6,7 @@ class AbstractNode
   def generate_data_uri_from_uri image_uri, cache_uri = false
     %x{
       var xhr = new __XMLHttpRequest__();
-      xhr.open('GET', uri, false);
+      xhr.open('GET', image_uri, false);
```
Mogztter: The good thing about running a compiler is that it even found a bug!

Mogztter: Test added and bug fixed in https://github.com/asciidoctor/asciidoctor.js/pull/662

mojavelinux: Nice!

PR #660, opened 2019-01-26
Asciidoctor core no longer issues a warning

Mogztter:


Mogztter:
I'm fixing the build on master so we can then discuss if we want to add a warning message or not.

PR #669, opened 2019-02-10
resolves #654 remove Time usage

### LINKED ISSUE #654: Do not use Time module to reduce the Opal runtime size
https://ruby-doc.org/stdlib-2.6/libdoc/time/rdoc/Time.html

We will need to provide our own implementation of [`Asciidoctor::Timings`](https://github.com/asciidoctor/asciidoctor/blob/e6ff47641a0e4bcc5d9e5271eaaf5b6389f5cacc/lib/asciidoctor/timings.rb) and Asciidoctor core is willing to abstract the following code:

https://github.com/asciidoctor/asciidoctor/blob/e6ff47641a0e4bcc5d9e5271eaaf5b6389f5cacc/lib/asciidoctor/document.rb#L490-L506

Mogztter:
resolves #654

* Built-in Ruby Logger is now using JavaScript Date
* Implement Document#fill_datetime_attributes

We should be able to remove the Opal module: `corelib/time`.

### DIFF
--- a/lib/asciidoctor/js/opal_ext/number.rb
+++ b/lib/asciidoctor/js/opal_ext/number.rb
@@ -1,0 +1,10 @@
+class Number < Numeric
+  def round(ndigits = undefined)
+    ndigits = Opal.coerce_to!(ndigits, Integer, :to_int)
+    if ndigits > 0
+      `Number(self.toFixed(ndigits))`
+    else
+      `Math.round(self)`
+    end
+  end
+end
--- a/lib/asciidoctor/js.rb
+++ b/lib/asciidoctor/js.rb
@@ -1,3 +1,5 @@
 require 'asciidoctor/js/opal_ext'
 require 'asciidoctor/js/rx'
+# it should be required first to resolve the alias in Asciidoctor core
+require 'asciidoctor/js/substitutors'
 require 'strscan'
--- a/lib/asciidoctor/js.rb
+++ b/lib/asciidoctor/js.rb
@@ -1,5 +1,3 @@
 require 'asciidoctor/js/opal_ext'
 require 'asciidoctor/js/rx'
-# it should be required first to resolve the alias in Asciidoctor core
-require 'asciidoctor/js/substitutors'
 require 'strscan'
--- a/lib/asciidoctor/js/asciidoctor_ext.rb
+++ b/lib/asciidoctor/js/asciidoctor_ext.rb
@@ -1,5 +1,6 @@
 require 'asciidoctor/js/asciidoctor_ext/stylesheet'
 require 'asciidoctor/js/asciidoctor_ext/substitutors'
+require 'asciidoctor/js/asciidoctor_ext/document'
 require 'asciidoctor/js/asciidoctor_ext/parser'
 
 %x(
--- a/lib/asciidoctor/js/asciidoctor_ext.rb
+++ b/lib/asciidoctor/js/asciidoctor_ext.rb
@@ -1,5 +1,6 @@
 require 'asciidoctor/js/asciidoctor_ext/stylesheet'
 require 'asciidoctor/js/asciidoctor_ext/document'
+require 'asciidoctor/js/asciidoctor_ext/substitutors'
 require 'asciidoctor/js/asciidoctor_ext/parser'
 
 %x(
--- a/lib/asciidoctor/js/asciidoctor_ext.rb
+++ b/lib/asciidoctor/js/asciidoctor_ext.rb
@@ -1,5 +1,4 @@
 require 'asciidoctor/js/asciidoctor_ext/stylesheet'
-require 'asciidoctor/js/asciidoctor_ext/substitutors'
 require 'asciidoctor/js/asciidoctor_ext/document'
 require 'asciidoctor/js/asciidoctor_ext/parser'
 
--- a/lib/asciidoctor/js/asciidoctor_ext/document.rb
+++ b/lib/asciidoctor/js/asciidoctor_ext/document.rb
@@ -1,0 +1,116 @@
+module Asciidoctor
+class Document < AbstractBlock
+
+  def fill_datetime_attributes attrs, input_mtime
+    %x{
+      var $truthy = Opal.truthy
+      var $falsy = Opal.falsy
+      var nil = Opal.nil
+      var utc_offset
+      var source_date_epoch
+
+      var getYear = function (time, utc_offset) {
+        return utc_offset === 0 ? time.getUTCFullYear() : time.getFullYear()
+      }
+      var getMonth = function (time, utc_offset) {
+        return utc_offset === 0 ? time.getUTCMonth() : time.getMonth()
+      }
+      var getDay = function (time, utc_offset) {
+        return utc_offset === 0 ? time.getUTCDate() : time.getDate()
+      }
+      var getHours = function (time, utc_offset) {
+        return utc_offset === 0 ? time.getUTCHours() : time.getHours()
+      }
+
+      var now = new Date()
+      // See https://reproducible-builds.org/specs/source-date-epoch/
+      if (Opal.const_get_qualified('::', 'ENV')['$key?']('SOURCE_DATE_EPOCH')) {
+        now.setTime(parseInt(Opal.const_get_qualified('::', 'ENV')['$[]']('SOURCE_DATE_EPOCH')) * 1000)
+        source_date_epoch = now
+        utc_offset = 0  // utc
+      } else {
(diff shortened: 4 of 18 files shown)


Mogztter:
Upstream issue: https://github.com/asciidoctor/asciidoctor/pull/3065 
If this pull request is merged, update the `fill_datetime_attributes` implementation accordingly.

Mogztter:
> If this pull request is merged, update the `fill_datetime_attributes` implementation accordingly.

Done

Mogztter:
It should reduce the raw size by ~ 25Ko

Mogztter:
Again some improvements regarding the size:

```
1704 build/asciidoctor-browser.js
596 build/asciidoctor-browser.min.js
168 build/asciidoctor-browser.min.js.gz
```

mojavelinux:
The hack for sub_placeholder can easily be resolved by adding ` unless RUBY_ENGINE == 'opal'` to the alias in core. I'll go ahead and do that.

I think in order for it to work when sprintf is not loaded, you'd have to provide an empty stub for sprintf on Kernel. Once that's available, then the alias can be overridden.

This code demonstrates that:

```rb
module Kernel
  def sprintx; end
end

module Substitutors
  alias sub_placeholder sprintx

  def xreftext
    sub_placeholder '%s', 'original'
  end
end

class AbstractNode
  include Substitutors
end

class Block < AbstractNode
end

module Substitutors
  def sub_placeholder fmt, str
    'override'
  end
end

puts Block.new.xreftext
```

If you leave out the definition of sprintx, then it will fail.

mojavelinux:
Upstream updated.

Mogztter:
> Upstream updated.

I will update the code accordingly.

Note: this is a "breaking change" (ie. this change is not compatible with Asciidoctor core Ruby 1.5.8 )

Mogztter:
I'll branch and remove the build against 1.5.8 from Travis.

mojavelinux:
Are you saying that removing Time is a breaking change for 1.5.x? If that's the case, I thought 1.5.x was already a branch. There's no question in my mind that master should be only concerned with Asciidoctor 2.0.

Mogztter:
**pull request** (Node.js 10.15.1 and Asciidoctor master)
```
Load scripts: 0.073s
Run #1: 0.671s
Run #2: 0.491s
Run #3: 0.429s
Run #4: 0.41s
Run #5: 0.449s
Run #6: 0.394s
Run #7: 0.389s
Run #8: 0.376s
Run #9: 0.387s
Run #10: 0.429s
Run #11: 0.394s
Run #12: 0.41s
Run #13: 0.395s
Run #14: 0.39s
Run #15: 0.42s
```

**v1.5.9 tag** (Node.js 10.15.1 and Asciidoctor v1.5.8)
```
Load scripts: 0.083s
Run #1: 0.648s
Run #2: 0.466s
Run #3: 0.457s
Run #4: 0.42s
Run #5: 0.438s
Run #6: 0.382s
Run #7: 0.384s
Run #8: 0.374s
Run #9: 0.406s
Run #10: 0.389s
Run #11: 0.382s
Run #12: 0.377s
Run #13: 0.406s
Run #14: 0.373s
Run #15: 0.376s
```

The script load a bit faster but `convert` is a bit slower.
I will try on the branch 1.5.9 to understand why it's slower.

Mogztter:
**branch 1.5.9** (Node.js 10.15.1 and Asciidoctor v1.5.8)

```
Load scripts: 0.087s
Run #1: 0.732s
Run #2: 0.516s
Run #3: 0.48s
Run #4: 0.436s
Run #5: 0.453s
Run #6: 0.425s
Run #7: 0.416s
Run #8: 0.402s
Run #9: 0.452s
Run #10: 0.411s
Run #11: 0.406s
Run #12: 0.42s
Run #13: 0.459s
Run #14: 0.4s
Run #15: 0.403s
```

Branch 1.5.9 is even slower, so something in Asciidoctor.js and/or in Opal is slower.

Mogztter:
**branch 1.5.9** (Node.js 10.15.1 and Asciidoctor master)
```
Load scripts: 0.082s
Run #1: 0.647s
Run #2: 0.474s
Run #3: 0.439s
Run #4: 0.425s
Run #5: 0.432s
Run #6: 0.395s
Run #7: 0.39s
Run #8: 0.374s
Run #9: 0.394s
Run #10: 0.417s
Run #11: 0.383s
Run #12: 0.383s
Run #13: 0.39s
Run #14: 0.388s
Run #15: 0.413s
```

Numbers are not very stable so it's hard to say for sure that Asciidoctor.js 1.5.9 (built against 1.5.8) is faster than this pull request.
But at the time, Asciidoctor.js 1.5.9 is the only version where I can get multiple `convert` under 0.38 seconds...

mojavelinux:
If it's anything, it could be this change:

> use count idiom to compare uniform lines

count is accelerated in CRuby, but we may not see that benefit in Opal. We could do something similar to what we did for sprintf. Perhaps bisect on that commit, which is `569e9e229b33de708038348dc94869da342dbaa8` (and perhaps ones right around it).

Mogztter:
Thanks for the insight, I will do bisect 👍

Mogztter:
I've added stats and I'm running 75 `convert`:

**pull request** (Node.js 10.15.1 and Asciidoctor master)
```
avg: 387.67
stddev: 32.87
50th: 380.00
75th: 389.00
90th: 409.20
99th: 504.76
```

**v1.5.9 tag** (Node.js 10.15.1 and Asciidoctor v1.5.8)
```
avg: 386.36
stddev: 32.66
50th: 378.00
75th: 389.50
90th: 407.80
99th: 499.24
```

Mogztter:
I can confirm that https://github.com/asciidoctor/asciidoctor/commit/569e9e229b33de708038348dc94869da342dbaa8 has a significant negative effect on performance.

**pull request** (Node.js 10.15.1 and Asciidoctor master revert count)

```
avg: 378.35
stddev: 33.63
50th: 368.00
75th: 381.00
90th: 402.80
99th: 496.90
```

mojavelinux:
If I create a branch upstream with an alternative approach, would you be willing to test it?

Mogztter:
Sure 👍

Mogztter:
@mojavelinux Overall it's a bit faster with your new approach. Fastest run:
```
avg: 370.80
stddev: 33.15
50th: 363.00
75th: 372.00
90th: 382.40
99th: 492.98
```

With a convert at 0.348s :zap:

mojavelinux:
That is fantastic news. I'll go ahead and proceed with the upstream change, then.

The other place we could have issues (which is not new) is that we use `count LF` to count the number of newlines. Here we might want to use `(split LF).size - 1` in Opal. However, it's not used very often, so I'm not sure if it will make much of a difference, so we'll hold off on that. Just something to look out for.

lib/asciidoctor/js/opal_ext/number.rb
```diff
@@ -0,0 +1,10 @@
+class Number < Numeric
+  def round(ndigits = undefined)
```
Mogztter: @mojavelinux do you think we need a more sophisticated implementation ? As far as I know Asciidoctor core only uses this function to round the column/table size.

Mogztter:
#### Summary

⚡️ resolves #654 switch to Asciidoctor Opal Runtime (#669)

* Remove Time usage
* Built-in Ruby Logger is now using JavaScript Date
* Implement Document#fill_datetime_attributes
* Switch to Asciidoctor Opal Runtime (specifically designed for Asciidoctor)
* Remove sprintf hack
* Implement base64
* Remove format in Logger
* Implement a naive round method

lib/asciidoctor/js/opal_ext/number.rb
mojavelinux: That seems to do the trick. If all column width tests pass, I'd say we're good.

Mogztter: > That seems to do the trick. If all column width tests pass, I'd say we're good.

I've added one, I will check the Asciidoctor tests suite to add more.

Mogztter:
Do you think we should add the minified version of asciidoctor.js (600kb) to the npm package ? Otherwise I can update the release task and add the file to the GitHub release page (so a user won't have to install Java + Google Closure to minify Asciidoctor.js) ?

I don't think we should provide a gzip version as this can be done by a reverse proxy or a web server.

mojavelinux:
> Do you think we should add the minified version of asciidoctor.js

Up to you.

> I don't think we should provide a gzip version as this can be done by a reverse proxy or a web server.

Agreed.

PR #672, opened 2019-02-19
Drop :ids table from document catalog

Mogztter:
Upstream: https://github.com/asciidoctor/asciidoctor/issues/3084

/cc @mojavelinux


mojavelinux:
Yep. If for some reason you want to keep the API, it's possible to rebuild the table from the :refs table. See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/document.rb#L610-L613

Mogztter:
No I think it's OK to remove this API.
User can easily migrate to `refs` and to be honest I don't think anyone is using it.

PR #678, opened 2019-02-24
Fix master version in Travis

Mogztter:
Proc#binding is not implemented in Opal.
Temporary fix to workaround: https://github.com/asciidoctor/asciidoctor/commit/60d27edec4bb033a5cf529801cbd9ebc0dab8c6f#r32447609

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ addons:
   chrome: stable
 
 env:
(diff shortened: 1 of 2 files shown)


mojavelinux:
I just fixed the code upstream that should allow Asciidoctor.js to continue working. Even better, no functionality is lost (I think).

PR #685, opened 2019-02-28
resolves #683 provide a JavaScript implementation for Parser#uniform?

### LINKED ISSUE #683: Provide a JavaScript implementation for Parser#uniform?
Initial discussion: https://github.com/asciidoctor/asciidoctor/pull/3098

```rb
module Asciidoctor
class Parser
  if `String.prototype.repeat`
    def self.uniform? str, chr, len
      `chr.repeat(len) === str`
    end
  else
    def self.uniform? str, chr, len
      `Array.apply(null, { length: len }).map(function () { return chr }).join('') === str`
    end
  end
end
end
```

Mogztter:
resolves #683 

//cc @mojavelinux


mojavelinux:
:+1:

PR #692, opened 2019-03-06
resolves #675 make it clear that some methods on Reader do not consume the lines...

### LINKED ISSUE #675: Make it clear that some methods on Reader do not consume the lines
And some other methods consume the lines. In other words, if you use `getString` you will get a copy of the lines and you will be able to read the same lines again. But if you use `read` you will consume the lines.

Mogztter:
... and some others do!

resolves #675

### DIFF
--- a/src/asciidoctor-core-api.js
+++ b/src/asciidoctor-core-api.js
@@ -1782,6 +1782,8 @@ Reader.prototype.getLines = function () {
 /**
  * Get the remaining lines managed by this Reader as a {string}.
  *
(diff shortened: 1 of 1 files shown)


mojavelinux:
The only additional point I would add is that any method that returns lines without consuming them (except for peek_lines) is reading the unprocessed lines. So your options are really:

* return the remaining lines unprocessed without consuming them (e.g., getLines())
* consume, preprocess, and return the remaining lines (e.g., readLines())
* preprocess and return without consuming the remaining lines (e.g., peekLines())

Mogztter:
I have added this information and used the same terms consistently.

PR #694, opened 2019-03-11
Rename asciidoctor.js to asciidoctor

### LINKED ISSUE #687: Consider renaming the package
The current name `asciidoctor.js` is inconsistent with others npm packages. More precisely, the `.js` suffix is redundant.

My instinct is to type:

```js
const asciidoctor = require('asciidoctor')
```

Furthermore, there's quite a few "asciidoctor" packages published on npmjs: https://www.npmjs.com/search?q=asciidoctor

To help users finding the right package, we could publish "official" package in the `@asciidoctor` organization:

* `@asciidoctor/asciidoctor`
* `@asciidoctor/cli`
* `@asciidoctor/backend-pdf`
* `@asciidoctor/backend-revealjs`
* `@asciidoctor/backend-bespoke`
* ...

And here's a few proposals, we are considering:
* `asciidoctor`
* `@asciidoctor/asciidoctor`
* `@asciidoctor/core`
* `@asciidoctor/processor`

What do you think ? should we do it for 2.0.0 ?

Mogztter:
I moved almost everything under `packages/core`.
To release the project we will need to do the following:

1. Prepare the release (local):
   * Set the version in both `package.json` and `packages/core/package.json`
   * Update the dependency in `package.json` (replace `file:packages/core` with the version)
2. Commit, tag and push
2. Travill will build and release `@asciidoctor/core` (the process is the same as before)
3. Travis will run `npm publish` at the root

Resolves #687


Mogztter:
I did some experiments but I'm not sure how to create an "alias" (and if it's a good idea anyway).
For now the `asciidoctor` package is just an empty package with dependencies. So the following won't work:

```js
require('asciidoctor')
```

The reason is that the package `asciidoctor` does not contain any files (and therefore does not define a `main` field).

The following is working though:

```
$ npm i asciidoctor
```

```js
require('@asciidoctor/core')
```

I could of course copy the _dist_ directory in the meta package but it feels wrong because we basically duplicate the package...

Mogztter:
If we mimic Gatsby then `npm i asciidoctor` will install the core library and the CLI.
Brief digression, we should probably rename the binary to "asciidoctor".

So it doesn't give use `require('asciidoctor')` but the user will be able to run the `asciidoctor` CLI and/or use the core library `require('@asciidoctor/core')`.

Mogztter:
I guess we could create an `index.js` in the `asciidoctor` package file with the following content:

```js
module.exports = require('@asciidoctor/core')
```

We introduce an extra export (maybe not that bad) and for browser environment we will still need to use `@asciidoctor/core` because `asciidoctor` package is an empty shell.

```html
<script src="https://unpkg.com/@asciidoctor/core@2.0.0/dist/browser/asciidoctor.js"/>
```

mojavelinux:
:tada:

PR #700, opened 2019-03-25
Map a few SyntaxHighlighter API

Mogztter:


packages/core/spec/share/asciidoctor-spec.js
```diff
@@ -1232,6 +1232,31 @@ April,32
...
+    describe('Syntax Highligther', function () {
+      it('should return the syntax highlighter class registered for the specified name', () => {
+        const highlightjsSyntaxHighlighter = asciidoctor.SyntaxHighlighter.for('highlight.js')
+        expect(highlightjsSyntaxHighlighter.$$name).to.equal('HighlightJsAdapter')
+        const rougeSyntaxHighlighter = asciidoctor.SyntaxHighlighter.for('rouge')
+        expect(rougeSyntaxHighlighter).to.be.undefined()
+      })
+
+      it('should create a new syntax highlighter', () => {
+        asciidoctor.SyntaxHighlighterBase.create('unavailable', function () {
```
Mogztter: My goal was to be as close as: https://github.com/asciidoctor/asciidoctor/blob/7d29276eb1db617ade8055f119b86db002d6f4b6/test/syntax_highlighter_test.rb#L66

For some reason, Asciidoctor.js is throwing the following exception if the name is not defined: `asciidoctor: FAILED: <stdin>: Failed to load AsciiDoc document - unavailable must specify a value for `name'`.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -2129,3 +2129,111 @@ Timings.prototype.printReport = function (to, subject) {
...
+  Opal.def(this, '$format', block)
+}
+
+/**
+ * Indicates whether highlighting is handled by this syntax highlighter or by the client.
+ *
+ * @memberof SyntaxHighlighterBase
+ * @param {Function} block - A {Function} that returns a {Boolean} indicating whether the highlight method should be used to handle the specialchars substitution.
+ */
+SyntaxHighlighterBase.prototype.highlighted = function (block) {
```
Mogztter: I'm not sure about the naming, `isHighlight` feels weird, what do you think ?

packages/core/src/asciidoctor-core-api.js
```diff
@@ -2129,3 +2129,111 @@ Timings.prototype.printReport = function (to, subject) {
...
+
+/**
+ * Creates a new {SyntaxHighlighter} instance.
+ *
+ * @memberof SyntaxHighlighterBase
+ * @param {string} name - The {string} name of the syntax highlighter to create.
+ * @param {Function} block - A {Function} with the {SyntaxHighlighter} instance passed as `this`
+ * @returns a {SyntaxHighlighter} instance for the specified name.
+ */
+SyntaxHighlighterBase.create = function (name, block) {
```
Mogztter: We already define a `create` on `SyntaxHighlighter` so it might be confusing. The main difference is that `SyntaxHighlighter.create` will create an instance of an existing `SyntaxHighlighter`, where `SyntaxHighlighterBase.create` will create a *new* instance.

packages/core/src/asciidoctor-core-api.js
Mogztter: Maybe `isHighlighted` ?

Mogztter:
@mojavelinux I wonder if we should mark this API as "experimental". The closest I've found in JSDoc is `@private`: http://usejsdoc.org/tags-private.html

But it feels like double standards.
We are using SemVer but if the API is experimental then we can break this API anytime! But at the same time it's hard to create a good API when nobody is using it...

mojavelinux:
I think it's totally fine to mark an API as experimental and I see that as still consistent with SemVer. To me, SemVer is about not breaking APIs that were understood to be public / stable. If you tell the user up front that it's experimental, there was no such expectation.

packages/core/src/asciidoctor-core-api.js
mojavelinux: I think exposing both exposes too much. Really, this is really designed to be an internal API anyway. The integrator should only have to register a syntax highlighter. It's up to the processor to create it and make use of it.

The only API I think we need from this side is `SyntaxHighlighter.for`. That resolves what was previously registered, which can be useful for extending and reregistering.

packages/core/src/asciidoctor-core-api.js
mojavelinux: The intended meaning of the name here is "shouldHighlight" or "handlesHighlighting". The premise is whether this adapter handles highlighting, or whether it just prepares highlighting to be performed by the viewer.

packages/core/spec/share/asciidoctor-spec.js
mojavelinux: I don't really like the idea that create is even being used in this context. The create method on the base class is definitely for use by the processor only.

Mogztter: Ok

packages/core/src/asciidoctor-core-api.js
Mogztter: I think we also need an API to register a syntax highlighter. It can be cumbersome to create (and instantiate) a new class in Opal that's why I introduces this `create` function.

But we could mimic the extension registry and do something like:


```js
asciidoctor.SyntaxHighlighter.register('unavailable', function () {
  this.format(function (node, language) {
    return `<pre class="highlight"><code class="language-${language}" data-lang="${language}">${node.getContent()}</code></pre>`
  })
  this.highlighted(() => false)
})
```

Or even:

```js
class UnavailableSyntaxHighlighter {
  format(node, language) {
    return `<pre class="highlight"><code class="language-${language}" data-lang="${language}">${node.getContent()}</code></pre>`
  }

  highlighted() {
    return false
  }
}

asciidoctor.SyntaxHighlighter.register('unavailable', UnavailableSyntaxHighlighter)
```

mojavelinux: I definitely support that idea. I'd say map it in the most consistent way as you see for Asciidcotor.js users.

PR #702, opened 2019-03-26
Improve the Logger API to log messages

Mogztter:


packages/core/src/asciidoctor-core-api.js
```diff
@@ -2041,6 +2044,24 @@ MemoryLogger.prototype.getMessages = function () {
...
 
+var Logging = Opal.const_get_qualified(Opal.Asciidoctor, 'Logging', true)
+
+Opal.Asciidoctor.Logging = Logging
+
+Logging.getLogger = function () {
+  return LoggerManager.$logger()
+}
+
+Logging.messageWithContext = function (text, context) {
```
mojavelinux: Since this isn't a DSL in JavaScript, perhaps we should go with a more method-like name such as `createMessageWithContext` or even just `createLogMessage` (implying that it is adding context).

Mogztter: :+1: 
Indeed, I've settled on `createLogMessage`

mojavelinux: I like it.

Issue #709, opened 2019-04-01
SyntaxHighlighter access to document?


oncletom:
Hello,

I'm trying to convert an existing syntax highlighter extension to the new `SyntaxHighlighter` introduced in `asciidoctor@2` (because the legacy syntax does not work anymore, I have to explicitly register the syntax highlighter).

How can I access, once, the document to grab one of its attributes? The constructor does not have access to it and it's not clear to me what methods I have access to (aside of the ones mentioned in the JSDoc header).

On a side note, I've tried to include docinfo stuff but having this piece of code did not suffice:

```js
module.exports = class PrismHighlighter {
  format () { /* ... */ }

  hasDocinfo () {
    return true;
  }

  docinfo (doc) {
    if (doc.backend !== 'html5') {
      return '';
    }

    // ...
  }
}
```

mojavelinux:
The document is passed as an option to the constructor of the syntax highlighter adapter. See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/document.rb#L1225-L1231

I'm not aware of how this is mapped in Asciidoctor.js.

Mogztter:
> I'm not aware of how this is mapped in Asciidoctor.js.

I'm still trying to figure out how to wire the special `constructor` method. The issue is that Opal will instantiate the class but at the same time we need to call the `constructor` method with the existing instance (created by Opal)...

You can try the following but with no guarantee as it's very hacky at the moment so I won't be surprise if it's not working as expected:

```js
module.exports = class PrismHighlighter {
  constructor(name, backend, opts) => {
    this.backend = backend
  }

  docinfo () {
    if (this.backend !== 'html5') {
      return '';
    }
    // ...
  }
}
```

If you are interessed here's the code that does the wiring:
https://github.com/asciidoctor/asciidoctor.js/blob/8aa8a4c318073cabb44d2daf2bde8f02e953721b/packages/core/src/asciidoctor-core-api.js#L73-L80

You can also try to use the native Ruby API or declare the PrismHighlighter as follow:

```js
asciidoctor.SyntaxHighlighter.register('prism', {
  initialize: (name, backend, opts) => {
    this.backend = backend
    this.super()
  },
  docinfo () {
    if (this.backend !== 'html5') {
      return ''
    }
    // ...
  }
})
```

oncletom:
Thank you for the pointers :-)

[...]

PR #710, opened 2019-04-01
🚑 Fix the SyntaxHighlighter API

Mogztter:
Not perfect but it fixes a few issues :grin: 

Ref: #709
//cc @oncletom

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -72,7 +72,9 @@ function initializeClass (superClass, className, functions, defaultFunctions, ar
   var initialize
   if (typeof constructorFunction === 'function') {
     initialize = function () {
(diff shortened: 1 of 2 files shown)


packages/core/src/asciidoctor-core-api.js
```diff
@@ -72,7 +72,9 @@ function initializeClass (superClass, className, functions, defaultFunctions, ar
   var initialize
   if (typeof constructorFunction === 'function') {
     initialize = function () {
-      var result = new constructorFunction(arguments) // eslint-disable-line
+      var args = Array.prototype.slice.call(arguments)
```
Mogztter: there's probably a better solution

Mogztter: Exactly thanks :+1:

PR #716, opened 2019-04-11
resolves #715 map syntax highlighter methods arguments

### LINKED ISSUE #715: SyntaxHighlighter provides an object when the source language is not specified
With `@asciidoctor/core@2.0.0-rc.3`, a custom `SyntaxHighlighter` and this listing:

```adoc
[source]
.options/zones.txt
----
Europe/London
America/New_York
----
```

In this case, the `format (node, lang)`'s `lang` parameter is equal to `constructor { '$$id': 4, apply: [Function], call: [Function] }`.

I expected the highlighter not to be called — or a blank string at best.

Mogztter:
resolves #715

### DIFF
--- a/packages/core/spec/share/asciidoctor-spec.js
+++ b/packages/core/spec/share/asciidoctor-spec.js
@@ -1273,7 +1273,7 @@ April,32
       it('should register a new syntax highlighter', () => {
         asciidoctor.SyntaxHighlighter.register('unavailable', {
           initialize: function (name, backend, opts) {
-            this.backend = backend
+            this.backend = opts.document.getAttribute('backend')
             this.super()
           },
           format: (node, language) => {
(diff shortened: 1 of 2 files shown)


packages/core/spec/share/asciidoctor-spec.js
```diff
@@ -1273,7 +1273,7 @@ April,32
       it('should register a new syntax highlighter', () => {
         asciidoctor.SyntaxHighlighter.register('unavailable', {
           initialize: function (name, backend, opts) {
-            this.backend = backend
+            this.backend = opts.document.getAttribute('backend')
```
Mogztter: `opts` (third argument) is now a JS object

packages/core/spec/share/asciidoctor-spec.js
```diff
@@ -1321,14 +1321,78 @@ puts 'Hello, World!'
...
+      it('should register a class as a new syntax highlighter with a docinfo', () => {
+        class PrismClientHighlighter {
           constructor (name, backend, opts) {
-            this.backend = backend
+            this.backend = opts.document.getAttribute('backend')
           }
 
           format (node, lang, opts) {
-            return `<pre${lang ? ` lang="${lang}"` : ''} class="prism"><code>${node.getContent()}</code></pre>`
+            if (lang) {
```
Mogztter: `lang` is a String (can be `undefined`).
Previously `lang` could be `Opal.nil`.

PR #719, opened 2019-04-14
🚑 resolves #717 do not use an advanced compilation level

### LINKED ISSUE #717: The 2.x.x minified version of AsciiDoctor JS not working
The minified version of Asciidoctor.js does not expose the Asciidoctor object. This problem does not exist in the non minified version.

image

Location of release zip used : https://github.com/asciidoctor/asciidoctor.js/releases/download/v2.0.0/asciidoctor.js.dist.zip

If we create a test html file called test.html in the "browser" folder of the unzipped distribution, this issue can be demonstrated with the following code.

**NOT WORKING (minified does not have the Asciidoctor object exposed)**

```
<html>
<head>
<script src="asciidoctor.min.js"></script>
<script>
function test() {
var asciidoctor = Asciidoctor();
var html        = asciidoctor.convert('Hello, _Asciidoctor_');
var body        = document.body;
body.innerHTML = html;
}
</script>
</head>
<body onload="test()">
</body>
</html>
```

**WORKING (unminified working fine)**

```
<html>
<head>
<script src="asciidoctor.js"></script>
<script>
function test() {
var asciidoctor = Asciidoctor();
var html        = asciidoctor.convert('Hello, _Asciidoctor_');
var body        = document.body;
body.innerHTML = html;
}
</script>
</head>
<body onload="test()">
</body>
</html>
```

Mogztter:
Code produced with advanced level is invalid because Opal is using `Object.defineProperty`.

### DIFF
--- a/packages/core/tasks/module/uglify.js
+++ b/packages/core/tasks/module/uglify.js
@@ -14,5 +14,5 @@ module.exports.uglify = async () => {
   const browserSource = 'build/asciidoctor-browser.js'
   const browserDestination = 'build/asciidoctor-browser.min.js'
   log.transform('minify', browserSource, browserDestination)
-  await new Uglify(['--jscomp_off=undefinedVars', '--compilation_level=ADVANCED', '--warning_level=QUIET']).minify(browserSource, browserDestination)
+  await new Uglify(['--jscomp_off=undefinedVars', '--warning_level=QUIET']).minify(browserSource, browserDestination)
 }


mojavelinux:
It might be interesting to see if it's possible to supply exceptions. That why, it surgically preserves certain required syntax such as defineProperty.

Mogztter:
>

I didn't find anything in the documentation that indicate that it's possible to supply exceptions but...

1. it's possible to annotate the code to give hints: https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler
2. we should use quoted string to refer to object properties: https://developers.google.com/closure/compiler/docs/limitations

```diff
-$defineProperty(klass.$$prototype, '$$class', klass);
+$defineProperty(klass['$$prototype'], '$$class', klass);
```

https://github.com/opal/opal/blob/a1f36da26a2da31770da3f0e24ea1ddd4fd4e7a1/opal/corelib/runtime.js#L439

Though I don't think it would be enough so I will proceed with this fix and try to patch Opal bit by bit.

mojavelinux:
Nice research. Good strategy. :+1:

PR #720, opened 2019-04-14
Add the project name, version and license in the browser file

Mogztter:
As suggested by @ainslec.
Google Clojure compiler preserves the `@license` tag, so the unminifed file will contain:

```
/**
 * @license Asciidoctor.js 2.0.0 | MIT - https://github.com/asciidoctor/asciidoctor.js/blob/master/LICENSE
 */
```

And the minified file will contain:

```
/*
 Asciidoctor.js 2.0.0 | MIT - https://github.com/asciidoctor/asciidoctor.js/blob/master/LICENSE
*/
```

I think it's important to add the project name and version. Wdyt ?

### DIFF
--- a/packages/core/tasks/module/builder.js
+++ b/packages/core/tasks/module/builder.js
@@ -96,7 +96,18 @@ const generateFlavors = (asciidoctorCoreTarget, environments) => {
     }
     templateModel['//{{asciidoctorCode}}'] = asciidoctorData
     const content = parseTemplateFile(templateFile, templateModel)
-    fs.writeFileSync(target, content, 'utf8')
+    if (environment === 'browser') {
+      const header = `/**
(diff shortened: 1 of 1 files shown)


mojavelinux:
I agree. Though, the license doesn't have to be to the file. I can just be the main project URL (where the license can be found).

Mogztter:
We can also use `@preserve`: https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#license-preserve-description but I think it's OK to have a single line.

> I can just be the main project URL (where the license can be found).

What do you think about:

```js
/**
 * @license MIT | Asciidoctor.js 2.0.0 - https://github.com/asciidoctor/asciidoctor.js
 */
```
```js
/*
 MIT | Asciidoctor.js 2.0.0 - https://github.com/asciidoctor/asciidoctor.js
 */
```

mojavelinux:
Excellent! Except I would choose either a hyphen or a pipe.

PR #737, opened 2019-06-21
resolves #736 map AbstractBlock accessors

### LINKED ISSUE #736: Create a block macro not in a paragraph
Hi, i'm trying to make a block macro that represent a block with an URL and a HTTP verb, like so: 

```asciidoc
url::http://localhost:8000/api[verb=PATCH]
```

Here is what I have at the moment:

```js
registry.inlineMacro("url", function() {
    this.process(
      (_parent, target, attrs) =>
        `<div class="openblock url-block">
          ${attrs.verb ? `<div class="title">POST</div>` : ""}
          <div class="content">
            <div class="listingblock">
              <div class="content">
                <pre class="highlight">
                  <code>${target}</code>
                </pre>
              </div>
            </div>
          </div>
        </div>`
    )
  })
```

I faked it with an `inlineMacro` as I don't know how to create this HTML properly.

It produces almost what I want but with a caveat: this HTML is inside a `<div class="paragraph">` and I would like to have it be its own block or at least, be inside a section.

I tried changing the `paragraph` context to `section` but this does not seem to work and I don't know where to look in the documentation.
Is it even possible? 

Thanks in advance.

Mogztter:
- #setLevel
- #setStyle
- #setSourceLocation + #getSourceLocation

resolves #736

### DIFF
--- a/packages/core/spec/share/asciidoctor-spec.js
+++ b/packages/core/spec/share/asciidoctor-spec.js
@@ -462,6 +462,21 @@ intro
         expect(block.getLevel()).to.equal(2)
       })
 
+      it('should assign the title on a block', function () {
+        const doc = asciidoctor.load(`= Title
+
+== Section A
+
+== Section B
+`)
+        const firstBlock = doc.getBlocks()[0]
+        const secondBlock = doc.getBlocks()[1]
+        expect(firstBlock.getTitle()).to.equal('Section A')
+        expect(secondBlock.getTitle()).to.equal('Section B')
+        secondBlock.setTitle('Section BB')
+        expect(secondBlock.getTitle()).to.equal('Section BB')
+      })
+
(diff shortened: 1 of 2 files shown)


mojavelinux:
I'd get rid of setLevel (or scope it to Section only). It's extremely poorly defined and we need to rethink its purpose.

It seems we're still missing setTitle. Is that true? (It looks like it's only defined on Document).

Also, setSourceLocation should only be done internally, so I don't think we need to map that one. Unless you can think of a case where it's needed.

Mogztter:
- Scoped `setLevel` to Section
- Added `AbstractBlock#setTitle`
- Removed `AbstractBlock#setSourceLocation`

Mogztter:
@mojavelinux Would you like to review it one last time before I merge?

Issue #752, opened 2019-09-23
Adding typescript declaration file


MrRefactoring:
I use this library in my typescript project and would like to see asciidoctor.d.ts in this project so that I can use this painlessly with typescript

Mogztter:
Hello @MrRefactoring 
I'm not really familiar with TypeScript definition file, if you can provide any guidance that would be great.

Thanks

PR #753, opened 2019-09-23
Typescript declaration file added

MrRefactoring:
Added typescript declaration file (#752)

### DIFF
--- a/packages/asciidoctor/index.d.ts
+++ b/packages/asciidoctor/index.d.ts
@@ -1,0 +1,19 @@
+export interface IRuntime {
+  ioModule: string | 'node';
+  platform: string | 'node';
+  engine: string | 'v8';
+  framework: string;
+}
+
+export class Asciidoctor {
+  getVersion(): string;
+  getCoreVersion(): string;
+  getRuntime(): IRuntime;
+  convert(input: string | Buffer, options?: any): string | Document;
+  convertFile(filename: string, options?: any): string | Document;
+  load(input: string | Buffer, options?: any): Document;
+  loadFile(filename: string, options?: any): Document;
+  toString(): string;
+}
+
+export default function asciidoctor(): Asciidoctor;
--- a/packages/asciidoctor/package.json
+++ b/packages/asciidoctor/package.json
@@ -3,6 +3,7 @@
   "version": "2.0.3",
   "description": "A JavaScript AsciiDoc processor, cross-compiled from the Ruby-based AsciiDoc implementation, Asciidoctor, using Opal",
   "main": "index.js",
+  "types": "index.d.ts",
   "bin": {
     "asciidoctor": "bin/asciidoctor",
     "asciidoctorjs": "bin/asciidoctor"


Mogztter:
Thanks @MrRefactoring 
The `asciidoctor` package is meta-package so I think `index.d.ts` should be moved to the `packages/core` directory because the core module contains the Asciidoctor API.

packages/asciidoctor/index.d.ts
```diff
@@ -0,0 +1,19 @@
+export interface IRuntime {
```
Mogztter: Is it possible to add any kind of tests to make sure that the syntax is valid and that the definition matches the API?

packages/asciidoctor/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+  platform: string | 'node';
+  engine: string | 'v8';
+  framework: string;
+}
+
+export class Asciidoctor {
+  getVersion(): string;
+  getCoreVersion(): string;
+  getRuntime(): IRuntime;
+  convert(input: string | Buffer, options?: any): string | Document;
```
Mogztter: I think we need to define `Document` right?

MrRefactoring: As I understand it, this is a standard document from the dom tree.  Typescript has its basic support

packages/asciidoctor/index.d.ts
MrRefactoring: Yes, I think I can write typescript tests for this functionality

packages/asciidoctor/index.d.ts
MrRefactoring: document

mojavelinux: Document is an Asciidoctor Document object. It holds the parsed blocks. That object is returned if to_file is a writable File object. Instead of returning undefined, we return the parsed document for further inspection.

MrRefactoring: Got it, thanks

Mogztter: Asciidoctor does not necessarily convert an AsciiDoc content to an HTML page. 
In this case the `Document` object is an AST: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/document.rb that represents the structure of the AsciiDoc.

Here's the link to the API documentation: 

http://asciidoctor.github.io/asciidoctor.js/master/#asciidoctorconvert
http://asciidoctor.github.io/asciidoctor.js/master/#document

Mogztter: Dan was faster :sweat_smile:

MrRefactoring: Thanks for the links to documentation, this is very useful for me!

Mogztter: You're welcome!
I realize that mapping the entire API at once might be a tedious task. So if you don't want to spend too much time on it that's OK :wink: 
I believe we can iterate on the definition file to keep improving the Typescript support, right?

MrRefactoring: Right, we can add declarations iteratively

Mogztter:
Thanks for the initial work @MrRefactoring :tada: 
I've added some changes in https://github.com/asciidoctor/asciidoctor.js/pull/756

Mogztter:
Superseded in #756

PR #763, opened 2019-11-10
⬆️💥 Upgrade to opal 1.0.0 (strict mode)

Mogztter:


mojavelinux:
Cool!

Mogztter:
I'm not 100% sure that the code produced by Opal 0.11.99 is compatible with the code produced by Opal 1.0.0 :thinking: 
Also Node 8 will be EOL soon (January 2020). 

@mojavelinux So we will probably need to do a major/breaking release to drop Node 8 and update to Opal 1.0+. I don't think I can reasonably do these changes in a minor/feature version... ?

mojavelinux:
Sounds correct to me.

Mogztter:
The following is working in Chrome and Firefox (with a few changes):

```html
<script type="module">
  import { convert } from './build/asciidoctor-browser.js';
  console.log(convert('Hello *world*'));
  // prints:
  // <div class="paragraph">
  // <p>Hello <strong>world</strong></p>
  // </div>
</script>
```

oncletom:
Ouh I love it!

Mogztter:
I think we should always enable strict mode right? Unless it has performance issues... I will run the complete test suites and the benchmark on the ESM file.

Mogztter:
Asciidoctor.js does not compile with the latest version of Opal (master).
The following commit https://github.com/opal/opal/commit/d46792d2160e4f524c3add711f6424dd99187d1c throws a `Invalid multibyte sequence` on the following line:

https://github.com/asciidoctor/asciidoctor/blob/8aadc1412267bf161f911e6c4800d08fa3d9263e/lib/asciidoctor/rx.rb#L19

```
Mary_Sue Brontë
```

To workaround this issue, I've replaced `ë` with `e` but then the compilation takes more than 8 minutes (I've actually killed the process so it can take 9 minutes or 20 I don't know).

If I revert the commit https://github.com/opal/opal/commit/d46792d2160e4f524c3add711f6424dd99187d1c then I can compile Asciidoctor.js.

Mogztter:
@mojavelinux @oncletom Is it a bad thing if Asciidoctor.js 3+ is only compatible with ECMAScript 6?

I really want to use modern JavaScript (including [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export)but not limited to) and I also want to avoid using a web bundler and/or Babel and/or polyfill.

mojavelinux:
I guess it really comes down to IE 11 support. With < 2% global usage, and plenty of support in Edge versions, I'm in favor of dropping. https://caniuse.com/#feat=es6

Mogztter:
> I guess it really comes down to IE 11 support. With < 2% global usage, and plenty of support in Edge versions, I'm in favor of dropping. https://caniuse.com/#feat=es6

OK! And the next question is, should I only publish Asciidoctor.js as an ESM or should I keep the UMD version?
It feels like AMD and UMD are artefacts of the past and I don't really want to publish yet another flavor of Asciidoctor.js. I believe that the JavaScript ecosystem now works flawlessly with ESM modules no?

mojavelinux:
I don't actually know that answer to that question. In other words, I don't know which module flavor I rely on now. Can you show an example of each so I know what you mean?

Mogztter:
> I don't actually know that answer to that question. In other words, I don't know which module flavor I rely on now. Can you show an example of each so I know what you mean?

Sure.
As far as I understand the JavaScript ecosystem is moving toward ESM.
In Node.js 14, you can do:

```js
import Asciidoctor from 'asciidoctor'
```

GraalVM also supports ESM. When using the GraalVM Polyglot API, the requirement is to use a file with a `mjs` extension when using the load directive:

```java
  public static void main(String[] args) {
    Context context = Context.newBuilder("js")
      .allowIO(true)
      .allowAllAccess(true)
      .allowPolyglotAccess(PolyglotAccess.ALL)
      .build();
    context.getPolyglotBindings().putMember("IncludeResolver", new IncludeResolver());
    context.eval("js", "load('./main.mjs')");
  }
```

Then in `main.mjs`, you can import Asciidoctor.js as an ESM:

```js
import Asciidoctor from 'asciidoctor'
```

Finally, in the browser, you can use ESM:

```js
<script type="module">
  import Asciidoctor from 'asciidoctor'
</script>
```

For Node.js, I think we will need to publish both a CJS and an ESM module: https://nodejs.org/api/esm.html#esm_conditional_exports
Otherwise, Asciidoctor.js will only be compatible with Node.js 14 or greater.

For browser and GraalVM, I think we only need an ESM. In other words, the following won't work anymore:

**global**
```html
<script src="./asciidoctor.js"></script>
<script>
  console.log(Asciidoctor) // undefined
</script>
```

**AMD**
```js
requirejs(['asciidoctor'], function (asciidoctor) {
  // nope!
})
```

Speaking of flavors, we are still going to publish a dedicated version for Node.js, the Browser and GraalVM to accommodate each runtime (mostly regarding I/O). 

In summary:
* Node.js: CJS + ESM
* Browser : ESM
* GraalVM : ESM

mojavelinux:
Thanks for this info!

Yep, requiring Node 14 right now seems overly aggressive as Node 12 is still active (https://nodejs.org/en/about/releases/).

I've been looking forward to this module feature for a long time. It always confuses me when I see `import Thing from 'thing'` in blog posts only to find Node.js can't make sense of it (until finally Node 14 can).

I find it a bit odd that the file extension has to change to .mjs in GraalVM. That seems like it's going to make migration difficult.

Mogztter:
> I find it a bit odd that the file extension has to change to .mjs in GraalVM.

Indeed, you can either use `.mjs` or use the inofficial `application/javascript+module` mime type.

https://github.com/graalvm/graaljs/blob/master/docs/user/JavaScriptCompatibility.md#javascript-modules

Node.js is also using this "hint" to enable ESM or not but you also can enable it using other "hints", see: https://nodejs.org/api/esm.html#esm_enabling

> That seems like it's going to make migration difficult.

The entry point must have `.mjs` to enable ESM but after that you don't need to rename all your files to `.mjs`.

mojavelinux:
> The entry point must have .mjs to enable ESM but after that you don't need to rename all your files to .mjs.

Ah, that makes sense. Thanks for clarifying.

oncletom:
Yeah, I would favour ESM as the main target.

If you declare [`"type": "module"` in the package.json](https://nodejs.org/api/esm.html#esm_package_json_type_field), you're covered!

Mogztter:
> I would also target Deno as a runtime, because it can be similar to Node, without the configuration hassle.

👍 
I will configure the tests suite to run on Deno as well.

Mogztter:
I will do it as part of https://github.com/asciidoctor/asciidoctor.js/issues/801 because I don't want to introduce too much changes in here.

The scope of this issue/pull request is to upgrade to Opal 1+ and use strict mode.

PR #769, opened 2019-11-30
resolves #768 initialize backend traits

### LINKED ISSUE #768: Retrieve backend traits from the converter instance when using a class
```js
class TEIConverter {
  constructor () {
    this.basebackend = 'xml'
    this.outfilesuffix = '.xml'
    this.filetype = 'xml'
    this.htmlsyntax = 'xml'
  }
  /* ... */
}

Asciidoctor.ConverterFactory.register(new TEIConverter(), ['tei'])
```

The above should set the backend traits on the converter.

//cc @oncletom

Mogztter:
The goal is to bridge the Asciidoctor (Ruby) module named `BackendTraits` in a pure JavaScript class or instance.

The main benefit is that the user doesn't have to manipulate the `Opal` object in order to create a class but it has a few downsides.
Most notably, when using the `constructor` because the class is not yet bridged to the `Converter::Base` (Ruby) class.
So you can't use `super` or call the method `init_backend_traits`.

resolves #768

### DIFF
--- a/packages/core/spec/node/asciidoctor.spec.js
+++ b/packages/core/spec/node/asciidoctor.spec.js
@@ -1723,6 +1723,82 @@ header_attribute::foo[bar]`
       const result = asciidoctor.convert('content', options)
       expect(result).to.contain('<dummy>content</dummy>')
     })
+    it('should retrieve backend traits from a converter class', () => {
+      class TEIConverter {
+        constructor (backend, _) {
+          this.backend = backend
+          this.backendTraits = {
+            basebackend: 'xml',
+            outfilesuffix: '.xml',
+            filetype: 'xml',
+            htmlsyntax: 'xml'
+          }
+          this.transforms = {
+            embedded: (node) => {
+              return `<tei>${node.getContent()}</tei>`
+            }
+          }
+        }
+
+        convert (node, transform) {
+          const name = transform || node.node_name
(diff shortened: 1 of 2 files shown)


packages/core/spec/node/asciidoctor.spec.js
```diff
@@ -1723,6 +1723,82 @@ header_attribute::foo[bar]`
...
+          }
+          return this.transforms[name](node)
+        }
+
+        convertParagraph (node) {
+          return node.getContent()
+        }
+      }
+
+      asciidoctor.ConverterFactory.register(new TEIConverter(), ['tei'])
```
Mogztter: Here we instantiate the JavaScript class. In this case, the constructor does not have argument.

packages/core/spec/node/asciidoctor.spec.js
```diff
@@ -1723,6 +1723,82 @@ header_attribute::foo[bar]`
...
+          }
+          return this.transforms[name](node)
+        }
+
+        convertParagraph (node) {
+          return node.getContent()
+        }
+      }
+
+      asciidoctor.ConverterFactory.register(TEIConverter, ['tei'])
```
Mogztter: Here we use a class. The processor will call the constructor with two arguments, the backend name and a list of options.

PR #770, opened 2019-12-01
resolves #5 add JavaScript and Pug templates (optional dependency)

### LINKED ISSUE #5: Support custom templates
Not exactly sure how we'll get around the browser sandbox, but I'm sure something can be figured out.

Mogztter:
The implementation is only available in a Node environment.
In a browser environment, I think it's better to implement a custom converter. Reading template files using a synchronous `XMLHttpRequest` is inefficient and I think it's better to encourage users to create their own converter where they can precompile templates and/or use a single JavaScript file.

resolves #5


packages/core/lib/asciidoctor/js/asciidoctor_ext/node/template.rb
```diff
@@ -0,0 +1,194 @@
+# frozen_string_literal: true
```
Mogztter: @mojavelinux The code is very close to the [original implementation in Ruby](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/template.rb). To avoid code duplication, we could extract some parts and add conditional blocks or we could create a `TiltTemplateConverter` with the Tilt related bits.

packages/core/lib/asciidoctor/js/asciidoctor_ext/node/template.rb
```diff
@@ -0,0 +1,194 @@
...
+        when :js
+          template = `{ render: require(file), '$file': function() { return file } }`
+        else
+          next
+        end
+      end
+      result[name] = template
+    end
+    if helpers || ::File.file?(helpers = %(#{template_dir}/helpers.js))
+      `require(helpers)`
```
Mogztter: I think this feature is useless because the helper functions won't be available in the templates.
We could probably add an "helpers" property to the context?

```rb
@helpers = `require(helpers)`
# ...
`template.render({node: node, opts: opts, helpers: this.helpers})`
```

Mogztter:
Apparently Pug is not working with GraalVM, I need to debug it.

Mogztter:
> Apparently Pug is not working with GraalVM, I need to debug it.

It's working but it's really slow :disappointed:

Mogztter:
I will postpone this feature to 2.2.0.

Mogztter:
The following template engines are supported out-of-the-box:

- Pug
- Nunjucks
- EJS
- Handlebars

Please note that the dependencies are optional, so you will need to explicitly install the template engine you want to use. For instance, if you want to use Ninjucks, you will need to open a terminal and type:

```
$ npm i asciidoctor ninjucks
```

I've also added an example on how to register a template engine. In this case, I'm using [doT](http://olado.github.io/doT/):

```js
const options = { safe: 'safe', backend: ':html5', template_dir: '/pat/to/templates/dot', template_engine: 'dot' }
const fs = require('fs')

class DotTemplateEngine {
  constructor () {
    this.doT = require('dot')
  }

  compile (file, _) {
    const templateFn = this.doT.template(fs.readFileSync(file, 'utf8'))
    return {
      render: templateFn
    }
  }
}

asciidoctor.TemplateEngineRegistry.register(new DotTemplateEngine(), ['dot'])
const result = asciidoctor.convert('content', options)
expect(result).to.contain('<p class="paragraph-dot">content</p>')
```

**/pat/to/templates/dot/paragraph.dot**
```dot
<p class="paragraph-dot">{{= it.node.getContent() }}</p>
```

Mogztter:
I've renamed `TemplateEngineRegistry` to `TemplateEngine` to be consistent with the others API like `SyntaxHighlighter` (where you can `register` a new `SyntaxHighlighter`).

I've also added documentation and the TypeScript definition. I will play with it a few days before merging!

PR #775, opened 2019-12-28
resolves #764 add TypeScript defs for extensions and converters

### LINKED ISSUE #764: Add TypeScript definitions for extensions and converters
It would be great to also have TypeScript definitions for the Extension and Converter APIs. It would especially be helpful to have a schema for the different Node types a converter can handle. I just wrote my first modification to the Html5Converter and had to look into the `asciidoctor` ruby source code to get an understanding what properties and functions are available. Having the type info at hand would make this a lot easier.

Mogztter:
resolves #764

### DIFF
--- a/packages/core/types/README.adoc
+++ b/packages/core/types/README.adoc
@@ -1,0 +1,21 @@
+= TypeScript Declaration File
+
+== How to Update the TypeScript Declaration File
+
+We are using https://github.com/englercj/tsd-jsdoc[tsd-jsdoc] to generate a TypeScript Declaration File (.d.ts) from JSDoc annotated source JavaScript files.
+
+1. Install https://www.npmjs.com/package/tsd-jsdoc[tsd-jsdoc] and https://www.npmjs.com/package/jsdoc[jsdoc]
+2. Run `node_modules/.bin/jsdoc -t node_modules/tsd-jsdoc/dist -r src/asciidoctor-*.js -d build`
+
+The above command will generate a TypeScript Declaration File at `build/types.d.ts`.
+Please note that this file must manually curated because the JSDoc is incomplete and, as a result, the generated file will be invalid.
+
+== TODO
+
+- Refine types
+** Some functions takes a string as argument but we could replace it by the list of values.
+For instance when we create a `List`, the context is one of: `ulist`, `olist`, `colist` or `dlist`.
+- Add more tests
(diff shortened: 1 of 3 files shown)


Mogztter:
It's a good opportunity to map missing methods/modules and add the missing documentation :grinning:

Mogztter:
I must say that it's very pleasant to discover the API with auto-completion when writing in TypeScript :nerd_face: 

But at the same time, it's tedious to declare all the types in the declaration file... hopefully once the heavy lifting will be done it won't be that hard to maintain.

sondr3:
If you manage to keep the JSDocs up to date you can generate the TypeScript types automatically, see https://voxpelli.com/2019/10/use-type-script-3-7-to-generate/ for example. I think this would be a better approach than manually having to make sure the types document is up to date.

Mogztter:
Thanks for the link.
I did try to automatically generate the TypeScript definition file with TypeScript but I get the following exception:

```
TypeError: Cannot read property 'valueDeclaration' of undefined
```

Even with the latest development version (3.8.0-dev).

Not sure if it's related or not to https://github.com/microsoft/TypeScript/issues/35801because we are not using webpack.

For now, I'm using `tsd-jsdoc` but it has a few limitations so I need to manually edit the generated file.

If you found a better alternative please feel free to add a comment 👍

sondr3:
I gave it a go myself this weekend and hit the same issue you had and couldn't figure it out myself either. Very frustrating when it looks like it is a regression between releases of the TypeScript compiler.

Mogztter:
> I gave it a go myself this weekend and hit the same issue you had and couldn't figure it out myself either. Very frustrating when it looks like it is a regression between releases of the TypeScript compiler. 

Indeed.
For now I'm using https://github.com/englercj/tsd-jsdoc to generate the TypeScript definition. It's not perfect but it does a decent job.

Mogztter:
I think it's good to go.
I'm pretty sure there are a few mistakes but overall it's working great.
I will document what we should do next:

- refine types
    * some functions takes a `string` as argument but we could replace it by the list of values. For instance when we create a `List`, the context is one of: `ulist`, `olist`, `colist` or `dlist`.
- add more tests
   * we should call at least each function once in the `tests.ts` file to make sure that the definition is valid
- automate updates
   * currently the type definition is carefully crafted by hand but we need to automate this process a bit (for instance when we add, remove or edit an existing API)

PR #780, opened 2019-12-30
Fix the Processor#resolvesAttributes function

Mogztter:
- `resolvesAttributes` is now deprecated (in favor of `resolveAttributes`)
- `resolveAttributes` can take multiple arguments
- `resolveAttributes` argument can be
  * a String
  * a list of String (either as an Array or as a list of arguments)
  * a boolean
  * undefined
  * a JSON object

I think we should document the format in the Asciidoctor core API.

### DIFF
--- a/packages/core/spec/node/asciidoctor.spec.js
+++ b/packages/core/spec/node/asciidoctor.spec.js
@@ -1437,30 +1437,84 @@ header_attribute::foo[bar]`
           expect(doc.getAttribute('foo')).to.equal('bar')
         })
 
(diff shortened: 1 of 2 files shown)


packages/core/spec/node/asciidoctor.spec.js
```diff
@@ -1737,6 +1817,7 @@ header_attribute::foo[bar]`
           return node.getContent()
         }
       }
+
```
Mogztter:
```suggestion
```

PR #817, opened 2020-01-17
resolves #816 chrome:// is a root path (in a browser environment)

### LINKED ISSUE #816: Add "chrome://" as a root path in the browser
Initial discussion: https://github.com/asciidoctor/asciidoctor.js/issues/62

To do that, we should override the `root?` method:
https://github.com/asciidoctor/asciidoctor/blob/35c1f9a8f7c1cee310d6c1d1ad46ec5ae4702d54/lib/asciidoctor/path_resolver.rb#L163-L169

Mogztter:
resolves #816

### DIFF
--- a/packages/core/lib/asciidoctor/js/asciidoctor_ext/browser/path_resolver.rb
+++ b/packages/core/lib/asciidoctor/js/asciidoctor_ext/browser/path_resolver.rb
@@ -1,0 +1,7 @@
+module Asciidoctor
+class PathResolver
+  def root? path
+    (absolute_path? path) || (path.start_with? 'file://', 'http://', 'https://', 'chrome://')
+  end
+end
+end
--- a/packages/core/spec/browser/asciidoctor-spec.js
+++ b/packages/core/spec/browser/asciidoctor-spec.js
@@ -16,7 +16,9 @@
   const expect = chai.expect
   const asciidoctor = Asciidoctor({ runtime: { platform: 'browser' } })
   const parts = window.location.href.split('/') // break the string into an array
-  parts.pop(); parts.pop(); parts.pop()
+  parts.pop()
+  parts.pop()
+  parts.pop()
   const baseDir = parts.join('/')
   const asciidoctorCoreSemVer = semVer(asciidoctor.getCoreVersion())
   const testOptions = {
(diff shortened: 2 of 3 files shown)


packages/core/lib/asciidoctor/js/asciidoctor_ext/browser/path_resolver.rb
```diff
@@ -0,0 +1,7 @@
+module Asciidoctor
+class PathResolver
+  def root? path
+    (absolute_path? path) || (path.start_with? 'file://', 'http://', 'https://', 'chrome://')
```
Mogztter: ping @bodtx!
Is the code correct?

packages/core/spec/browser/asciidoctor-spec.js
```diff
@@ -43,6 +45,28 @@
...
+          standalone: true,
+          attributes
+        }
+        try {
+          asciidoctor.convert('Hello world', options)
+          expect.fail('the resource does not exist and the converter should throw an exception!')
+        } catch (e) {
+          // the resource does not exist but chrome:// is a root path and should not be prepended by "./"
+          // we expect the processor to try to load the resource 'chrome://asciidoctor/extension/css/asciidoctor-plus.css' and fail!
+          expect(e.message).to.include(`Failed to load 'chrome://asciidoctor/extension/css/asciidoctor-plus.css'`)
```
Mogztter: @bodtx I've added comments to explain this test.
Basically I've copied and adapted your code from [chrome/content/myCompose.js](https://framagit.org/bodtx/asciidoctor_tb/blob/f4f0d83d92c5372c6af4521bdaf0d8cce9630ecf/chrome/content/myCompose.js#L71-85) but since the resource `chrome://asciidoctor/extension/css/asciidoctor-plus.css` does not exist, I'm catching the exception and asserting that the resource path was not prepended by `./`. 

Does it make sense?

Mogztter:
I want to include this fix in the next release, please let me know if you think that the code is OK.
Thanks!

packages/core/lib/asciidoctor/js/asciidoctor_ext/browser/path_resolver.rb
bodtx: yes this is correct

packages/core/spec/browser/asciidoctor-spec.js
bodtx: indeed the `./` added to my url was the problem.
if there is just `chrome://` the url should be ok

Mogztter: Thanks for your feedback!

PR #819, opened 2020-01-19
Add a changelog to keep track of changes

Mogztter:
It also makes the release process (ie. create the changelog on the release page) less tedious.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -1,0 +1,183 @@
+= Asciidoctor.js Changelog
+:uri-repo: https://github.com/asciidoctor/asciidoctor.js
+:icons: font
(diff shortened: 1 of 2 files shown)


mojavelinux:
I strongly recommend requiring a CHANGELOG entry in every PR (unless there is good reason not to). That way, the CHANGELOG is always current.

Mogztter:
Is there a way to avoid conflicts? That's the main reason why I would add the changelog after...

mojavelinux:
Place the items strategically. Or you can space them out, then remove the spaces before the release. I've found this to not be too much of an issue since there are usually conflicts somewhere else if there's a conflict in the CHANGELOG. So it's just part of the process.

PR #827, opened 2020-01-26
Include the TypeScript Definition File in the npm packages

### LINKED ISSUE #828: types missing in 2.1.0 release?
Thank you for the 2.1.0 release!

I was happy to see that it now provides a TypeScript Declaration File but the TS compiler is not finding it, and I can't see it in the installed module:

```
npm ls @asciidoctor/core                                                               
exercises@1.0.0
└── @asciidoctor/core@2.1.0

tree node_modules/@asciidoctor/core                                                      
node_modules/@asciidoctor/core
├── LICENSE
├── README.md
├── dist
│   ├── browser
│   │   ├── asciidoctor.js
│   │   └── asciidoctor.min.js
│   ├── css
│   │   └── asciidoctor.css
│   ├── graalvm
│   │   └── asciidoctor.js
│   └── node
│       └── asciidoctor.js
└── package.json

5 directories, 8 files
```

Am I missing something obvious?

Mogztter:
resolves #828

### DIFF
--- a/packages/asciidoctor/package.json
+++ b/packages/asciidoctor/package.json
@@ -3,6 +3,7 @@
   "version": "2.1.0",
   "description": "A JavaScript AsciiDoc processor, cross-compiled from the Ruby-based AsciiDoc implementation, Asciidoctor, using Opal",
   "main": "index.js",
+  "types": "types",
   "bin": {
     "asciidoctor": "bin/asciidoctor",
     "asciidoctorjs": "bin/asciidoctor"
@@ -13,7 +14,8 @@
     "yarn": ">=1.1.0"
   },
   "files": [
-    "bin"
+    "bin",
+    "types/index.d.ts"
   ],
   "repository": {
     "type": "git",
--- a/tasks/publish.js
+++ b/tasks/publish.js
@@ -23,6 +23,8 @@ const publish = async (directory) => {
       log.warn('Dry run! To publish the release, run the command again without DRY_RUN environment variable')
     }
     const projectRootDirectory = path.join(__dirname, '..')
+    // copy packages/core/types/index.d.ts to packages/asciidoctor/types/index.d.ts
+    fs.copyFileSync(`${projectRootDirectory}/packages/core/types/index.d.ts`, `${projectRootDirectory}/packages/asciidoctor/types/index.d.ts`)
     await publish(path.join(projectRootDirectory, 'packages', 'core'))
     await publish(path.join(projectRootDirectory, 'packages', 'asciidoctor'))
   } catch (e) {
--- a/packages/asciidoctor/types/README.adoc
+++ b/packages/asciidoctor/types/README.adoc
@@ -1,0 +1,3 @@
+= README
+
+The file [.path]_index.d.ts_ will be automatically copied from [.path]_packages/core/types/index.d.ts_ when a release is performed.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -12,6 +12,7 @@
   },
   "files": [
     "dist",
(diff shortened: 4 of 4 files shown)


tasks/publish.js
```diff
@@ -22,6 +22,8 @@ const publish = async (directory) => {
     if (process.env.DRY_RUN) {
       log.warn('Dry run! To publish the release, run the command again without DRY_RUN environment variable')
     }
+    // copy packages/core/types/index.d.ts to packages/asciidoctor/types/index.d.ts
+    fs.copyFileSync(`${projectRootDirectory}/packages/core/types/index.d.ts`, `${projectRootDirectory}/packages/asciidoctor/types/index.d.ts`)
```
Mogztter: I've decided to copy the `index.d.tx` just before publishing the `asciidoctor` package. I don't think we need to version this file a second time.
I'm not sure if a file link works so I prefer to just copy the file.

It's mandatory otherwise the TypeScript Declaration File won't be found in the example below:

```js
import Processor from 'asciidoctor'
```

packages/asciidoctor/package.json
```diff
@@ -13,7 +14,8 @@
     "yarn": ">=1.1.0"
   },
   "files": [
-    "bin"
+    "bin",
+    "types/index.d.ts"
```
Mogztter: I believe that only `index.d.ts` is mandatory.

Mogztter:
I'd love to hear your feedback on this issue as I'm not really familiar with TypeScript.

//cc @akosma @oncletom @mojavelinux @sondr3

---- 
For reference, the `asciidoctor` package is a meta-package with a dependency on `@asciidoctor/core` and `@asciidoctor/cli`. The idea is to be able to do:

```
npm i asciidoctor
```
```js
const asciidoctor = require('asciidoctor')()
```

In the above example, `asciidoctor` is basically an alias for `@asciidoctor/core` but the package also includes the CLI, so you can run:

```
./node_modules/.bin/asciidoctor doc.adoc
```
```

cexbrayat:
@Mogztter For it's worth, here's my feedback as I've been using it in a TS project for a while.

I used to import it with `import asciidoctor from '@asciidoctor/core';` and have a cutom type file in `types/asciidoctor__core/index.d.ts`

If I manually reproduce this PR in my installed node_modules, I have to import from `asciidoctor` instead, which is not a big deal. It then mainly works expect for one thing: I use `asciidoctor().SyntaxHighlighter.register('prism', prismExtension);` and it looks like `SyntaxHighlighter` is not exposed in the `Asciidoctor` class so I have to cast it as `any` for now.

Mogztter:
Thanks Cédric!

> If I manually reproduce this PR in my installed node_modules, I have to import from asciidoctor instead, which is not a big deal.

It should also be available in the `@asciidoctor/core`:

https://github.com/asciidoctor/asciidoctor.js/blob/87c48add4a1877dd016841c2598d315394f19076/packages/core/package.json#L7

https://github.com/asciidoctor/asciidoctor.js/blob/master/packages/core/types/index.d.ts

> I use asciidoctor().SyntaxHighlighter.register('prism', prismExtension); and it looks like SyntaxHighlighter is not exposed in the Asciidoctor class so I have to cast it as any for now.

Indeed, I forgot to include `SyntaxHighlighter` definition. I will open another pull request to fix it.
If you find any other omissions/quirks/bugs in the TypeScript definition please feel free to open issues on GitHub :+1:

tasks/publish.js
Mogztter: `projectRootDirectory` is undefined here! 🐛

PR #831, opened 2020-01-26
resolves #829 types for getCaption

### LINKED ISSUE #829: Property 'getCaption' in type 'Section' is not assignable to the same property in base type 'AbstractBlock'
When using the types defined in `index.d.ts` and running `tsc` with TS 3.7.5 and `strict` tsconfig, we run into:

```
> $(npm bin)/tsc -v
Version 3.7.5

> $(npm bin)/tsc
node_modules/asciidoctor/types/index.d.ts:2190:5 - error TS2416: Property 'getCaption' in type 'Section' is not assignable to the same property in base type 'AbstractBlock'.
  Type '() => string | undefined' is not assignable to type '() => string'.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.

2190     getCaption(): string | undefined;
         ~~~~~~~~~~
```

Note that this only happens with a `strict` tsconfig:

```
{
  "compilerOptions": {
    "target": "es2015",
    "module": "commonjs",
    "outDir": "./dist" /,
    "strict": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "typeRoots": ["node_modules/@types", "types"],
    "esModuleInterop": true
  }
}
```

cexbrayat:
This also tightens the tsconfig to use a strict configuration, and fixes the problems that appeared in tests.ts

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -394,7 +394,8 @@ AbstractBlock.prototype.getSourceLocation = function () {
  * @memberof AbstractBlock
  */
 AbstractBlock.prototype.getCaption = function () {
-  return this.$caption()
+  var caption = this.$caption()
+  return caption === Opal.nil ? undefined : caption
 }
 
 /**
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -1989,7 +1989,7 @@ export namespace Asciidoctor {
      * Get the caption for this block.
      * @returns the caption for this block
      */
-    getCaption(): string;
+    getCaption(): string | undefined;
 
     /**
      * Set the caption for this block.
--- a/packages/core/types/tests.ts
+++ b/packages/core/types/tests.ts
@@ -537,7 +537,7 @@ class BlogConverter {
     };
   }
 
-  convert(node: Asciidoctor.AbstractNode, transform: string | null, opts: any) {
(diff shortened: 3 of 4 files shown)


Mogztter:
It's looking good thanks!
I think we should also update the code of `AbstractBlock.prototype.getCaption` in `asciidoctor-core-api.js` to return undefined if the value is `Opal.nil` to be consistent.

cexbrayat:
@Mogztter I pushed another commit, let me know if that's what you expected  :)

packages/core/src/asciidoctor-core-api.js
```diff
@@ -394,7 +394,7 @@ AbstractBlock.prototype.getSourceLocation = function () {
  * @memberof AbstractBlock
  */
 AbstractBlock.prototype.getCaption = function () {
-  return this.$caption()
+  return this.$caption() === Opal.nil ? undefined : this.$caption()
```
Mogztter: Could you please assign the result of the function `this.$caption()`? Both for consistency and probably performance as the `$caption` method is [not a simple accessor](https://github.com/asciidoctor/asciidoctor/blob/194d191f24dd6a069055c903cd7019772cfa988f/lib/asciidoctor/abstract_block.rb#L244).

```suggestion
  var caption = this.$caption()
  return caption === Opal.nil ? undefined : caption
```

cexbrayat:
@Mogztter Done!

PR #852, opened 2020-03-10
resolves #845 map Substitutors.resolveSubs

### LINKED ISSUE #845: Map Substitutors#resolve_pass_subs
https://github.com/asciidoctor/asciidoctor/blob/2acdcece395a8327373b5f7ebf01c71dcd80ec24/lib/asciidoctor/substitutors.rb#L1225

Mogztter:
resolves #845

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -797,6 +797,48 @@ AbstractNode.prototype.applySubstitutions = function (text, subs) {
 }
 
 /**
+ * Resolve the list of comma-delimited subs against the possible options.
+ *
+ * @param {string} subs - The comma-delimited String of substitution names or aliases.
+ * @param {string} [type] - A String representing the context for which the subs are being resolved (default: 'block').
+ * @param {Array<string>} [defaults] - An Array of substitutions to start with when computing incremental substitutions (default: undefined).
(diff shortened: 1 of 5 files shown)


packages/core/src/asciidoctor-core-api.js
```diff
@@ -793,6 +793,48 @@ AbstractNode.prototype.applySubstitutions = function (text, subs) {
...
+AbstractNode.prototype.resolveSubstitutions = function (subs, type, defaults, subject) {
+  if (typeof type === 'undefined') {
+    type = 'block'
+  }
+  if (typeof defaults === 'undefined') {
+    defaults = Opal.nil
+  }
+  if (typeof subject === 'undefined') {
+    subject = Opal.nil
+  }
```
djencks: Would it be appropriate to also check for null?

Mogztter: I'm not sure, undefined means basically, the parameter was absent. If the value is `null` then you pass explicitly a `null` value:

```js
resolveSubstitutions('attributes+') // type, defaults and subject are undefined
```
```js
resolveSubstitutions('attributes+', null, null, null) // type, defaults and subject are null
```

I think it's a mis-usage of the API so I don't really want to support it by converting `null` to `Opal.nil`.

PR #853, opened 2020-03-10
resolves #844 map Document.playbackAttributes

### LINKED ISSUE #844: Map Document#playback_attributes
https://github.com/asciidoctor/asciidoctor/blob/2acdcece395a8327373b5f7ebf01c71dcd80ec24/lib/asciidoctor/document.rb#L820

Mogztter:
resolves #844

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -1858,6 +1858,43 @@ Document.prototype.hasHeader = function () {
 }
 
 /**
+ * Replay attribute assignments at the block level.
+ *
+ * <i>This method belongs to an internal API that deals with how attributes are managed by the processor.</i>
+ * If you understand why this group of methods are necessary, and what they do, feel free to use them.
+ * <strong>However, keep in mind they are subject to change at any time.</strong>
+ *
+ * @param {Object} blockAttributes - A JSON of attributes
+ * @memberof Document
+ */
+Document.prototype.playbackAttributes = function (blockAttributes) {
+  blockAttributes = toHash(blockAttributes)
+  if (blockAttributes) {
+    var attrEntries = blockAttributes['$[]']('attribute_entries')
+    if (attrEntries && Array.isArray(attrEntries)) {
+      var result = []
+      for (var i = 0; i < attrEntries.length; i++) {
+        var attrEntryObject = attrEntries[i]
+        if (attrEntryObject && typeof attrEntryObject === 'object' && attrEntryObject.constructor.name === 'Object') {
+          attrEntryObject.$name = function () {
+            return this.name
+          }
+          attrEntryObject.$value = function () {
+            return this.value
+          }
+          attrEntryObject.$negate = function () {
+            return this.negate
+          }
+        }
+        result.push(attrEntryObject)
+      }
+      blockAttributes['$[]=']('attribute_entries', result)
+    }
+  }
+  this.$playback_attributes(blockAttributes)
+}
+
+/**
  * Delete the specified attribute from the document if the name is not locked.
  * If the attribute is locked, false is returned.
  * Otherwise, the attribute is deleted.
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,7 @@ Bug Fixes::
 Improvements::
 
 * Map `Document.restoreAttributes` (#843)
+* Map an internal API `Document.playbackAttributes` (#844)
 
 
 == v2.1.1 (2020-02-01)
--- a/packages/core/spec/share/asciidoctor-spec.js
+++ b/packages/core/spec/share/asciidoctor-spec.js
@@ -1100,6 +1100,49 @@ content`)
           }
         })
       })
+
+      it('should playback attributes', function () {
+        const content = `= Title
+:next-section:
+
+This is a preamble!
+
+:next-section: First section
+
+== First section
+
+:next-section: Second section
+
+== Second section
+`
+        const doc = asciidoctor.load(content)
+        const blocks = doc.getBlocks()
+        expect(blocks.length).to.equal(3)
+        for (let i = 0; i < blocks.length; i++) {
+          const block = blocks[i]
+          if (i === 0) {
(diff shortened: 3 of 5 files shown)


packages/core/src/asciidoctor-core-api.js
```diff
@@ -1857,6 +1857,41 @@ Document.prototype.hasHeader = function () {
...
+ * Replay attribute assignments at the block level
+ *
+ * @param {Object} blockAttributes - A JSON of attributes
+ *
+ * @memberof Document
+ */
+Document.prototype.playbackAttributes = function (blockAttributes) {
+  blockAttributes = toHash(blockAttributes)
+  if (blockAttributes) {
+    var attrEntries = blockAttributes['$[]']('attribute_entries')
```
djencks: Why not const?  Same for just about everything here.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -1857,6 +1857,41 @@ Document.prototype.hasHeader = function () {
   return this['$header?']()
 }
 
+/**
+ * Replay attribute assignments at the block level
+ *
+ * @param {Object} blockAttributes - A JSON of attributes
```
djencks: This might be misleading.  The only real use case for this is to playback the actual document attributes, which I would expect to be an Opal hash rather than JSON.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -1857,6 +1857,41 @@ Document.prototype.hasHeader = function () {
...
+ */
+Document.prototype.playbackAttributes = function (blockAttributes) {
+  blockAttributes = toHash(blockAttributes)
+  if (blockAttributes) {
+    var attrEntries = blockAttributes['$[]']('attribute_entries')
+    if (attrEntries && Array.isArray(attrEntries)) {
+      var result = []
+      for (var i = 0; i < attrEntries.length; i++) {
+        var attrEntry = attrEntries[i]
+        var attrEntryObject = attrEntry
```
djencks: Why two variables?  (also why not const?)

packages/core/src/asciidoctor-core-api.js
```diff
@@ -1857,6 +1857,41 @@ Document.prototype.hasHeader = function () {
...
+          attrEntryObject.$name = function () {
+            return this.name
+          }
+          attrEntryObject.$value = function () {
+            return this.value
+          }
+          attrEntryObject.$negate = function () {
+            return this.negate
+          }
+        }
```
djencks: Does this essentially turn a json object into an Opal-compatible object?

mojavelinux:
I'll admit I'm a bit uneasy about mapping these APIs. The APIs and the behavior were never meant to be public. (That's why they seem so awkward). Perhaps we can add a disclaimer to them that it's an advanced feature, then focus our attention the proper solution upstream, which is to proxy the document attributes so they reflect the correct state when viewed from the tree.

packages/core/src/asciidoctor-core-api.js
Mogztter: I'm being conservative, `const` and `let` are not supported in IE < 11 but arguably nobody should use IE < 11.
I will probably drop `var` in a major release so it does not come as a surprise in a feature version.

packages/core/src/asciidoctor-core-api.js
Mogztter: It's a mistake, nice catch 👍

packages/core/src/asciidoctor-core-api.js
Mogztter: Yes, Asciidoctor (Ruby) will call the `attr_reader` on the `AttributeEntry` class: https://github.com/asciidoctor/asciidoctor/blob/88bc891d6a0143f3a2142f41d29cabe3cb21fe3f/lib/asciidoctor/document.rb#L91 so the object must respond to `$name`, `$value` and `$negate`.

In other words, I'm creating a very primitive `AttributeEntry` class.

djencks:
I'm uneasy too. Rather than (only) marking them 'advanced' I think it would be better to indicate that a better solution is hoped for in the future and these methods may disappear.

packages/core/src/asciidoctor-core-api.js
Mogztter: Actually if you are using the JavaScript API, `block.getAttributes()` will also return JSON.
`Opal.hash` should not leak in the Asciidoctor.js API.

Mogztter:
> I'll admit I'm a bit uneasy about mapping these APIs. The APIs and the behavior were never meant to be public. (That's why they seem so awkward). 

I agree but right now they can be useful in some scenarios

> Perhaps we can add a disclaimer to them that it's an advanced feature, then focus our attention the proper solution upstream, which is to proxy the document attributes so they reflect the correct state when viewed from the tree.
> I'm uneasy too. Rather than (only) marking them 'advanced' I think it would be better to indicate that a better solution is hoped for in the future and these methods may disappear.

Sure, as soon as Asciidoctor core provides a better alternative, I will remove these methods and map the new ones.
I guess that Asciidoctor core will remove these methods in a major release (because even if they are not widely used, they are marked as "public").

Do you have suggestions about what we should add in the JavaScript documentation? 

```js
/**
 * Replay attribute assignments at the block level.
 * /!\ This low-level API will probably be removed in the next major version /!\
 */
```

mojavelinux:
> I agree but right now they can be useful in some scenarios

I definitely recognize the need.

> I guess that Asciidoctor core will remove these methods in a major release (because even if they are not widely used, they are marked as "public").

We really have no idea at this point.

They are public only because they haven't been updated to be private. Remember that prior to Asciidoctor 2, there was no delineation. So it's more that a decision just hasn't been made yet.

> Do you have suggestions about what we should add in the JavaScript documentation?

I think it should be a very direct message.

```
This method is part of an internal API that deals with how attributes are managed by the processor. If you understand why these methods are necessary, and what they do, feel free to use them. However, keep in mind they are subject to change at any time.
```

Mogztter:
@mojavelinux Sounds good, thanks for your input! I will update the documentation to use a clear and direct message as suggested.

PR #854, opened 2020-03-10
Implemented Table, Column and Cell

henriette-einstein:
Added 

* Table
* Column
* Cell

and some test cases

### DIFF
--- a/packages/core/spec/share/asciidoctor-spec.js
+++ b/packages/core/spec/share/asciidoctor-spec.js
@@ -1939,7 +1939,7 @@ America/New_York
         expect(table.getBody()[0][0].getWidth()).to.equal(1)
         expect(table.getHead()[0][0].getStyle()).to.equal(undefined)
       })
-      it('should return tablewidth', function () {
+      it('should return tablewidth 75%', function () {
         const options = {}
         const source = `
 [width=75%]
@@ -1953,6 +1953,66 @@ America/New_York
         const table = doc.getBlocks()[0]
         expect(table.getContext()).to.equal('table')
         expect(table.getAttribute('width')).to.equal('75%')
+      })
+      it('should handle colspan and rowspan', function () {
+        const options = {}
+        const source = `
+|===
+
+|Column 1, row 1 |Column 2, row 1 |Column 3, row 1 |Column 4, row 1
+
+|Column 1, row 2
+2.3+|Content in a single cell that spans over rows and columns
+|Column 4, row 2
+
+|Column 1, row 3
+|Column 4, row 3
+
+|Column 1, row 4
+|Column 4, row 4
+|===`
+        const doc = asciidoctor.load(source, options)
+        const table = doc.getBlocks()[0]
+        expect(table.getContext()).to.equal('table')
+        expect(table.getBody()[1][1].getColumnSpan()).to.equal(2)
+        expect(table.getBody()[1][1].getRowSpan()).to.equal(3)
+      })
+      it('should handle cell duplication', function () {
+        const options = {}
+        const source = `
+|===
+
+|Column 1, row 1 |Column 2, row 1 |Column 3, row 1 |Column 4, row 1
+
+3*|Same text
+|Different text
+|===`
+        const doc = asciidoctor.load(source, options)
+        const table = doc.getBlocks()[0]
+        expect(table.getContext()).to.equal('table')
+        expect(table.getBody()[1][0].getText()).to.equal('Same text')
+        expect(table.getBody()[1][1].getText()).to.equal('Same text')
+        expect(table.getBody()[1][2].getText()).to.equal('Same text')
+        expect(table.getBody()[1][3].getText()).to.equal('Different text')
+      })
+      it('should handle CSV format', function () {
+        const options = {}
+        const source = `
+[%header,format=csv]
+|===
+Artist,Track,Genre
+Baauer,Harlem Shake,Hip Hop
+The Lumineers,Ho Hey,Folk Rock
+|===`
+        const doc = asciidoctor.load(source, options)
+        const table = doc.getBlocks()[0]
+        expect(table.getContext()).to.equal('table')
+        expect(table.getHead()[0][0].getText()).to.equal('Artist')
+        expect(table.getHead()[0][1].getText()).to.equal('Track')
+        expect(table.getHead()[0][2].getText()).to.equal('Genre')
+        expect(table.getBody()[0][0].getText()).to.equal('Baauer')
+        expect(table.getBody()[0][1].getText()).to.equal('Harlem Shake')
+        expect(table.getBody()[0][2].getText()).to.equal('Hip Hop')
       })
     })
   })
--- a/packages/core/spec/share/asciidoctor-spec.js
+++ b/packages/core/spec/share/asciidoctor-spec.js
@@ -1689,7 +1689,7 @@ America/New_York
               const self = this
               self.process(function (parent) {
                 const table = asciidoctor.Table.create(parent, {})
-                table.setAttribute('rowcount', 1)
+                table.setRowCount(1)
                 const firstColumn = asciidoctor.Table.Column.create(table, 0, {})
                 const firstCell = asciidoctor.Table.Cell.create(firstColumn, 'Cell in column 1, row 1', {}, {})
                 table.getBody().push([firstCell])
@@ -1704,6 +1704,43 @@ America/New_York
 <tbody>
 <tr>
 <td class="tableblock halign-left valign-top"><p class="tableblock">Cell in column 1, row 1</p></td>
+</tr>
+</tbody>
+</table>`)
+        } finally {
+          asciidoctor.Extensions.unregisterAll()
+        }
+      })
+      it('should create a table, a column and a cell with a row span', function () {
+        try {
+          asciidoctor.Extensions.register(function () {
+            this.blockMacro('table', function () {
+              const self = this
+              self.process(function (parent) {
+                const table = asciidoctor.Table.create(parent, {})
+                table.setRowCount(2)
+                const firstColumn = asciidoctor.Table.Column.create(table, 0, {})
+                const firstColumnFirstCell = asciidoctor.Table.Cell.create(firstColumn, 'Asciidoctor', {}, {})
+                firstColumnFirstCell.setRowSpan(2)
+                const secondColumn = asciidoctor.Table.Column.create(table, 0, {})
+                const secondColumnFirstCell = asciidoctor.Table.Cell.create(secondColumn, 'Awesome way to write documentation', {}, {})
+                const secondColumnSecondCell = asciidoctor.Table.Cell.create(secondColumn, 'Works on the JVM', {}, {})
+                table.getBody().push([firstColumnFirstCell, secondColumnFirstCell], [secondColumnSecondCell])
+                return table
+              })
+            })
+          })
+          const html = asciidoctor.convert('table::[]')
+          expect(html).to.equal(`<table class="tableblock frame-all grid-all stretch">
+<colgroup>
+</colgroup>
+<tbody>
+<tr>
+<td class="tableblock halign-left valign-top" rowspan="2"><p class="tableblock">Asciidoctor</p></td>
+<td class="tableblock halign-left valign-top"><p class="tableblock">Awesome way to write documentation</p></td>
+</tr>
+<tr>
+<td class="tableblock halign-left valign-top"><p class="tableblock">Works on the JVM</p></td>
 </tr>
 </tbody>
 </table>`)
@@ -1796,7 +1833,7 @@ America/New_York
         expect(table.hasHeaderOption()).to.equal(true)
         expect(table.hasRows()).to.equal(true)
       })
-      it('should return 3 for getRowcount() since header row is also counted', function () {
+      it('should return 3 for getRowCount() since header row is also counted', function () {
         const options = {}
         const source = `
 |===
@@ -1808,9 +1845,9 @@ America/New_York
         const doc = asciidoctor.load(source, options)
         const table = doc.getBlocks()[0]
         expect(table.getContext()).to.equal('table')
-        expect(table.getRowcount()).to.equal(3)
+        expect(table.getRowCount()).to.equal(3)
       })
-      it('should return 1 for getColcount() ', function () {
+      it('should return 1 for getColumnCount() ', function () {
         const options = {}
         const source = `
 |===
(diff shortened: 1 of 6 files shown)


packages/core/spec/share/asciidoctor-spec.js
```diff
@@ -1665,6 +1665,117 @@ America/New_York
         expect(html).to.include('<style>pre.prism{background-color: lightgrey}</style>')
       })
     })
+    describe('Table stuff', function () {
+      it('should return true for hasRows()', function () {
+        const options = {}
+        const doc = asciidoctor.load('= Document Title\n\n|===\n\n|This is a title\n|===', options)
```
Mogztter: Could you please use template literals? I find it easier to read.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,402 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+Table.create = function (parent, attributes) {
+  return this.$new(parent, attributes)
+}
+
+/**
+ * Check, if a given option is set
+ * @param {string} opt the option to check
+ * @returns {boolean}
+ */
+Table.prototype.hasOption = function (opt) {
```
Mogztter: As a first step I would like to map only existing methods from Asciidoctor core and then we can add utility functions like this one.
Could you please move utility functions in another pull request?

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,402 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+/**
+ * Create a new Column element
+ * @param {Table} table
+ * @param {number} index
+ * @param {Object|undefined} attributes
+ * @returns Column
+ */
+Column.create = function (table, index, attributes) {
+  let ret = this.$new(table, index, attributes)
+  return ret
```
Mogztter: It could be inlined right?

```suggestion
  return this.$new(table, index, attributes)
```

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,402 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+
+/**
+ * Create a new Column element
+ * @param {Table} table
+ * @param {number} index
+ * @param {Object|undefined} attributes
+ * @returns Column
+ */
+Column.create = function (table, index, attributes) {
+  let ret = this.$new(table, index, attributes)
```
Mogztter: I think we should call `toHash` on attributes.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,402 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+  return this.getAttribute('valign')
+}
+
+/**
+ * Get the style of this cell
+ * @returns {string}
+ */
+Column.prototype.getStyle = function () {
+  let ret = this.style
+  return (ret !== Opal.nil) ? ret : undefined
```
Mogztter: Could you please use the following format (for consistency):

```suggestion
  var style = this.style
  return style === Opal.nil ? undefined : style
```

So:
- `var` instead of `let` (we might use `let` and `const` soon but for now we stick with `var` mostly to maintain compatibility)
- use the name of the method as the name of the variable (`this.style` -> `var style`)
- compare the variable to `Opal.nil` and return undefined or the value (`style === Opal.nil ? undefined : style`) without parenthesis.

settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
Mogztter: I don't mind adding this file. Is that a Visual Studio Code settings? Is it safe? (ie. I don't want to override user-specific settings or force a specific settings)

Mogztter:
Overall that's look very good :+1: 
I left a few comments.

We also maintain an handcrafted TypeScript Definition File: https://github.com/asciidoctor/asciidoctor.js/blob/master/packages/core/types/index.d.ts

So we will need to add the new functions in here too. Let me know if you are not familiar with TypeScript or if you need help to get started. I can also follow-up on your work and add the missing definition in the `index.d.ts` file.

settings.json
henriette-einstein: The settings are defaults, if you use the javascript standard lint rules. However, I would not merge this, since other users may have different settings

packages/core/src/asciidoctor-core-api.js
henriette-einstein: Done. Will be included in the next commit

packages/core/src/asciidoctor-core-api.js
henriette-einstein: Done. Will be included in the next commit

packages/core/src/asciidoctor-core-api.js
henriette-einstein: That may be in more than one place. I'll check it before committing

packages/core/spec/share/asciidoctor-spec.js
henriette-einstein: Done. Will be included in the next commit

packages/core/src/asciidoctor-core-api.js
henriette-einstein: Done. Will be included in the next commit

henriette-einstein:
Sorry, I am not familiar with the TypeScript definition files

henriette-einstein:
@Mogztter I have no idea, why the build

Build / build (ubuntu-latest, 13.x, master) (pull_request)

failed. What can I do?

Mogztter:
> Sorry, I am not familiar with the TypeScript definition files

No worries, I will take care of it 👍

> I have no idea, why the build
Build / build (ubuntu-latest, 13.x, master) (pull_request)
failed. What can I do?

I think it's an intermittent failure, I might need to slightly increase the timeout. Nothing to worry about but now GitHub actions does not find the commit reference 🤔 
Can you please force push to trigger a new build?

settings.json
Mogztter: Ok, in this case can you please remove it?

Mogztter:
Build is green, nice 🎉 
If that's OK with you, I will add commits on top of your work to adjust a few things and add the type definition.

henriette-einstein:
That is fine with me.

danyill:
Thanks @henriette-einstein, thanks @Mogztter much appreciated.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,381 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+ */
+Cell.prototype.getWidth = function () {
+  return this.getAttribute('width')
+}
+
+/**
+ * Get the colpcwidth of this cell
+ * @returns {string|undefined}
+ */
+Cell.prototype.getColpcwidth = function () {
```
Mogztter: I think we should use long form (ie. `getColumnPercentageWidth`) because the rest of the API is using long form.
At some point, we might introduce alias:

- `getColPercentWidth`
- `getColPcWidth`

But we need to be careful because users will expect `Col` to be equivalent to `Col` in other functions as well.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,381 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+Cell.prototype.getLines = function () {
+  var lines = this['$lines']()
+  return lines === Opal.nil ? undefined : lines
+}
+
+/**
+ * Get the lineno of the cell
+ * @returns {string}
+ */
+Cell.prototype.getLineno = function () {
```
Mogztter: It should be `getLineNumber`

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,381 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+  var value = this.getAttributes()[name]
+  return value === Opal.nil ? undefined : value
+}
+
+/**
+ * Get the colspan of this {@link Cell} node
+ * @memberof Cell
+ * @returns {number}  An Integer of the number of columns this cell will span (default undefines)
+ */
+Cell.prototype.getColspan = function () {
```
Mogztter: If we apply the same rules, then `getColumnSpan`

packages/core/src/asciidoctor-core-api.js
henriette-einstein: Changed name to getColumnPercentageWidth. No alias defined by now

packages/core/src/asciidoctor-core-api.js
henriette-einstein: Changed to getColumnSpan

packages/core/src/asciidoctor-core-api.js
henriette-einstein: Changed to getLineNumber

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,387 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+  return this.getAttribute('width')
+}
+
+/**
+ * Get the column width in percentage of this cell.
+ * @returns {string|undefined}
+ * @memberof Cell
+ */
+Cell.prototype.getColumnPercentageWidth = function () {
+  return this.getAttribute('colpcwidth')
```
Mogztter: @mojavelinux How do you feel about these utilities functions to "quickly" access the underlying attribute?

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,387 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+}
+
+/**
+ * Check if the table has the autowidth option set.
+ * @returns {boolean}
+ * @memberof Table
+ */
+Table.prototype.hasAutowidthOption = function () {
+  var autowidthOption = this.getAttributes()['autowidth-option']
+  return autowidthOption === Opal.nil ? false : autowidthOption === ''
```
Mogztter: @mojavelinux I'm always a bit concerned that someone could use `setAttribute` with a boolean value:

```js
table.setAttribute('autowidth-option', true)
table.hasAutowidthOption() // false?!
```

Do we need to check if the value is truthy instead? 

```suggestion
  return autowidthOption === Opal.nil ? false : autowidthOption
```

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,387 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+Table.prototype.getFoot = function () {
+  return this.rows.foot
+}
+
+/**
+ * Check if the table has rows.
+ * @returns {boolean}
+ * @memberof Table
+ */
+Table.prototype.hasRows = function () {
```
Mogztter: I think it should be `hasBodyRows`. What do you think?

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,387 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+Table.prototype.getBody = function () {
+  return this.rows.body
+}
+
+/**
+ * Get the foot rows of this table.
+ * @returns {Array<Array<Cell>>} - an Array of Array of Cell
+ * @memberof Table
+ */
+Table.prototype.getFoot = function () {
```
Mogztter: `getFootRows`

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,387 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+Table.prototype.hasRows = function () {
+  return this.rows !== Opal.nil && this.rows.body.length > 0
+}
+
+/**
+ * Check if the table has a header.
+ * @returns {boolean}
+ * @memberof Table
+ */
+Table.prototype.hasHeader = function () {
```
Mogztter: `getHeadRows` ?

packages/core/src/asciidoctor-core-api.js
henriette-einstein: You are right. I'll change that

packages/core/src/asciidoctor-core-api.js
henriette-einstein: getFootRows is definitely better. I'll change it

packages/core/src/asciidoctor-core-api.js
```diff
@@ -3047,3 +3047,387 @@ Opal.Asciidoctor.SyntaxHighlighterBase = SyntaxHighlighterBase
...
+ */
+Table.prototype.hasHeaderOption = function () {
+  return this.has_header_option
+}
+
+/**
+ * Check if the table has a footer.
+ * @returns {boolean}
+ */
+Table.prototype.hasFooter = function () {
```
Mogztter: I think we should be consistent with the naming.
Here the underlying property is `foot` so `hasFoot`?

@mojavelinux What is the reasoning behind `foot` and `head` (instead of footer` and `header`)?

packages/core/src/asciidoctor-core-api.js
Mogztter: Thanks!

packages/core/src/asciidoctor-core-api.js
henriette-einstein: It is probably due to the fact, that in HTML the corresponding tags are 'head', 'body' and 'foot'

packages/core/src/asciidoctor-core-api.js
henriette-einstein: I'll change that and also hasFooter to hasFootRows

Mogztter:
I did a little cleanup, it's looking very good.
I want to hear Dan's opinion on a few specific cases and then I think we will be ready to move forward!

Thanks again for the heavy lifting :muscle:

henriette-einstein:
Thanks for the great project. You did the heavy work! I'll add some more tascaseos and will commit in a few minutes

packages/core/src/asciidoctor-core-api.js
Mogztter: That's a good point but I wonder if it would do more harm than good in the API (ie. it can be confusing to use `hasHeader` but `getHead`).

packages/core/src/asciidoctor-core-api.js
henriette-einstein: I'll also add a method hasBodyRows

Mogztter:
> Thanks for the great project. You did the heavy work!

:blush: 
Then a team effort!

> I'll add some more test cases and will commit in a few minutes

Perfect :ok_hand:

packages/core/src/asciidoctor-core-api.js
henriette-einstein: The methods are now renamed to hasHeadRows, hasBodyRows and hasFootRows

Mogztter:
I've rebased on master (force-push), added the TypeScript definitions and a changelog.
I've also fix a few inconsistencies/differences.
It should be ready to go!

PR #886, opened 2020-03-23
✨ resolves #884 use a local HTTP server

### LINKED ISSUE #884: Use a local HTTP server in the tests suite
The tests suite is regularly failing because it exceeds a specified timeout (5 seconds) to fetch a resource from a remote server.

I think we should start a local HTTP server. The main goal is to make the tests suite more reliable.

Mogztter:
resolves #884

### DIFF
--- a/packages/core/package-lock.json
+++ b/packages/core/package-lock.json
@@ -2101,6 +2101,65 @@
         "safe-json-parse": "~1.0.1"
       }
     },
+    "body-parser": {
+      "version": "1.19.0",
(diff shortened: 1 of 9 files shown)


Mogztter:
We need a local HTTP to test ECMAScript Module in the browser otherwise the following error is thrown:

> Failed to load module script: The server responded with a non-JavaScript MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec.

PR #898, opened 2020-03-31
Add bytefield-svg links

brunchboy:
Honest, I didn’t choose my name to try to be on the top of the list, but you definitely seem to have them sorted alphabetically! 😄

### DIFF
--- a/docs/modules/extend/pages/extensions/ecosystem.adoc
+++ b/docs/modules/extend/pages/extensions/ecosystem.adoc
@@ -3,6 +3,7 @@
 :uri-plantuml-server: https://github.com/plantuml/plantuml-server
 :uri-chartist: https://gionkunz.github.io/chartist-js/
 :uri-kroki: https://kroki.io/
+:uri-bytefield: https://bytefield-svg.deepsymmetry.org/
 
 :uri-ext-interactive-runner-gh: https://github.com/oncletom/asciidoctor-extension-interactive-runner
 :uri-ext-plantuml-gh: https://github.com/eshepelyuk/asciidoctor-plantuml.js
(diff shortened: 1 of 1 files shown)


Mogztter:
>Honest, I didn’t choose my name to try to be on the top of the list, but you definitely seem to have them sorted alphabetically!

No worries! :laughing: 
Speaking of name, I wonder if we should just use `bytefield` instead of `bytefield-svg`? Currently, we only generate SVG but maybe one day we will generate png/jpeg from the same DSL and the suffix "-svg" will be misleading and superfluous. I think we should use the simplest name and, as far as I know, https://www.npmjs.com/package/bytefield is available.
Plus the Asciidoctor extension is named `asciidoctor-bytefield` (not `asciidoctor-bytefield-svg`) so we might want to use the same name for consistency. What do you think ?

brunchboy:
The fundamental architecture of the library is SVG oriented. That is pervasively interwoven with all levels of the code and design. I can’t imagine trying to fold in some other drawing approach, it would be a different project.

brunchboy:
(But I may be missing something, perhaps there would be an easy way to generate the SVG and turn it into something else? But in terms of favoring composition as a design principle, it might be better to wrap such a tool around bytefield-svg, rather than adding it as a feature within it.)

Mogztter:
> But I may be missing something, perhaps there would be an easy way to generate the SVG and turn it into something else?

It might be possible to generate a png/jpeg from an SVG with a Node library. My point is that the output format does not necessarily define your project or maybe it does and then the `-svg` suffix is fine. But in this case, I think we should use `asciidoctor-bytefield-svg` instead of `asciidoctor-bytefield`.

>  But in terms of favoring composition as a design principle, it might be better to wrap such a tool around bytefield-svg, rather than adding it as a feature within it.

That's a good point :+1:

brunchboy:
Yes, I think that the underlying library will remain svg-driven, but perhaps the extension will grow to include an ability to convert the SVG to other formats. Perhaps David Jencks’ `antora-generic-svg-plugin` could do that as well someday.

But if you think that is too tentative, I am happy to rename the extension to `asciidoctor-bytefield-svg`.

Mogztter:
> Yes, I think that the underlying library will remain svg-driven,

Ok :+1: 

> But if you think that is too tentative, I am happy to rename the extension to asciidoctor-bytefield-svg.

That's up to you but I think it makes sense to keep the same name.
Sorry for the digression :sweat_smile:

brunchboy:
No apology needed! It is always valuable to think of names and long-term implications, and I value your perspective, thank you.

PR #921, opened 2020-04-12
resolves #920 replace Opal.nil with undefined in AbstractNode.getAttributes

### LINKED ISSUE #920: getAttributes() returns Opal.nil as value (instead of undefined)
The API must not leak `Opal.nil` value. 
So the `getAttributes` function must replace `Opal.nil` values with `undefined`.

Please note that this is also problematic because the behavior is not consistent with `getAttribute`:

```js
getAttribute('foo') // undefined
getAttributes().foo // Opal.nil
```

Mogztter:
resolves #920

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -14,18 +14,24 @@ var toHash = function (object) {
  */
 var fromHash = function (hash) {
   var object = {}
-  var data = hash.$$smap
-  for (var key in data) {
-    object[key] = data[key]
+  if (hash) {
+    var data = hash.$$smap
+    for (var key in data) {
+      var value = data[key]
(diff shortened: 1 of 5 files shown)


packages/core/src/asciidoctor-core-api.js
```diff
@@ -851,7 +851,14 @@ AbstractNode.prototype.getNodeName = function () {
  * @memberof AbstractNode
  */
 AbstractNode.prototype.getAttributes = function () {
-  return fromHash(this.attributes)
```
Mogztter, line 854: In theory we could do it in `fromHash` but in practice it has side effects.

Mogztter: I changed my mind, I think it's more consistent to do it in `fromHash`. It also avoids an additional loop over the attributes.

PR #923, opened 2020-04-13
Add Handlebars toJSON helper function in the tests

henriette-einstein:
I have added some Handlebars helper functions and some test cases.
It is really ugly to write test cases that have to respect the correct indentation. Does somebody know a better way to test that?

### DIFF
--- a/packages/core/spec/fixtures/templates/handlebars-json/helpers.js
+++ b/packages/core/spec/fixtures/templates/handlebars-json/helpers.js
@@ -1,0 +1,316 @@
+function isString (obj) {
+  return Object.prototype.toString.call(obj) === '[object String]'
+}
+
+/**
+ * Helper function to determine, what kind of block
+ * the given block is. Must be enhanced, once table is
+ * provided as a class in AsciDoctor
+ * @param {AbstractBlock} node the node to test
+ */
+function getBlockType (node) {
+  const nodeName = node.getNodeName()
+  if (['document', 'section'].includes(nodeName)) {
+    return nodeName
+  } else if (['olist', 'ulist', 'dlist', 'colist'].includes(nodeName)) {
+    return 'list'
+  } else {
+    return 'block'
+  }
+}
+
+function getOptions (node) {
+  const options = {}
+  for (const key in node.getAttributes()) {
+    if (key.endsWith('-option')) {
+      const last = key.lastIndexOf('-option')
+      const name = key.substring(0, last)
+      options[name] = true
+    }
+  }
+  return options
+}
+
+/**
+ * Return the JSON attributes common to all
+ * objects
+ * @param {AbstractNode} node the node to process
+ */
+function getCommonNodeInfo (node) {
+  return {
+    id: node.getId(),
+    nodeName: node.getNodeName(),
+    roles: node.getRoles(),
+    isInline: node.isInline(),
+    isBlock: node.isBlock(),
+    isDocument: node.getNodeName() === 'document',
+    attributes: node.getAttributes(),
+    options: getOptions(node),
+    role: node.getRoles().length > 0 ? node.getRole() : '',
+    isReftext: node.isReftext(),
+    reftext: node.getReftext()
+  }
+}
+
+/**
+ * Return the JSON common to all block objects
+ * @param {AbstractNode} node the node to process
+ */
+function getAbstractBlockInfo (node) {
+  const ret = getCommonNodeInfo(node)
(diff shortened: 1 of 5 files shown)


Mogztter:
It made me think, maybe we should work on a AsciiDoc to JSON converter and then use a two-step converter to pass the JSON to Handlebars.
This format could then be used in toolings/editors see https://editorjs.io.

Could you please move your templates into a dedicated directory `fixtures/templates/handlebars-json`?

> It is really ugly to write test cases that have to respect the correct indentation. Does somebody know a better way to test that?

Indeed.
I think you should read each lines and strip leading spaces so indentation does not matter.

henriette-einstein:
> Could you please move your templates into a dedicated directory `fixtures/templates/handlebars-json`?
> 
I moved the files to the new directory

henriette-einstein:
> It made me think, maybe we should work on a AsciiDoc to JSON converter and then use a two-step converter to pass the JSON to Handlebars.
> This format could then be used in toolings/editors see https://editorjs.io.
> 
I think it would be a great idea to have a JSON representation of the AST. That would definitely ease the usage in several tools. I don't know editors.io, but I played around with CKEditor5. 
The editor has a very interesting concept of distinguishing the data-model and the view model. So basically you have to convert from an AsciiDoctor abstract node to a CKEditor view model and back to implement a WYSIWYG editor. More detail can be found here: 
https://ckeditor.com/docs/ckeditor5/latest/framework/guides/architecture/editing-engine.html#model

henriette-einstein:
I added some regex to make the contains function ignore additional whitespace and CR/LF. Looks weird but works
```
.replace(/\s{2,}|\r?\n|\r/g, '')
```

henriette-einstein:
I don't know why the build failed. Looks like a problem not related to the files in this pull-request

Mogztter:
> I don't know why the build failed. Looks like a problem not related to the files in this pull-request

The macOS runner is a bit flaky and Windows, well, sometimes it's working and sometimes not... :man_shrugging:

henriette-einstein:
> > I don't know why the build failed. Looks like a problem not related to the files in this pull-request
> 
> The macOS runner is a bit flaky and Windows, well, sometimes it's working and sometimes not... 🤷‍♂️

Hihi, always the same problems with windows...

packages/core/spec/fixtures/templates/handlebars-json/helpers.js
```diff
@@ -0,0 +1,309 @@
...
+ * @param {string} name the name of the attribute
+ * @param {value} value the value of the attribute
+ */
+function toAttribute (name, value) {
+  return value !== undefined && value.length !== 0 ? ` ${name}="${value}"` : ''
+}
+
+function configure (options) {
+  options = options || {}
+  const handlebars = options.engine || require('handlebars')
```
Mogztter, line 276: With https://github.com/asciidoctor/asciidoctor.js/pull/926 it will be:

```suggestion
  options = options || { handlebars: { } }
  const handlebars = options.handlebars.environment || require('handlebars')
```

Mogztter: @henriette-einstein sorry since #926 is not yet merged, the property `handlebars` can be `undefined` :sweat_smile:

henriette-einstein: I have added an additional check for that.

packages/core/spec/fixtures/templates/handlebars-json/helpers.js
```diff
@@ -0,0 +1,316 @@
...
+function getSectionInfo (node) {
+  const ret = getAbstractBlockInfo(node)
+  ret.index = node.getIndex()
+  ret.sectionName = node.getSectionName()
+  ret.isSpecial = node.isSpecial()
+  ret.isNumbered = node.isNumbered()
+  ret.caption = node.getCaption()
+  ret.name = node.getName()
+  ret.sectionNumber = calcSectionNumber(node)
+  ret.content2 = node.getContent()
```
Mogztter, line 179: note to self, don't forget to rename to `content` once #930 is merged.

henriette-einstein:
Sorry, I changed the names too early. Now waiting for #930. Sometimes it helps to read more carefully..

Mogztter:
@henriette-einstein I've merged #930 and rebased + squashed your work on top of master

Mogztter:
Many thanks for your work that helped uncover this bug #924 :+1:

PR #927, opened 2020-04-16
 Replace header_footer option with standalone

Mogztter:
My understanding is that `header_footer` is "legacy" and we should encourage users to use `standalone`? The term standalone is great because it translates better when we explain that we can produce "'embedded" or "standalone" document using `--embedded` from the CLI or the `standalone` option.

@mojavelinux Is that correct?

Also remove "just".


mojavelinux:
Yep, that's correct.

Issue #801, opened 2020-01-05
Produce an ESM compatible file in the distribution (in addition to UMD)


Mogztter:
With strict mode enabled (see https://github.com/asciidoctor/asciidoctor.js/pull/763) we are now able to produce an ESM compatible file.
For compatibility reason, we should keep the UMD file and produce an additional file:

```
dist
├── browser
│   ├── asciidoctor.js
│   └── asciidoctor.min.js
├── css
│   └── asciidoctor.css
├── esm
│   ├── asciidoctor.js
│   └── asciidoctor.min.js
├── graalvm
│   └── asciidoctor.js
└── node
    └── asciidoctor.js

5 directories, 7 files
```

Mogztter:
We should use the field "module" in the `package.json` file to reference this new ESM compatible file.
https://stackoverflow.com/a/42817320/3600979

To be consistent, we should export the "constructor" function:

```html
<script type="module">
import Asciidoctor from './build/asciidoctor-browser.js';

const asciidoctor = Asciidoctor()
console.log(asciidoctor.convert('Hello *world*'))
// prints:
// <div class="paragraph">
// <p>Hello <strong>world</strong></p>
// </div>
</script>
```

We should also export the `Opal` global for advanced usage.
We might also export a "default" instance and/or common functions.


```html
<script type="module">
import { Asciidoctor, asciidoctor as processor, convert, Opal } from './build/asciidoctor-browser.js';

const asciidoctor = Asciidoctor()
const nil = Opal.nil
console.log(asciidoctor.convert('Hello *world*'))
console.log(processor.convert('Hello *world*'))
console.log(convert('Hello *world*'))
// prints:
// <div class="paragraph">
// <p>Hello <strong>world</strong></p>
// </div>
</script>
```

iovuio:
How can  `./build/asciidoctor-browser.js` be obtained?

Mogztter:
@iovuio You mean from the example above? It was just an example to illustrate the possibilities so this ESM compatible file does not exist yet. Since it's a breaking change it's planned for Asciidoctor.js 3.0.0.

oncletom:
+1 for the `module` field, it would make Asciidoctor compatible with Deno — cf. [its search engine Pika](https://www.pika.dev/search?q=asciidoctor).

Mogztter:
I can produce an ESM that's going to work in the browser but we are not ready yet to produce an ESM for Node/Deno.

[...]

PR #1000, opened 2020-06-06
🐛 Fix tests against an ECMAScript Module

Mogztter:
We must use a local HTTP to run the tests against a ECMAScript Module.
Otherwise the following error is thrown:

> Failed to load module script: The server responded with a non-JavaScript MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec.

Ref #801

### DIFF
--- a/packages/core/src/template-asciidoctor-browser.js
+++ b/packages/core/src/template-asciidoctor-browser.js
@@ -1,6 +1,6 @@
 //{{opalCode}}
 
-export default function (moduleConfig) {
+export default function Asciidoctor(moduleConfig) {
 //{{asciidoctorCode}}
 
 //{{asciidoctorAPI}}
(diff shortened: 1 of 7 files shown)


packages/core/src/template-asciidoctor-browser.js
```diff
@@ -18,3 +18,28 @@ export default function (moduleConfig) {
...
+ * var input = '= Hello, AsciiDoc!\n' +
+ *   'Guillaume Grossetie <ggrossetie@example.com>\n\n' +
+ *   'An introduction to http://asciidoc.org[AsciiDoc].\n\n' +
+ *   '== First Section\n\n' +
+ *   '* item 1\n' +
+ *   '* item 2\n';
+ *
+ * var html = asciidoctor.convert(input);
+ */
+export function convert(input, options) {
```
Mogztter, line 43: Not sure how I could avoid this code duplication :grimacing:

Mogztter: I will focus on fixing the tests and then open a new pull request.

PR #1083, opened 2020-10-19
✨ Add AbstractConvert & ConverterConstructor

### LINKED ISSUE #1077: Export CustomConverter Interface
### Background
I have to use for a typescript project the CustomConverter option for asciidoctor.

### Problem
I realised that the current types do not include a CustomConverter Interface. Interestedly extending form the `Converter` class break (at least for me) with:  
```
TypeError: Cannot read property 'Converter' of undefined

      1 | import { Asciidoctor } from 'asciidoctor'
      2 | 
    > 3 | export default class MyCustomConverter extends Asciidoctor.Converter  {
        |                                                            ^
      6 | }
```

### Solution
This could be easily solved by adding the following to the type definitions:
```typescript
interface CustomConverter {
     convert(node: AbstractNode, transform?: string, opts?: any): string;
}
```
This could be implemented by the `Converter` class and a notable side effect the [`ConverterFactory`](https://github.com/asciidoctor/asciidoctor.js/blob/9e7e3786945f2d5fecdd15811bc62c0eebee7708/packages/core/types/index.d.ts#L3030) could get rid of the any.

bitPogo:
Please note: this changes opts of the convert method as well. It's now
'unknown', so consumer are able to redeclare the type for their need.

closes #1077

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -2980,7 +2980,26 @@ export namespace Asciidoctor {
     function create(): Html5Converter;
   }
 
-  class Converter {
+  interface AbstractConverter {
+    /**
+     * Converts an {AbstractNode} using the given transform.
+     * This method must be implemented by a concrete converter class.
+     *
+     * @param node - The concrete instance of AbstractNode to convert.
+     * @param [transform] - An optional String transform that hints at which transformation should be applied to this node.
+     * If a transform is not given, the transform is often derived from the value of the {AbstractNode#getNodeName} property. (optional, default: undefined)
+     * @param [opts]- An optional JSON of options hints about how to convert the node. (optional, default: undefined)
+     *
+     * @returns the {String} result.
+     */
+    convert(node: AbstractNode, transform?: string, opts?: unknown): string;
+  }
+
+  interface ConverterConstructor {
+    new(backend?: string, opts?: unknown): AbstractConverter;
+  }
+
+  class Converter implements AbstractConverter {
     /**
      * Converts an {AbstractNode} using the given transform.
      * This method must be implemented by a concrete converter class.
@@ -3009,7 +3028,7 @@ export namespace Asciidoctor {
      * @param opts - a JSON of options to pass to the converter (default: {})
      * @returns a {Converter} instance for converting nodes in an Asciidoctor AST.
      */
-    function create(backend: string, opts?: any): Converter;
+    function create(backend: string, opts?: unknown): Converter;
   }
 
   /**
@@ -3027,7 +3046,7 @@ export namespace Asciidoctor {
      * @param converter - The {Converter} instance to register
      * @param backends- A {string} {Array} of backend names that this converter should be registered to handle (optional, default: ['*'])
      */
-    register(converter: any, backends?: string[]): void;
+    register(converter: AbstractConverter | ConverterConstructor, backends?: string[]): void;
 
     /**
      * Retrieves the singleton instance of the converter factory.
--- a/packages/core/types/tests.ts
+++ b/packages/core/types/tests.ts
@@ -510,7 +510,7 @@ interface Transforms {
   [key: string]: (node: Asciidoctor.AbstractNode) => string;
 }
 
-class BlogConverter {
+class BlogConverter implements Asciidoctor.AbstractConverter {
   private readonly baseConverter: Asciidoctor.Html5Converter;
   private readonly transforms: Transforms;
 
@@ -1076,11 +1076,20 @@ assert(blockSubs5[0] === 'specialcharacters');
 let converterRegistry = processor.ConverterFactory.getRegistry();
 assert(typeof converterRegistry.html5 === 'function');
 
-class BlankConverter {
+class BlankConverter implements Asciidoctor.AbstractConverter {
   convert() {
     return '';
   }
 }
+
+const BlankConstructor = BlankConverter as Asciidoctor.ConverterConstructor;
+processor.ConverterFactory.register(BlankConstructor, ['blank']);
+converterRegistry = processor.ConverterFactory.getRegistry();
+assert(typeof converterRegistry.html5 === 'function');
+assert(typeof converterRegistry.blank === 'function');
+assert(typeof processor.ConverterFactory.for('html5') === 'function');
+assert(typeof processor.ConverterFactory.for('blank') === 'function');
+assert(typeof processor.ConverterFactory.for('foo') === 'undefined');
 
 processor.ConverterFactory.register(new BlankConverter(), ['blank']);
 converterRegistry = processor.ConverterFactory.getRegistry();


mojavelinux:
Is unknown the way you declare a generic key-value pair object in TypeScript? Because that's what it should be.

The API docs are technically wrong, because JSON is a string. That's different than a JavaScript object which is what is accepted.

Mogztter:
Thanks @bitPogo I will take a closer look in the next few days.

> The API docs are technically wrong, because JSON is a string. That's different than a JavaScript object which is what is accepted.

You're right...
I don't know why but in my mind JSON was equivalent to "JavaScript Object" and "JSON string" was a JavaScript Object as a string.
Anyway, we should fix this issue in another pull request to replace globally the term JSON where it should be JavaScript Object.

> Is unknown the way you declare a generic key-value pair object in TypeScript? Because that's what it should be.

Not sure if this is the right way to do it but I'm using `interface` to describe a key-value pair object:

https://github.com/asciidoctor/asciidoctor.js/blob/747caef08be1dd14f05a55746b9d4d897497363e/packages/core/types/index.d.ts#L208-L213

The main benefit is that you can still describe the type of some known/expected attributes.

> It's now 'unknown', so consumer are able to redeclare the type for their need.

I'm not familiar with the "unknown" type but it can be pretty useful to be able to redeclare the type.

bitPogo:
> Is unknown the way you declare a generic key-value pair object in TypeScript? Because that's what it should be.
`unknown` is the better any - so a consumer has to define what it is and cannot make false assumptions (and if they want to use any, they can override it). In difference to `any` you even have to cast or redefine it. See [here](https://stackoverflow.com/questions/51439843/unknown-vs-any). But in this case it does not even bother since we do not use it. The redefinition must happen when you use effectively the parameter - so the consumer side. This person should known how that thing  what is in use looks like.
However if you mean [`Record<string><unknown>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype), I can change it to that. I used also `unknown`, since I  really not known, which type is intended. So when it should be just a simple HashMap/Dictionary (I have no clue how Ruby call it), then is `Record<string><unknown>` really the way to go.

> Not sure if this is the right way to do it but I'm using interface to describe a key-value pair object:
[Callout are probably similar](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMJwDYYPYFczIQAekIAJgM7IBKEC2UZAPBWFKAOYA0yuIA1iGwB3EAD5kAbwBQyZMDIB+AFzJW7EBwDcs5AzKhMK5CFwBbAEbQdAX2nT6IVsjir0WPAQC8U3XIWqAOQAjABMAMyBXH56jIYYqgAsodF20kA):
```typescript
interface Callout extends Record<string, unknown> {
  id?: string;
  ordinal?: number;
}
```
Regarding the docs - I simply c&p them. If it is really a `object` (aka HashMap), I can change that, too.

Since there are some other `any`'s floating around, maybe this could be the start to get rid of them. So this would also clear one of the points in the to  do.

Short story long...just tell me, what you like I should do and I change it.

bitPogo:
Amendment:
Maybe (I do not known where), the repo should have a example to redeclare the type and make it testable. Actually I like how [Vue](https://github.com/vuejs/vue/tree/dev/types/test) does it, but this would need a separate tsconfig, therefore in my head in a separate PR.

packages/core/types/index.d.ts
```diff
@@ -2980,7 +2980,26 @@ export namespace Asciidoctor {
...
+     * If a transform is not given, the transform is often derived from the value of the {AbstractNode#getNodeName} property. (optional, default: undefined)
+     * @param [opts]- An optional JSON of options hints about how to convert the node. (optional, default: undefined)
+     *
+     * @returns the {String} result.
+     */
+    convert(node: AbstractNode, transform?: string, opts?: unknown): string;
+  }
+
+  interface ConverterConstructor {
+	new(backend?: string, opts?: unknown): AbstractConverter;
```
Mogztter, line 2999: neat-pick: could you please replace the tab by spaces?

Mogztter (approved):
Looks good to me :+1: 
I left a nitpick but otherwise it's all good.

Mogztter:
In a way, we are "breaking" the TypeScript API because now custom converters must implement `Asciidoctor.AbstractConverter`.

So this change should land in 2.3 or even 3.0... thoughts?

Mogztter:
Since the master branch contains a major upgrade of Opal (from 0.11 to 1.0) it will be part of Asciidoctor.js 3.0

packages/core/types/index.d.ts
bitPogo: Sure!

bitPogo: done

bitPogo:
Sry, I was a little bit busy last week (even for tabs), so it took a bit longer...
The browser tests are currently broken, this is the reason, why this patch is currently red.
However, if there is anything I can help with, just tell me...It's such a lovely project, so I would be glad to be of service beyond this patch.

Mogztter:
@bitPogo No worries! I fixed the failure in https://github.com/asciidoctor/asciidoctor.js/pull/1089 could you please rebase on master one last time?

Mogztter:
> However, if there is anything I can help with, just tell me...It's such a lovely project, so I would be glad to be of service beyond this patch.

There are two small issues in the API if you want to give it a try: [issues with ":wave: help wanted" label](https://github.com/asciidoctor/asciidoctor.js/labels/%3Awave%3A%20help%20wanted).
Let me know if you are interested and don't hesitate to ask for help/clarification :hugs:

bitPogo:
I made a rebase...but something else broke...I interestingly I cannot reproduce the error locally...

Mogztter:
I think it's a "regression", or a false assertion on our side. @mojavelinux is merging quite a few pull requests on Asciidoctor core.

For reference, the error is:

```
  1) Node.js
       Node.js
         Loading
           should get the line of number of a block when sourcemap is enabled:
     AssertionError: expected 3 to be undefined
      at Context.<anonymous> (spec\share\asciidoctor-spec.js:359:49)
      at processImmediate (internal/timers.js:461:21)
```

> I interestingly I cannot reproduce the error locally...

You will need to run `npm run clean --prefix packages/core` and then `npm run build:quick --prefix packages/core` to build Asciidoctor.js against the latest version of Asciidoctor core (master).

Mogztter:
Confirmed, a bug was resolved on Asciidoctor core: https://github.com/asciidoctor/asciidoctor/pull/3800
Now the line number is available on a preamble, so we need to update the assertion:

```js
// preamble
expect(blocks[0].getLineNumber()).to.equal(3)
expect(blocks[0].getBlocks().length).to.equal(1)
expect(blocks[0].getBlocks()[0].getLineNumber()).to.equal(3)
```

I will fix this issue.

Mogztter:
@bitPogo issue is fixed on master! I will rebase and merge tomorrow 😉

Mogztter:
All good, thanks again @bitPogo :+1:

bitPogo:
Awesome! I will check out some other issues next week and see how I can give there a hand.

Mogztter:
That would be great, thanks :tada:

PR #1089, opened 2020-10-25
Asciidoctor core now uses read_contents to download and embed custom stylesheet

Mogztter:
Adjust the implementation to be compatible with https://github.com/asciidoctor/asciidoctor/pull/3766

### DIFF
--- a/packages/core/lib/asciidoctor/js/asciidoctor_ext/browser/abstract_node.rb
+++ b/packages/core/lib/asciidoctor/js/asciidoctor_ext/browser/abstract_node.rb
@@ -3,7 +3,7 @@ class AbstractNode
   def read_contents target, opts = {}
     doc = @document
     if (Helpers.uriish? target) || ((start = opts[:start]) && (Helpers.uriish? start) && (target = doc.path_resolver.web_path target, start))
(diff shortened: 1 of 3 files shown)


mojavelinux:
Long term, we really need to think about how the definition for what a remote URL is can be controlled. For example, even though chrome:// looks like a remote URL to core, it's actually a specialized file URI for the browser extension. We shouldn't be circumventing allow-uri-read, but changing when allow-uri-read is requested. And that is only once we know, for certain, that the target requires a network call.

Mogztter:
@mojavelinux Yes!
In this case, we should probably clarify the intent of `allow-uri-read`:

> For security reasons, this capability is not enabled by default. To allow content to be read from a URI, you must enable the URI read permission by:

-- [gh.com/asciidoctor/asciidoctor.org/master/docs/_includes/uri-include.adoc](https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/uri-include.adoc#:~:text=To%20allow%20content%20to%20be%20read%20from%20a%20URI)


It should be: "To allow content to be read from a **remote** URI..."

> (And we want enough logic hooks for that).

I think a `is_remote?` check/hook in core would help but since I'm already monkey-patching the whole `read_contents` there's no rush.

PR #1107, opened 2020-11-01
Import Asciidoctor Opal as a CommonJS

Mogztter:
`asciidoctor-opal-runtime` is now available as both a CommonJS and ES6 module.
This is the first step to publish the Node package `@asciidoctor/core` as an ES6 module.

### DIFF
--- a/packages/core/package-lock.json
+++ b/packages/core/package-lock.json
@@ -1894,12 +1894,12 @@
       "dev": true
     },
     "asciidoctor-opal-runtime": {
-      "version": "2.0.0-alpha.1",
-      "resolved": "https://registry.npmjs.org/asciidoctor-opal-runtime/-/asciidoctor-opal-runtime-2.0.0-alpha.1.tgz",
-      "integrity": "sha512-BdZXf0AGOnjJV3fIfCfrUNBTLMX3K78EV2yKCNZhRdbRpcHByfQGHnPf7hkG+gts5nW00HrB2bcSeMDsvhb+wA==",
+      "version": "2.0.0-beta.1",
+      "resolved": "https://registry.npmjs.org/asciidoctor-opal-runtime/-/asciidoctor-opal-runtime-2.0.0-beta.1.tgz",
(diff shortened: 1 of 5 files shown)


Mogztter:
`String#bytes` does not work when JavaScript strict mode is enabled: https://github.com/opal/opal/pull/2194

/cc @mojavelinux

Mogztter:
The following test is also failing:

https://github.com/asciidoctor/asciidoctor.js/blob/c213cb94c702954fa4d61ad7ee7c4454b7432030/packages/core/spec/share/asciidoctor-spec.js#L626-L633


The BOM is not correctly recognized 😞 
It will produce:

```html
<div class="paragraph">
<p>ï»¿= Document Title
:lang: fr
:fixtures-dir: spec/fixtures</p>
</div>
<div class="paragraph">
<p>content is in {lang}</p>
</div>
<div class="paragraph">
<p>Unresolved directive in &lt;stdin&gt; - include::{fixtures-dir}/include.adoc[]</p>
</div>
```

mojavelinux:
> The BOM is not correctly recognized disappointed

It that become of the String#bytes issue?

I think in strict mode, if you try to assign a property to String, it just ignores it. Is that consistent with your observations?

Mogztter:
> It that become of the String#bytes issue?

Not sure... quite a few things changed in Opal. Might be related to https://github.com/opal/opal/pull/2117 but it's just a guess.

> I think in strict mode, if you try to assign a property to String, it just ignores it. Is that consistent with your observations?

Not exactly, we get: `Cannot create property 'bytes' on string 'foo'`.
You can reproduce this issue in your browser (DevTools) using:

```js
(function strict() { 'use strict'; 'foo'.bytes = '' })()
```

mojavelinux:
Ah. What I meant is that in non-strict mode it ignores it, whereas in strict mode it throws an error. Either way, assigning a property to a string literal doesn't work. (It has to be instantiated using new String). So generally, not a good practice. A string should be treated as immutable.

Mogztter:
BOM issue fixed in #1265

Mogztter:
Superseded #1359

Issue #1066, opened 2020-09-13
AbstractNode#getRole() can return Opal.nil, it should return undefined


Mogztter:
As the title says!

bitPogo:
Is that only a TS thing or must a potential patch touch also some logic?

Mogztter:
The implementation must be updated:

https://github.com/asciidoctor/asciidoctor.js/blob/cb0db00597ae7ecfd65cc717f3e67f8c284d5abe/packages/core/src/asciidoctor-core-api.js#L1037-L1039

It should be:

```js
AbstractNode.prototype.getRole = function () { 
  const role = this.$role()
  return role === Opal.nil ? undefined : role
}
```

The related tests and the type definition should also be updated accordingly.

bitPogo:
Got it!

bitPogo:
Does that also applies for [`getRoles`](https://github.com/asciidoctor/asciidoctor.js/blob/0440f5cb5edb7b69d13f61fd808c6ff731b7c710/packages/core/src/asciidoctor-core-api.js#L1059)?

Mogztter:
No, the underlying method `roles` always returns an Array:

https://github.com/asciidoctor/asciidoctor/blob/6ef1cabdc699bce7767024729d133887b488e40f/lib/asciidoctor/abstract_node.rb#L195

Mogztter:
Fixed in #1127

PR #1127, opened 2020-11-09
🐛 Fix AbstractNode#getRole

bitPogo:
see #1066

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -1035,7 +1035,8 @@ AbstractNode.prototype.isRole = function (expectedValue) {
  * @memberof AbstractNode
  */
 AbstractNode.prototype.getRole = function () {
-  return this.$role()
+  const role = this.$role()
+  return role === Opal.nil ? undefined : role
 }
 
 /**
--- a/packages/core/spec/share/asciidoctor-spec.js
+++ b/packages/core/spec/share/asciidoctor-spec.js
@@ -673,6 +673,20 @@ This is another paragraph.
         expect(localSeconds).to.be.within(seconds - 10, seconds + 10)
       })
 
+      it('should get block role', function () {
+        const doc = asciidoctor.load(`= Title
+
+[.foreword]
+== Section with role
+
+== Section without role
+`)
+        const sectionWithRole = doc.getBlocks()[0]
+        expect(sectionWithRole.getRole()).to.equal('foreword')
+        const sectionWithoutRole = doc.getBlocks()[1]
+        expect(sectionWithoutRole.getRole()).to.be.undefined()
+      })
+
       describe('Get authors', function () {
         it('should return an empty list when the document has no author', function () {
           const input = `= Getting Real: The Smarter, Faster, Easier Way to Build a Successful Web Application
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -2772,7 +2772,7 @@ export namespace Asciidoctor {
      *
      * @returns the role as a space-separated String.
      */
-    getRole(): string;
+    getRole(): string | undefined;
 
     /**
      * Checks if the specified role is present in the list of roles for this node.
--- a/packages/core/types/tests.ts
+++ b/packages/core/types/tests.ts
@@ -460,6 +460,18 @@ assert(html === `<div class="paragraph">
 <p><a href="https://github.com/mojavelinux">@mojavelinux</a></p>
 </div>`);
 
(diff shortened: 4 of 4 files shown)


bitPogo:
Test are missing...I am not sure where to put them (under types). Also I made a output from get role and currently it returns a object, if I use a plain blocknode.

bitPogo:
Odd..I cannot reproduce the error (of the test) locally, but I get another one: 

```bash
 1) Node.js
       Node.js
         Loading
           should get the line of number of a block when sourcemap is enabled:
     AssertionError: expected undefined to equal 3
      at Context.<anonymous> (spec/share/asciidoctor-spec.js:360:48)
      at processImmediate (internal/timers.js:461:21)
```

Mogztter:
> Odd..I cannot reproduce the error (of the test) locally, but I get another one:

This error was fixed on master (https://github.com/asciidoctor/asciidoctor.js/commit/0d00235b6e473f51cf835f8cd627c4ef9675f67b) maybe your branch is a bit out of sync?

Anyway, you should test locally against a release because we are currently merging a few changes on the master branch of Asciidoctor (Ruby) that might impact Asciidoctor.js.

You can run (at the root of this project):

```
$ npm run clean --prefix packages/core
$ ASCIIDOCTOR_CORE_VERSION=v2.0.10 npm run build --prefix packages/core
```

bitPogo:
That does the trick - probably the some tingeling with a cache.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -1035,7 +1035,8 @@ AbstractNode.prototype.isRole = function (expectedValue) {
  * @memberof AbstractNode
  */
 AbstractNode.prototype.getRole = function () {
-  return this.$role()
+  const role = this.$role()
+  return Opal.nil ? undefined : role
```
Mogztter, line 1039: I think it should be:

```suggestion
  return role === Opal.nil ? undefined : role
```

a test will confirm that :wink:

Mogztter:
> That does the trick - probably the some tingeling with a cache.

We are using an aggressive cache because transpiling Asciidoctor Ruby to JavaScript (using Opal) is somewhat "slow".

packages/core/src/asciidoctor-core-api.js
bitPogo: Have you any idea where to put a test for it...I mean to test in test for types seems a bit weird...but would be do for now...maybe to restructure/split in general test.ts would be a good idea for more structure. So what are your thoughts?

Mogztter: > Have you any idea where to put a test for it...I mean to test in test for types seems a bit weird...but would be do for now...maybe to restructure/split in general test.ts would be a good idea for more structure. So what are your thoughts?

Yes, at some point, we should `tests.ts`, the file is starting to get quite long but we should do that in a dedicated pull request.

The idea is to add a test in `tests.ts` to make sure that the type definition is correct.

You can add a test in `packages/core/spec/share/asciidoctor-spec.js`.
The idea is to load a document and call `getRole` on a block that does not have a role:

_untested-code.js_
```js
const doc = asciidoctor.load(`= Title

== Section without role
`)
const sectionWithoutRole = doc.getBlocks()[0]
expect(sectionWithoutRole.getRole()).to.be.undefined()
```

All tests in "share/asciidoctor-spec.js" will be executed against all supported runtime, currently:
- Node
- Browser
- GraalVM

Mogztter:
@bitPogo Let me know if it does make sense, if not I will try my best to clarify :wink: 
The code looks good but we need to add a few tests to make sure that everything works as intended and that we won't break the code in the future.

bitPogo:
I already tinkered with some test - however I do not get the expected return value (`string`). I actually get a `object` which looks like:

```javascript
constructor {
        '$$id': 4,
        apply: [Function (anonymous)],
        call: [Function (anonymous)]
}
```

Mogztter:
This object is `Opal.nil`, did you run clean and build?

Mogztter:
@bitPogo I hope you don't mind, I've added tests and I will merge this pull request as soon as the build is green :tada:

Mogztter:
Actually I will let this pull request open a bit longer so you get a chance to review my changes :wink:

bitPogo (commented):
This looks pretty good, however, I currently have 2 project, where `getRole` returns something like I mentions in the discussion. After the release I will take a 2nd look in the projects and try to pin point the problem and report back.

Mogztter:
And merged, thanks @bitPogo :+1:

Issue #1113, opened 2020-11-04
Timings#printReport - all parameter should be optional


bitPogo:
Compare the [types](https://github.com/asciidoctor/asciidoctor.js/blob/cb0db00597ae7ecfd65cc717f3e67f8c284d5abe/packages/core/types/index.d.ts#L442) with the [actual code](https://github.com/asciidoctor/asciidoctor.js/blob/cb0db00597ae7ecfd65cc717f3e67f8c284d5abe/packages/core/src/asciidoctor-core-api.js#L3013).

Mogztter:
You're correct :+1:

PR #1128, opened 2020-11-09
📑 Fix Timings#printReport type definition

bitPogo:
This patch makes all parameters of Timings#printReport optional.

See #1113

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -439,7 +439,7 @@ export namespace Asciidoctor {
      * @param to - an optional output (by default stdout)
      * @param subject - an optional subject (usually the file name)
      */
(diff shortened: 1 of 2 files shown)


Mogztter:
The build is failure is related to a change merged on asciidoctor/asciidoctor (master): https://github.com/asciidoctor/asciidoctor/pull/3835

Asciidoctor (master) does not force the "px" unit anymore so the assertion needs to be updated. I will take care of that in a separated pull request.

Could you please call `timings.printReport(memoryLogger)` and `timings.printReport()` in `tests.ts` to confirm that everything is working as expected?

bitPogo:
Done

Mogztter:
Looks good, I've rebased on master to make sure that everything is running fine.

Mogztter (approved):
Nice work, thanks :ok_hand:

PR #1145, opened 2020-11-21
✨ resolves #1133 Map AbstractNode#role=

### LINKED ISSUE #1133: Map AbstractNode#role=
The function should be named `setRole` and accepts a single role name (`string`), a space-separated String of role names (`string`), or an Array of role names (`string[]`):

```js
node.setRole('foo')
node.setRole('foo,bar,baz')
node.setRole(['foo', 'bar', 'baz'])
```

We might decide to allow a list of values (`...names: string[]`):

```js
node.setRole('foo', 'bar', 'baz')
```

Reference: https://github.com/asciidoctor/asciidoctor/blob/40a9c2c84606ec40e27af59298f8047eef1f9c63/lib/asciidoctor/abstract_node.rb#L219-L226

Introduced in https://github.com/asciidoctor/asciidoctor/commit/40a9c2c84606ec40e27af59298f8047eef1f9c63 (v2.0.11)

Mogztter:
resolves #1133

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -2775,6 +2775,24 @@ export namespace Asciidoctor {
     getRole(): string | undefined;
 
     /**
+     * Sets the value of the role attribute on this node.
+     *
+     * @param names - A single role name, a space-separated String of role names or a list of role names
+     *
+     * @returns the value of the role attribute.
+     */
+    setRole(...names: string[]): string;
+
+    /**
+     * Sets the value of the role attribute on this node.
+     *
+     * @param names - an Array of role names
+     *
+     * @returns the value of the role attribute.
+     */
+    setRole(names: string[]): string;
+
+    /**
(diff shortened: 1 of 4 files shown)


packages/core/types/index.d.ts
```diff
@@ -2774,6 +2774,24 @@ export namespace Asciidoctor {
...
+    setRole(...names: string[]): string;
+
+    /**
+     * Sets the value of the role attribute on this node.
+     *
+     * @param names - an Array of role names
+     *
+     * @returns the value of the role attribute.
+     */
+    setRole(names: string[]): string;
```
Mogztter, line 2793: @bitPogo Is it the right to do it? I want to allow the following:

```js
block.setRole('afterword');
block.setRole('lastword', 'closing');
block.setRole(['finalword', 'conclude']);
```

I tried to declare only `setRole(...names: string[]): string;` but the following is invalid:

```js
block.setRole(['finalword', 'conclude'])
```

PR #1147, opened 2020-11-22
🏷️ Change Inline#create opts: any -> Record

### LINKED ISSUE #1116: Inline#create - opts should be object
See the [type definition](https://github.com/asciidoctor/asciidoctor.js/blob/cb0db00597ae7ecfd65cc717f3e67f8c284d5abe/packages/core/types/index.d.ts#L2251) and compare with the [actual code](https://github.com/asciidoctor/asciidoctor.js/blob/cb0db00597ae7ecfd65cc717f3e67f8c284d5abe/packages/core/src/asciidoctor-core-api.js#L2311).

bitPogo:
resolves #1116

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -2248,7 +2248,7 @@ export namespace Asciidoctor {
      * Create a new Inline element.
      * @returns a new Inline element
      */
-    static create(parent: AbstractBlock, context: string, text?: string, opts?: any): Inline;
+    static create(parent: AbstractBlock, context: string, text?: string, opts?: Record<string, unknown>): Inline;
 
     /**
      * Get the converted content for this inline node.


packages/core/types/index.d.ts
```diff
@@ -2248,7 +2248,7 @@ export namespace Asciidoctor {
      * Create a new Inline element.
      * @returns a new Inline element
      */
-    static create(parent: AbstractBlock, context: string, text?: string, opts?: any): Inline;
+    static create(parent: AbstractBlock, context: string, text?: string, opts?: Record<string, unknown>): Inline;
```
Mogztter, line 2251: :thinking: I'm trying to think if it's possible to use something other than `string` as key... but I think the answer is no.

bitPogo: Could there anything besides a `string` as value?

Mogztter: No I think it would be fine.
In Ruby, you can have numeric keys but that's not possible in JavaScript so `Record<string, unknown>` should be OK.

```rb
h = {1 => 'one', 2 => 'two', 3 => 'three'}
```

Mogztter:
Thanks again :hugs:

PR #1148, opened 2020-11-22
🐛 Fix isAttribute declaration

### LINKED ISSUE #1120: Wrong return value of AbstractNode#isAttribute
`isAttribute` of `AbstractNode` is currently type with a return value of `void` but should be boolean. See [here](https://github.com/asciidoctor/asciidoctor.js/blob/cb0db00597ae7ecfd65cc717f3e67f8c284d5abe/packages/core/types/index.d.ts#L2716).
Additionally it might be more expressive if the body of the function changes from:
```javascript
AbstractNode.prototype.isAttribute = function (name, expectedValue, fallbackName) {
  const result = this['$attr?'](name, expectedValue, fallbackName)
  return result === Opal.nil ? false : result
}
```
to:
```javascript
AbstractNode.prototype.isAttribute = function (name, expectedValue, fallbackName) {
  const result = this['$attr?'](name, expectedValue, fallbackName)
  return !(result === Opal.nil)
```

See  also [here](https://github.com/asciidoctor/asciidoctor.js/blob/cb0db00597ae7ecfd65cc717f3e67f8c284d5abe/packages/core/src/asciidoctor-core-api.js#L950-L953).

bitPogo:
resolves #1120

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -948,8 +948,7 @@ AbstractNode.prototype.hasAttribute = function (name) {
  * @memberof AbstractNode
  */
 AbstractNode.prototype.isAttribute = function (name, expectedValue, fallbackName) {
-  const result = this['$attr?'](name, expectedValue, fallbackName)
-  return result === Opal.nil ? false : result
+  return this['$attr?'](name, expectedValue, fallbackName)
 }
 
 /**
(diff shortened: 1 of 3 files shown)


Mogztter (changes requested):
LGTM, could you please add a test in `packages/core/types/tests.ts`?


```js
const options = { attributes: 'sectnums' };
const doc = asciidoctor.load('== Test', options);
assert(doc.getAttribute('sectnums') === '')
assert(doc.isAttribute('sectnums'))
assert(doc.isAttribute('sectnums', ''))
assert(doc.isAttribute('sectnums', 'not this') === false)
assert(doc.isAttribute('foo') === false)
assert(doc.isAttribute('foo', '') === false)
assert(doc.isAttribute('foo', 'bar') === false)
```

bitPogo:
Done

Mogztter:
Perfect, thank you :ok_hand:

PR #1149, opened 2020-11-22
🏷️ Fix return value definitions in AbstractNode#resolveSubs*

### LINKED ISSUE #1118: Enhance return value defintions in AbstractNode
Some return values are marked as `SOMETYPE | void` and they should be `SOMETYPE | undefined`.

bitPogo:
resolves #1118

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -2639,21 +2639,21 @@ export namespace Asciidoctor {
      *
      * @returns An Array of Strings representing the substitution operation or nothing if no subs are found.
      */
-    resolveSubstitutions(subs: string, type?: string, defaults?: string[], subject?: string): string[] | void;
+    resolveSubstitutions(subs: string, type?: string, defaults?: string[], subject?: string): string[] | undefined;
 
     /**
      * Call {@link AbstractNode#resolveSubstitutions} for the 'block' type.
      *
      * @see {@link AbstractNode#resolveSubstitutions}
      */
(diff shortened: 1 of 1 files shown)


Mogztter:
We also need to update the JavaScript API:

https://github.com/asciidoctor/asciidoctor.js/blob/aa12065fdc67d3262ff4cb710a03bc9934f73d7a/packages/core/src/asciidoctor-core-api.js#L867

`this.$resolve_subs(subs, type, defaults, subject)` can return `Opal.nil` so we need to check the return value and return `undefined` (instead of `Opal.nil`).

We should also expand the following test:

https://github.com/asciidoctor/asciidoctor.js/blob/aa12065fdc67d3262ff4cb710a03bc9934f73d7a/packages/core/spec/node/asciidoctor.spec.js#L889-L893

And add the following assertions to make sure that the return value is `undefined`:

```js
expect(block.resolveSubstitutions()).to.be.undefined()
expect(block.resolveSubstitutions([])).to.be.undefined()
expect(block.resolveSubstitutions(undefined)).to.be.undefined()
```

Mogztter (changes requested):
The TypeScript definition is correct but we need to fix the JavaScript API.
Though, it could be done in a subsequent pull request. Let me know what you prefer.

bitPogo:
I trust you completely on this, so just let me know, what you thinks fits best for the project.

Mogztter:
> I trust you completely on this, so just let me know, what you thinks fits best for the project.

I've created a follow-up issue: https://github.com/asciidoctor/asciidoctor.js/issues/1153

PR #1160, opened 2020-12-08
📝 suppress autolinks on include support matrix page

mojavelinux:


Mogztter:
Thanks, we should also probably use unicode for Emoji instead of text notation (`:ok_hand:`)

mojavelinux:
I agree. You could use an attribute so you can use the value that is appropriate / available for a given environment.

PR #1161, opened 2020-12-08
📝 Improve support matrix readability

Mogztter:
| Current | After |
| -- | -- |
|current |compat-less-monospaced

### DIFF
--- a/docs/modules/spec/pages/include-support-matrix.adoc
+++ b/docs/modules/spec/pages/include-support-matrix.adoc
@@ -20,47 +20,53 @@ This document describes how the include directive is resolved in the following e
 |Node.js
 |MRI
 
-|+path/to/file.adoc+
-|+file:///root+
+| +path/to/file.adoc+
+| +file:///root+
 |{icon-ok}
 |{icon-ok}
-|{icon-ko} +
-The resolved path is `/root/file:/root/path/to/file.adoc`.
+|{icon-ko}
+5+a|
+MRI: The resolved path is [.path]_/root/file:/root/path/to/file.adoc_.
 
 |+path/../path/to/file.adoc+
 |+file:///root+
 |{icon-ok}
 |{icon-ok}
-|{icon-ko} +
-The resolved path is `/root/file:/root/path/to/file.adoc`.
+|{icon-ko}
+5+a|
+MRI: The resolved path is [.path]_/root/file:/root/path/to/file.adoc_.
 
 |+file:///root/path/to/file.adoc+
 |
 |{icon-ok}
 |{icon-ok}
-|{icon-ko} +
-include uri not readable: `\file:///root/path/to/file.adoc`.
+|{icon-ko}
+5+a|
+MRI: include uri not readable: [.path]_\file:///root/path/to/file.adoc_.
 
 |+file:///root/path/../path/to/file.adoc+
 |
 |{icon-ok}
 |{icon-ok}
-|{icon-ko} +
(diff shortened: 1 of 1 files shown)


mojavelinux:
That's definitely a much tighter layout.

You are missing two escapes for raw.com.

I don't think the column should be named MRI. It should just be named Ruby.

mojavelinux:
You only need the AsciiDoc table cell if the cell has blocks. So I think the colspan rows can just be default table cells. If you need a hard line break, I would use the trailing + instead.

mojavelinux:
I wonder if the title of the first column should be File / Message. Then it's clear what the colspan rows are about.

Mogztter:
> You only need the AsciiDoc table cell if the cell has blocks. So I think the colspan rows can just be default table cells. If you need a hard line break, I would use the trailing + instead.

True, but the first cell is using a monospaced style so I'm using the inline AsciiDoc cell style to override it.
Alternatively, I could remove the monospaced style and apply on every path/file... :thinking: 

> You are missing two escapes for raw.com.

Can't see it :eyes: 

> I don't think the column should be named MRI. It should just be named Ruby.

:+1: 

> I wonder if the title of the first column should be File / Message. Then it's clear what the colspan rows are about.

:+1:

mojavelinux:
> True, but the first cell is using a monospaced style so I'm using the inline AsciiDoc cell style to override it.

You can use the "d" style to return to the default/normal cell style.

mojavelinux:
> > You are missing two escapes for raw.com.

> Can't see it

Nevermind, it must just be GitHub messing up. I don't see it either.

mojavelinux:
I have one other idea for this table. The messages line still seems hard to map visually. One idea would be to use the first column (or two) for the header, then put the messages across the remaining columns.

```
| File              | Base dir     | Browser | Node.js | Ruby
| path/to/file.adoc | file:///root | 👌      | 👌       | ❌
                          Messages | Ruby: The resolved path is /root/file:/root/path/to/file.adoc.
```

(Disclaimer: I am not a table specialist, so this may be off base)

Mogztter:
> I have one other idea for this table. The messages line still seems hard to map visually. One idea would be to use the first column (or two) for the header, then put the messages across the remaining columns.


What about a collapsible block?

```
.Messages
[%collapsible]
====
Ruby: The resolved path is [.path]_/root/file:/root/path/to/file.adoc_.
====
```

mojavelinux:
Perhaps, though I'd like to see the simpler approach first. I would also include that row even if there are no messages. Just write "Not applicable". That way, the table looks uniform and we can style the rows using even/odd highlighting if necessary.

Mogztter:
> I would also include that row even if there are no messages. Just write "Not applicable"

That's a good idea :+1: 

> One idea would be to use the first column (or two) for the header, then put the messages across the remaining columns.

I tried but it does not look great. Also I'm trying to avoid using bold/strong in the cells otherwise the table header does not stand out visually anymore.

mojavelinux:
How did you do the expand? Is that extra JavaScript?

Mogztter:
> How did you do the expand? Is that extra JavaScript?

Yes! As requested, I added a unique id and the table structure is uniform so we can use even/odd and target the next row to show/hide details using JavaScript. Of course, styling should be done using CSS.

Mogztter:
I will merge this pull request as it already greatly improves readability even if we don't add some JavaScript to hide/show the reason.

PR #1175, opened 2021-01-07
use gender-neutral pronoun to refer to Asciidoctor.js [skip ci]

mojavelinux:


Mogztter:
Thanks!
I'm still experimenting with textlint but my goal is to "re-implement" a few rules created by GitLab for Vale: https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/.vale/gitlab/InclusionGender.yml

As you know, the textlint AsciiDoc parser is still experimental but using it on an actual (rather large) documentation site will hopefully foster development!

PR #1256, opened 2021-04-05
🗃️ Change Document#getConverter return value

### LINKED ISSUE #1122: Document#getConverter should be typed with a return value of AbstractConverter
The current type of `getConverter` in `AbstractNode` is object, but should be `AbstractConverter`.

See also [here](https://github.com/asciidoctor/asciidoctor.js/blob/cb0db00597ae7ecfd65cc717f3e67f8c284d5abe/packages/core/types/index.d.ts#L1944).

bitPogo:
Document#getConverter return value is now marked as AbstractConverter.
This resolves #1122.

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -1958,7 +1958,7 @@ export namespace Asciidoctor {
     /**
      * Get the {Converter} instance being used to convert the current {Document}.
      */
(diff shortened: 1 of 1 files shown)


Mogztter:
Thanks @bitPogo, I think it's a reasonable change 👌🏻

PR #1265, opened 2021-04-11
🐛 Implement unpack 'C3' with charCodeAt

Mogztter:


packages/core/lib/asciidoctor/js/opal_ext/string.rb
```diff
@@ -11,7 +11,7 @@ def limit_bytesize size
 
   def unpack format
     if format == 'C3'
-      self[0, 3].bytes.select.with_index {|_, i| i.even? }
+      `self.substring(0, 3).split('').map((c) => c.charCodeAt(0))`
```
mojavelinux, line 14: This could be more consise using substr and dropping the optional argument to charCodeAt (which defaults to 0). Also, I updated it to use an anonymous function be consistent with the code in this branch.

```suggestion
      `self.substr(0, 3).split('').map(function (c) { return c.charCodeAt() })`
```

mojavelinux (changes requested):

packages/core/lib/asciidoctor/js/opal_ext/string.rb
mojavelinux: ..and perhaps it should be codePointAt instead of charCodeAt? See https://stackoverflow.com/questions/36527642/difference-between-codepointat-and-charcodeat

mojavelinux: It seems like splitting a string on empty string can cut characters in half. This is the safer approach:

```js
`[self.codePointAt(0), self.codePointAt(1), self.codePointAt(2)].filter((function (it) { return it != null }))`
```

or written longhand as:

```js
`
  var bytes = []
  var idx = -1
  var b
  while (++idx < 3 && (b = str.codePointAt(idx)) != null) bytes.push(b)
  bytes
`
```

Mogztter: > and perhaps it should be codePointAt instead of charCodeAt

In theory, maybe... but this code is only used to determine if a BOM is present: https://github.com/asciidoctor/asciidoctor/blob/ac3391dba2ff630760a67b945ff93aa1876004d9/lib/asciidoctor/helpers.rb#L98-L107

More specifically, we check if the 2 first code points are equals to `[0xff, 0xfe]` (`BOM_BYTES_UTF_16LE`) or `[0xfe, 0xff]` (`BOM_BYTES_UTF_16BE`).

The generated code (by Opal) is:

```js
Opal.const_set($nesting[0], 'BOM_BYTES_UTF_16LE', [255, 254]);
Opal.const_set($nesting[0], 'BOM_BYTES_UTF_16BE', [254, 255]);
```

```js
'\xFE\xFF'.codePointAt(0) //254
'\xFE\xFF'.codePointAt(1) // 255

'\xFE\xFF'.charCodeAt(0) // 254
'\xFE\xFF'.charCodeAt(1) // 255
```

> It seems like splitting a string on empty string can cut characters in half. This is the safer approach:

Is it really an issue in this context?
Did you find a case where Asciidoctor.js incorrectly detects a BOM? I would like to add a test case before making this change.

mojavelinux: Since we're overriding unpack, it just seems like a good idea to make it as accurate as possible. Here's an example of where they are different:

```
'𝟘'.codePointAt(0)
# => 120792
'𝟘'.charCodeAt(0)
# >= 55349
```

mojavelinux: Though I guess it doesn't matter since neither one yields the same result as Ruby. So this is uncharted territory anyway.

mojavelinux: So perhaps just this would do:

```js
`
var bytes = []
var idx = -1
var b
while (++idx < 3 && (b = str.charCodeAt(idx)) != null) bytes.push(b)
bytes
`
```

Mogztter: Indeed, here's the result in Ruby:

```rb
p "𝟘".unpack "C4" # [240, 157, 159, 152]
```

On Node.js, it's relatively easy to implement using `Buffer` and `Uint8Array`:

```js
new Uint8Array(Buffer.from('𝟘', 'utf8'))
// Uint8Array(4) [ 240, 157, 159, 152 ]
```

Here's a polyfill of Buffer: https://github.com/feross/buffer/blob/795bbb5bda1b39f1370ebd784bea6107b087e3a7/index.js#L1956-L2034

It's interesting to note that this polyfill is using `charCodeAt`.

> Though I guess it doesn't matter since neither one yields the same result as Ruby. So this is uncharted territory anyway.

The module `unpack` is not required by default because it's quite large but I believe that the Opal implementation is accurate.


> So perhaps just this would do:

Ruby returns an Array with `nil`:

```rb
"a".unpack "C3" # [97, nil, nil]
```

mojavelinux:
Aha! I forgot that Ruby left the nil entries. Then just calling charCodeAt
3 times for the purpose of this polyfill will do. It's been interesting to
follow this analysis, but as you pointed out, we're only using it for BOM
detection. So I now realize it's best to keep it simple and focused on the
task at hand.

mojavelinux (approved):

PR #1268, opened 2021-04-13
🐛 Fix bytes issues in Opal

Mogztter:


mojavelinux:
I think that's a great idea!

PR #1291, opened 2021-05-08
📝 Fix GraalVM example based on test

abelsromero:
The example does not set permission, wich makes it fail.

```java
      .allowAllAccess(true)
      .allowPolyglotAccess(PolyglotAccess.ALL)
```

Also split the example into a file under `examples` for convenience.

### DIFF
--- a/docs/modules/setup/examples/AsciidoctorGraalVM.java
+++ b/docs/modules/setup/examples/AsciidoctorGraalVM.java
@@ -1,0 +1,52 @@
+import org.graalvm.polyglot.Context;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


Mogztter:
LGTM!

Maybe we should add the corresponding GraalVM version (on which the code example works)? I remember a breaking change a few versions ago so it would be useful to state that the code example was tested against GraalVM version x.y.z. What do you think?

abelsromero:
I don't know enough to say. I am using `21.1.0.r16`, on one hand displaying the "r*" makes no sense because they release patches quite often and are not relevant, but on the other I don't know how reliable is to say "21.1.0", "21.1.x" or "21.x.x".

Maybe just  mention this is experimental? I mean I just confirmed that only String-to-String conversion works.

Mogztter:
> but on the other I don't know how reliable is to say "21.1.0", "21.1.x" or "21.x.x".

We should just say that this was "tested" on version 21.1.0. It might work on 21.x or even on 22.x but no guarantee. 

> Maybe just mention this is experimental? I mean I just confirmed that only String-to-String conversion works.

Yes we can also state that this is experimental 👍🏻

PR #1304, opened 2021-05-23
resolves #1297 pass block argument to Logger#add

### LINKED ISSUE #1297: initializeClass helper does not work with methods that accept a block
The way the initializeClass helper registers methods, it does not provide a way to access or yield to the block argument. This, in turn, impacts a custom logger registered using the LoggerManager.newLogger helper. The Logger#add method can accept the message as a block argument. And, indeed, some logger calls in Asciidoctor core pass the message as a block. In these cases, the custom logger can only see an undefined message.

Here's a possible solution to the problem that emulates how Opal handles blocks. In the initializeClass helper, first store a reference to the function passed to Opal.def:

```js
var $function
Opal.def(scope, '$' + functionName, $function = function () {
```

Then, before dispatching to userFunction, detect and append the block as the final argument:

```js
var $block = $function.$$p
if ($block) {
  args[args.length] = function () { return Opal.yield1($block) }
  args.length += 1
  $function.$$p = null
}
```

This assumes the block itself doesn't take any arguments. Ideally, the generated function would account for block arguments. And there may be a better way of doing this altogether. But, at least when registering a custom logger using newLogger, the block argument to Logger#add should be honored.

Mogztter:
resolves #1297

### DIFF
--- a/packages/core/spec/node/asciidoctor.spec.js
+++ b/packages/core/spec/node/asciidoctor.spec.js
@@ -428,6 +428,35 @@ intro
         expect(errorMessage.message).to.equal('hi')
         expect(errorMessage.progname).to.equal('asciidoctor.js')
       })
+      it('should accept a message as block argument', () => {
+        const messages = []
+        const memoryLogger = asciidoctor.LoggerManager.newLogger('CustomMemoryLogger', {
+          add: function (severity, message, programName, block) {
+            if (typeof block === 'function') {
+              messages.push(block())
+            } else {
+              messages.push(message)
+            }
+          }
+        })
+        const defaultLogger = asciidoctor.LoggerManager.getLogger()
+        try {
+          asciidoctor.LoggerManager.setLogger(memoryLogger)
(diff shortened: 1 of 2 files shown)


mojavelinux (approved):
This looks reasonable to me. There may come a time when we need to support a block that accepts arguments. Though for the cases we have thus far, the basic yield is sufficient.

packages/core/spec/node/asciidoctor.spec.js
```diff
@@ -428,6 +428,33 @@ intro
...
+        const defaultLogger = asciidoctor.LoggerManager.getLogger()
+        try {
+          asciidoctor.LoggerManager.setLogger(memoryLogger)
+          memoryLogger.add('error', 'test', 'asciidoctor.js')
+          asciidoctor.convert('Hello, {name}!', {
+            attributes: {
+              'attribute-missing': 'drop-line'
+            }
+          })
+          expect(messages.length).to.equal(2)
```
Mogztter, line 451: I will also add a message after `convert` to make sure that the block is correctly removed.

PR #1322, opened 2021-06-06
resolves #1321 invoke LoggerManager#setLogger setter

### LINKED ISSUE #1321: LoggerManager.setLogger needs to invoke logger setter
The memoization in the LoggerManager requires that the setter method for the logger property be invoked for proper operation when assigning a new logger. This setter method is responsible for instantiating the specified logger class when the logger value is falsy. See https://github.com/asciidoctor/asciidoctor/blob/3e39292ce3fab18b35cd0066c853681b23150bb5/lib/asciidoctor/logging.rb#L93-L95

Here's a problematic example:

```js
LoggerManager.getLogger()
LoggerManager.setLogger(null)
LoggerManager.getLogger()
```

The third call will return Opal.nil instead of a new instance of the logger.

Whereas this will work:

```
LoggerManager.getLogger()
LoggerManager['$logger='](null)
LoggerManager.getLogger()
```

Mogztter:
resolves #1321

### DIFF
--- a/packages/core/spec/node/asciidoctor.spec.js
+++ b/packages/core/spec/node/asciidoctor.spec.js
@@ -970,6 +970,18 @@ image::https://asciidoctor.org/images/octocat.jpg[GitHub mascot]`
       }
     })
 
+    it('should instantiate the specified logger class when the logger value is falsy', () => {
+      const defaultLogger = asciidoctor.LoggerManager.getLogger()
+      try {
+        asciidoctor.LoggerManager.setLogger(null)
+        const initialLogger = asciidoctor.LoggerManager.getLogger()
+        expect(initialLogger).to.be.instanceof(Object)
+        expect(initialLogger.progname).to.equal('asciidoctor')
+      } finally {
+        asciidoctor.LoggerManager.setLogger(defaultLogger)
+      }
+    })
+
     it('should be able to apply specific inline substitutions to text', () => {
       const doc = asciidoctor.load('para', { attributes: { start: 'start', finish: 'finish' } })
       const para = doc.getBlocks()[0]
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -2762,7 +2762,7 @@ LoggerManager.getLogger = function () {
  * @memberof LoggerManager
  */
 LoggerManager.setLogger = function (logger) {
-  this.logger = logger
+  this['$logger='](logger)
 }
 
 /**


packages/core/spec/node/asciidoctor.spec.js
```diff
@@ -970,6 +970,18 @@ image::https://asciidoctor.org/images/octocat.jpg[GitHub mascot]`
       }
     })
 
+    it('should instantiate the specified logger class when the logger value is falsy', () => {
+      const defaultLogger = asciidoctor.LoggerManager.getLogger()
+      try {
+        const initialLogger = asciidoctor.LoggerManager.setLogger(null)
+        console.log(initialLogger)
```
mojavelinux, line 977: You left behind a console.log statement.

packages/core/spec/node/asciidoctor.spec.js
```diff
@@ -970,6 +970,18 @@ image::https://asciidoctor.org/images/octocat.jpg[GitHub mascot]`
       }
     })
 
+    it('should instantiate the specified logger class when the logger value is falsy', () => {
+      const defaultLogger = asciidoctor.LoggerManager.getLogger()
+      try {
+        const initialLogger = asciidoctor.LoggerManager.setLogger(null)
```
mojavelinux, line 976: I think you should treat the setLogger as void and follow with a separate call to `getLogger()`. In Ruby, it also takes two calls.

mojavelinux (changes requested):

packages/core/spec/node/asciidoctor.spec.js
Mogztter: Technically, `LoggerManager#logger=` Ruby method returns the value right?

https://github.com/asciidoctor/asciidoctor/blob/f580230151bdb8392dcc55d6e1d46bb57d858cf7/lib/asciidoctor/logging.rb#L93-L95

But I agree, it's a bit weird to get a value from a setter function.

mojavelinux: Setter methods always return nil in Ruby...even if you specify a return value.

Mogztter: Interesting, it seems that Opal does things differently since `this['$logger=']` returns a value 🤔

mojavelinux: Actually, I just remembered. It's even weirder than that. Ruby always returns the value passed to the setter. In other words, it passes it through to act like any other assignment. Thus, it disregards any activity in the method itself.

PR #1345, opened 2021-07-03
🐛 resolves #1344 look for alternate char code for BOM and return standard BOM

### LINKED ISSUE #1344: UTF-8 BOM is not removed when running on Node.js and input is a file
Although there's a test that covers removing the UTF-8 BOM from the AsciiDoc source, this only works on Node.js when the input is a string created in JavaScript. When the AsciiDoc source is read from a file, or the string is created from Buffer.from, the UTF-8 BOM is not removed.

This snippet explains why we're hitting this problem:

```js
const fs = require('fs')

fs.writeFileSync('test.adoc', Buffer.concat([Buffer.from([0xEF, 0xBB, 0xBF]), Buffer.from('= Document Title')]))

const contentsFromFile = fs.readFileSync('test.adoc')
console.log(contentsFromFile.toString().charCodeAt())
console.log(contentsFromFile.toString().charCodeAt(1))
// => 65279
// => '='

const contentsFromString = '\xef\xbb\xbf= Document Title'
console.log(contentsFromString.charCodeAt())
console.log(contentsFromString.charCodeAt(3))
// => 239
// => '='
```

It appears that when Node.js creates a string by way of a Buffer, such as when reading the contents of a file, it changes the UTF-8 BOM into a different BOM character (code: 65279, char ref: 0xFEFF). I have not found any way to disable this behavior. It's basically a quirk of Node.js.

I think Asciidoctor.js should detect this alternate BOM and remove the character. (I'm open to changing Asciidoctor Ruby, if we determine it's necessary).

mojavelinux:
resolves #1344

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,6 +8,10 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 == Unreleased
 
+Bug Fixes::
+
+* Strip alternate BOM that uses char code 65279 when input passes through a Buffer (#1344)
+
 Infrastructure::
 
 * Upgrade GraalVM to 20.1.0 (#1035)
--- a/packages/core/lib/asciidoctor/js/opal_ext/string.rb
+++ b/packages/core/lib/asciidoctor/js/opal_ext/string.rb
@@ -7,11 +7,21 @@ class String
   end unless method_defined? :limit_bytesize
   alias :limit :limit_bytesize unless method_defined? :limit
 
+  alias :_original_byteslice :byteslice
+
+  def byteslice index, length = 1
+    if index == 3 && length >= index && `#{self}.charCodeAt() === 65279`
+      `#{self}.substr(1)`.byteslice 0, length - 3
+    else
+      _original_byteslice index, length
+    end
+  end
+
   alias :_original_unpack :unpack
 
   def unpack format
     if format == 'C3'
-      self[0, 3].bytes.select.with_index {|_, i| i.even? }
+      `#{self}.charCodeAt() === 65279` ? [239, 187, 191] : self[0, 3].bytes.select.with_index {|_, i| i.even? }
     else
       _original_unpack format
     end
--- a/packages/core/spec/node/asciidoctor.spec.js
+++ b/packages/core/spec/node/asciidoctor.spec.js
@@ -566,6 +566,15 @@ intro
       const doc = asciidoctor.load('== Test')
       expect(doc.getBaseDir()).to.equal(process.cwd().replace(/\\/g, '/'))
     })
+
+    it('should load source with BOM from Buffer', function () {
+      const source = Buffer.concat([Buffer.from([0xEF, 0xBB, 0xBF]), Buffer.from('= Document Title\n:lang: fr\n\ncontent is in {lang}')]).toString()
+      const opts = { safe: 'safe', base_dir: testOptions.baseDir }
+      const doc = asciidoctor.load(source, opts)
+      expect(doc.getAttribute('lang')).to.equal('fr')
+      const html = doc.convert()
+      expect(html).to.include('content is in fr')
+    })
   })
 
   describe('Loading file', () => {
--- a/packages/core/spec/share/asciidoctor-spec.js
+++ b/packages/core/spec/share/asciidoctor-spec.js
@@ -623,9 +623,10 @@ This is another paragraph.
         expect(appendixSection.getCaption()).to.equal('Appx A: ')
       })
 
-      it('should load a file with bom', function () {
+      it('should load source with BOM from String', function () {
+        const source = '\xef\xbb\xbf= Document Title\n:lang: fr\n:fixtures-dir: spec/fixtures\n\ncontent is in {lang}\n\ninclude::{fixtures-dir}/include.adoc[]'
         const opts = { safe: 'safe', base_dir: testOptions.baseDir }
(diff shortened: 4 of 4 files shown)


mojavelinux:
It's possible this test will fail. I can't figure out why the test isn't using the updated code. But when I run it on a test document, it works.

mojavelinux:
Yep. As expected. Can you think of a reason why the test is not hitting the updated code?

Mogztter:
> Yep. As expected. Can you think of a reason why the test is not hitting the updated code?

Are you saying that `packages/core/lib/asciidoctor/js/opal_ext/string.rb` is somehow not used? That would be very surprising...

mojavelinux:
> Are you saying that packages/core/lib/asciidoctor/js/opal_ext/string.rb is somehow not used?

In the test environment it doesn't seem to be used. At least, I don't have any other explanation.

Mogztter:
> But when I run it on a test document, it works.
> In the test environment it doesn't seem to be used. At least, I don't have any other explanation.

Not sure how you are testing but I can confirm that our modified version of `String#unpack` is used.
The method returns `[239, 187, 191]`.
In the code below, `leading_bytes` is equals to `[239, 187, 191]` which is equal to `BOM_BYTES_UTF_8`.

```rb
    elsif leading_bytes == BOM_BYTES_UTF_8
      data = data.byteslice 3, data.bytesize
      data = data.encode UTF_8 unless data.encoding == UTF_8
```
Reference: https://github.com/asciidoctor/asciidoctor/blob/6220026962b40cee56046ae11b8ebe07dae544c9/lib/asciidoctor/helpers.rb#L103-L105

The culprit is `byteslice` which removes the leading `=`, instead of just the first 3 bytes `0xEF`, `0xBB` and `0xBF` 😞 

In Opal codebase, `String#byteslice` is aliased to `String#[]` and the implementation is basically: `self.substr(index, length)`: 

https://github.com/opal/opal/blob/14c62d1827dcbadaf5ac1af3e9a2871a3c88834b/opal/corelib/string.rb#L248


Instead, we probably need to do `self.$bytes().slice(index, length).$pack('c*')`

Mogztter:
> Instead, we probably need to do self.$bytes().slice(index, length).$pack('c*')

But `$bytes` implementation is incorrect in Opal 1.0.0 so we need to upgrade to Opal 1.1.1 😵

mojavelinux:
> Not sure how you are testing but I can confirm that our modified version of String#unpack is used.

Thanks for confirming. I'll keep digging to see what's going on.

I was to clarify that the existing case *does work*. We are adding a new case that checks for the BOM in a different way. And I'm confident that code does work. Where I'm stuck is trying to hook the test up to that code.

Mogztter:
> I want to clarify that the existing case does work. 

I think we are lucky that it works with `\xef\xbb\xbf` because in this case using `substr` is somehow correct (and will remove the 3 first characters corresponding to the BOM).

When using a buffer with bytes, we actually have a single character ([U+FEFF](https://www.fileformat.info/info/unicode/char/feff/index.htm)) which represents the BOM so when calling `String#byteslice` it will actually remove the 3 first characters (not bytes): `\uFEFF =`.

Does it make sense?

mojavelinux:
Yes, that makes sense. I'm quite sure I understand why this is happening. Strings in JavaScript live in both the UTF-8 and UTF-16 world and can represent characters using either encoding. And that's what is happening here. Again, I know how to solve it. I just was confused why it wasn't running my code in the test environment. Once I get that worked out, I'm sure this is going to pass.

mojavelinux:
Now I understand what you were trying to point out. You are correct, we need to also patch String#byteslice to cover the same scenario. I had confused myself since I was using a modified version of Asciidoctor core. When I cleaned everything and started over, things started working.

As to why I was having difficulty getting tests to use the updated code, I needed to discover npm run clean:core and npm run clean:patch. Very useful.

mojavelinux:
> we actually have a single character (U+FEFF) which represents the BOM

This is the really strange part about Node. That character is actually a UTF-16 character code, and when it is URI encoded becomes `%EF%BB%BF`. This is one of those cases where JavaScript can hold both UTF-16 data and UTF-8 data at the same time. One the one hand, it's like a super power. On the other hand, a pain in our side.

mojavelinux:
> This is one of those cases where JavaScript can hold both UTF-16 data and UTF-8 data at the same time. 

I guess it's more that it always holds UTF-16 data. We just don't expect it to be data which is *only* UTF-16. And that's what trips us up here.

mojavelinux:
Green bar!

mojavelinux:
@Mogztter Can you check if this branch works with the latest Opal? https://github.com/mojavelinux/asciidoctor.js/tree/bom-detection-redux Since we know why we're calling these methods, it seems reasonable to go as low level as possible.

Mogztter:
@mojavelinux With the latest version of Opal, I think we don't even need to override the `byteslice` (not 100% about `unpack`). The remaining issue is that the current fix does not work on the v2.2.x branch so we cannot fix the BOM issue on Asciidoctor.js 2.2.x.

Mogztter:
> With the latest version of Opal, I think we don't even need to override the byteslice (not 100% about unpack).

I was wrong, your changes are needed when building against [Opal 1.2.0 (master)](https://github.com/opal/opal/commit/7bef3e055ae5bf934b42992e1b49900cb2ec5f01) 😅

mojavelinux:
Excellent!

Like I said, while we could maybe find a way to use byteslice, I think we should just stay lower level and address the use case we are overriding for directly.

PR #1347, opened 2021-07-04
🐛 look for alternate char code for BOM and return standard BOM

Mogztter:
cherry-pick #1345


Mogztter:
@mojavelinux interestingly, "should load source with BOM from String" does not work anymore on the v2.2.x branch but the new test case "should load source with BOM from Buffer" is working 😞

mojavelinux:
I think I may have a better way to do both. I'll going to do some testing.

PR #1358, opened 2021-07-10
💥 resolves #1106 publish the Node module as an ES6 module

### LINKED ISSUE #1106: Publish the Node module as an ES6 module (instead of a CommonJS module)
Node.js now supports ECMAScript modules and it's now possible to publish Node package as ES6 module: https://nodejs.org/api/esm.html
Currently, the Node package is published as a CommonJS module.

Mogztter:
resolves #1106


Panakotta00:
Any progress on this?

I'm not quite sure and still investigating, but this seems to solve the issue of asciidoctor.js not beeing able to be used with svelte-kit.
Svelte-kit uses Vite to convert the commonjs module published on npmjs to a ES6 module. Tho it seems that in there is some kind of issue that is unable to handle strict mode, causing a runtime-error by using it.
I'm fairly new to this, but that what I think happens.

Steps to reproduce:
- Setup basic latest svelt-kit project
- "pnpm i @asciidoctor/core"
- in any component add in the `<script context="module">` following line: `import Asciidoctor from '@asciidoctor/core';`
- try to use the component

Expected behaviour: Component renders normally
Actual bahviour: Component errors with `TypeError: Cannot assign to read only property 'length' of function [...]`

Mogztter:
Until recently, Opal was not strict compliant. That's the reason why you are seeing this exception `TypeError: Cannot assign to read only property 'length' of function [...]`.

I believe that Opal 1.3 is now strict compliant but there are a few issues that we need to fix. Also, Asciidoctor.js 2.x is built on Opal 0.11 so this is a major upgrade and we need to be careful.

PR #1359, opened 2021-07-10
💥 Import Asciidoctor Opal as a CommonJS

Mogztter:
`asciidoctor-opal-runtime` is now available as both a CommonJS and ES6 module.
This is the first step to publish the Node package `@asciidoctor/core` as an ES6 module.

### DIFF
--- a/packages/core/lib/asciidoctor/js/opal_ext/string.rb
+++ b/packages/core/lib/asciidoctor/js/opal_ext/string.rb
@@ -21,7 +21,26 @@ class String
 
   def unpack format
     if format == 'C3'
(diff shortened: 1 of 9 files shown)


mojavelinux:
I wonder if this asciidoctor-opal-runtime should be published as @asciidoctor/opal-runtime. wdyt? Should I file an issue in that project?

Mogztter:
> I wonder if this asciidoctor-opal-runtime should be published as @asciidoctor/opal-runtime. wdyt? Should I file an issue in that project?

Yes please 👍🏻

PR #1365, opened 2021-07-13
📝 Add a link in the menu on the JSDoc

ggrossetie:


docs/modules/processor/nav.adoc
```diff
@@ -4,3 +4,4 @@
 ** xref:extract-api.adoc[Extract information from a document]
 ** xref:manipulate-api.adoc[Manipulate a document]
 ** xref:logging-api.adoc[Logging]
+* https://asciidoctor.github.io/asciidoctor.js/2.2.3/[JSDoc]
```
mojavelinux, line 7: You might want to think about making the version an attribute.

Mogztter: I wasn't sure whether we could use AsciiDoc attributes in `nav.adoc`

mojavelinux: Yep, they work there.

Issue #1369, opened 2021-07-21
How to make 'include::' work?


saneef:
I'm using asciidoctor.js to convert some Asciidoc documents. Including code files is a crucial need for the project. This is the script I'm using. _But, the `include::` doesn't seem to be working. The output only has `link::` or `<a>`._

test.js
```js
const asciidoctor = require("asciidoctor")();

const html = asciidoctor.convertFile("./a.adoc", {
	to_file: false,
	standalone: true,
});

console.log(html);
```

a.doc
```asciidoc
= A document

[source,ruby]
----
include::./c.rb[]
----

include::./b.adoc[]
```

b.doc
```asciidoc
= B Document

A line of text.

```

c.rb
```rb
print "hello world";
```

And the output I'm getting:
```html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="UTF-8">
<meta http-equiv="X-UA-Compatible" content="IE=edge">
<meta name="viewport" content="width=device-width, initial-scale=1.0">
<meta name="generator" content="Asciidoctor 2.0.15">
<title>A document</title>
<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,300italic,400,400italic,600,600italic%7CNoto+Serif:400,400italic,700,700italic%7CDroid+Sans+Mono:400,700">
<link rel="stylesheet" href="./asciidoctor.css">
</head>
<body class="article">
<div id="header">
<h1>A document</h1>
</div>
<div id="content">
<div class="listingblock">
<div class="content">
<pre class="highlight"><code class="language-ruby" data-lang="ruby">link:./c.ruby[]</code></pre>
</div>
</div>
<div class="paragraph">
<p><a href="./b.adoc" class="bare">./b.adoc</a></p>
</div>
</div>
<div id="footer">
<div id="footer-text">
Last updated 2021-07-21 17:36:39 +0550
</div>
</div>
</body>
</html>
```

saneef:
Made it working! Figured that `safe: 'unsafe'` should be set in asciidoctor options to make `include::` work.

saneef:
It would be great if the dependency of `include` with `safe` is noted in the [converter options](https://docs.asciidoctor.org/asciidoctor.js/latest/processor/convert-options/) documentation. I figured it out from [CLI options](https://docs.asciidoctor.org/asciidoctor.js/latest/cli/options/).

Mogztter:
That's a good suggestion!
[...]

PR #1372, opened 2021-07-23
📝 Draw a connection between safe mode and include

Mogztter:
Mention that safe mode can disable potentially dangerous macros such as `include::[]`

Ref #1369

//cc @saneef

### DIFF
--- a/docs/modules/processor/pages/convert-options.adoc
+++ b/docs/modules/processor/pages/convert-options.adoc
@@ -54,7 +54,8 @@ Instead of providing a JavaScript function containing extensions to register, th
 |_boolean_
 
 |safe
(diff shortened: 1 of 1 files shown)


mojavelinux:
It might be good to link to https://docs.asciidoctor.org/asciidoctor/latest/safe-modes/, which explains these modes in detail.

Mogztter:
That's a good idea!
I also added a link to the safe modes page in: https://docs.asciidoctor.org/asciidoctor.js/latest/cli/options/ (related commit https://github.com/asciidoctor/asciidoctor-cli.js/commit/b835d4753e33d5fa9a0bc91170731a02566030f8)

PR #1413, opened 2021-09-04
📝 Fix two 404ing links to the GraalVM docs

StarfallProjects:
Spotted a couple of 404s in the JS installation guide. Suggested fixes:

https://www.graalvm.org/docs/reference-manual/languages/js/ -> https://www.graalvm.org/reference-manual/js/
https://www.graalvm.org/docs/graalvm-as-a-platform/embed/ -> https://www.graalvm.org/reference-manual/embed-languages/ (was unsure about this one, also considered https://www.graalvm.org/reference-manual/polyglot-programming/ and https://www.graalvm.org/reference-manual/js/JavaInteroperability/)

### DIFF
--- a/docs/modules/setup/pages/install.adoc
+++ b/docs/modules/setup/pages/install.adoc
@@ -71,7 +71,7 @@ const asciidoctor = Asciidoctor()
 
 === Node.js application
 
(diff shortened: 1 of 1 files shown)


Mogztter:
Thanks @StarfallProjects!

> was unsure about this one, also considered...

I think https://graalvm.org/reference-manual/embed-languages is good 👍🏻

PR #1414, opened 2021-09-04
add list of valid context names to findBy docs

StarfallProjects:
Proposal: add a list of valid context names to the findBy docs. I've added all the names I've found so far. However, if this docs repo PR is accepted (https://github.com/asciidoctor/asciidoc-docs/pull/90), it might be nicer to just add a link to the docs, rather than maintain the list in two places.

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -684,6 +684,25 @@ AbstractBlock.prototype.convert = function () {
  * that match the specified selector (context, style, id, and/or role).
  * If a function block is given, it's used as an additional filter.
  * If no selector or function block is supplied, all block-level nodes in the tree are returned.
+ * Valid context names include: 
+ * document
+ * section
+ * paragraph
+ * image
+ * dlist (description list)
+ * ulist
+ * olist
+ * list_item
+ * literal
+ * listing
+ * admonition
+ * sidebar
+ * example
+ * open
+ * pass (a passthrough block)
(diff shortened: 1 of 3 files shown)


packages/core/src/asciidoctor-core-api.js
```diff
@@ -550,6 +550,25 @@ AbstractBlock.prototype.convert = function () {
  * that match the specified selector (context, style, id, and/or role).
  * If a function block is given, it's used as an additional filter.
  * If no selector or function block is supplied, all block-level nodes in the tree are returned.
+ * Valid context names include:
```
Mogztter, line 553: Could you please update the list with https://docs.asciidoctor.org/asciidoc/latest/blocks/#summary-of-built-in-contexts
And could you also please add a link to this documentation page?

Mogztter:
@StarfallProjects I left a comment and we should also update the documentation of `findBy` for the TypeScript types definition in: https://github.com/asciidoctor/asciidoctor.js/blob/041e98e180683d5d4b1e7cbf35b966bcbe1c9def/packages/core/types/index.d.ts#L2132

ggrossetie (changes requested):
@StarfallProjects I left a comment and we should also update the documentation of `findBy` for the TypeScript types definition in: https://github.com/asciidoctor/asciidoctor.js/blob/041e98e180683d5d4b1e7cbf35b966bcbe1c9def/packages/core/types/index.d.ts#L2132

PR #1489, opened 2021-12-02
Extensions parseContent should call toHash on attrs.

### LINKED ISSUE #1036: Extensions parseContent should call toHash on attrs.
current code:
```js
/**
 * Parses blocks in the content and attaches the block to the parent.
 * @param {AbstractBlock} parent - the parent block
 * @param {string|Array<string>} content - the content
 * @param {Object|undefined} attrs - an object of attributes
 * @returns {AbstractNode} - The parent node into which the blocks are parsed.
 * @memberof Extensions/Processor
 */
Processor.prototype.parseContent = function (parent, content, attrs) {
  return this.$parse_content(parent, content, attrs)
}
```

Should be:
```js
  return this.$parse_content(parent, content, toHash(attrs))
```

OskarsPakers:
Fixes #1036

### DIFF
--- a/packages/core/src/asciidoctor-extensions-api.js
+++ b/packages/core/src/asciidoctor-extensions-api.js
@@ -703,7 +703,7 @@ Processor.prototype.createInline = function (parent, context, text, opts) {
  * @memberof Extensions/Processor
  */
 Processor.prototype.parseContent = function (parent, content, attrs) {
-  return this.$parse_content(parent, content, attrs)
+  return this.$parse_content(parent, content, toHash(attrs))
 }
 
 /**


Mogztter:
Thanks @OskarsPakers, could you please add a test case to make sure that this fix is working as excepted? 

You can take a look at:

https://github.com/asciidoctor/asciidoctor.js/blob/7c3482569f2643050593aeeae685a21dd17e1f90/packages/core/spec/node/asciidoctor.spec.js#L1136-L1155

The idea is to create a simple extension in your test case that calls `parseContent` and then call `asciidoctor.convert` with your extension registry to make sure everything is working fine.
Here's a rough example:

```js
const registry = asciidoctor.Extensions.create(function () { 
  this.blockMacro(function () { 
    this.named('test') 
    this.process((parent) => { 
      return this.parseContent(parent, '*Hello world*', {'some': 'attribute'}) 
    }) 
  }) 
}) 
const html = asciidoctor.convert('test::[]', { extension_registry: registry } ) 
// assert that html contains something produced by the extension
````

Let me know if you need help to get started.

**Comment by mojavelinux** [comment]
Just to be sure it's working as expected, I would recommend parsing block content, like a list item. It's already possible to parse inline content using applySubstitutions, so we want the test to go deeper.

**Comment by Mogztter** [comment]
@OskarsPakers you can take a look at what @uniqueck did to add a test case: https://github.com/asciidoctor/asciidoctor.js/pull/1508/files

**Comment by OskarsPakers** [comment]
It is almost embarrassing how many hints you gave me :)
But I get stuck in RangeError error when I try to do a simple test below
```
      it("should be able pass through attrs", () => {
        const extensions = asciidoctor.Extensions
        try {
          extensions.register('test', function () {
            this.block(function () {
              this.named('test')
              this.process((parent) => {
                return this.parseContent(parent, '*Hello world*', {'some': 'attribute'});
              })
            })
          })
          const html = asciidoctor.convert('[test]\nreplace me')
          expect(html).to.contain(`<div class="ulist"></div>`)
          extensions.unregister('test')
        } finally {
          asciidoctor.Extensions.unregisterAll()
        }
      });
```

Error 
```
        Extension
           should be able pass through attrs:
     RangeError: Maximum call stack size exceeded
      at Function.$Opal_instance_variable_name$excl$8 (file:///C:/Users/opakers/Documents/Projects/asciidoctor.js/packages/core/node_modules/asciidoctor-opal-runtime/src/opal.js:2882:95)
      at constructor.$$instance_variable_get (file:///C:/Users/opakers/Documents/Projects/asciidoctor.js/packages/core/node_modules/asciidoctor-opal-runtime/src/opal.js:4806:62)
      at $$28 (file:///C:/Users/opakers/Documents/Projects/asciidoctor.js/packages/core/node_modules/asciidoctor-opal-runtime/src/opal.js:4771:68)
      at Opal.yield1 (file:///C:/Users/opakers/Documents/Projects/asciidoctor.js/packages/core/node_modules/asciidoctor-opal-runtime/src/opal.js:1613:14)
      at Array.$$each (file:///C:/Users/opakers/Documents/Projects/asciidoctor.js/packages/core/node_modules/asciidoctor-opal-runtime/src/opal.js:13967:21)
      at Object.Opal.send2 (file:///C:/Users/opakers/Documents/Projects/asciidoctor.js/packages/core/node_modules/asciidoctor-opal-runtime/src/opal.js:1847:17)
      at Opal.send (file:///C:/Users/opakers/Documents/Projects/asciidoctor.js/packages/core/node_modules/asciidoctor-opal-runtime/src/opal.js:1837:17)
```
I plan to extend this further to render an element that passes attributes further down.

**Comment by Mogztter** [comment]
Don't be, it's definitely not an easy task to add a test case that's why I gave you a few hints 😉

I'm glad you were able to add a test case. 

> But I get stuck in RangeError error when I try to do a simple test below

That's actually a good news. @uniqueck had the same issue so being able to reproduce it in a test case will help us fix it.

I don't have access to a computer right now so I can't really investigate. I might be able to take a look after Christmas.

Thanks for your work on this pull request and I wish you a happy holiday season. 🎄

**Comment by Mogztter** [comment]
Here's what's going on:

`parseContent` is calling `next_block` which will call `build_block`:

https://github.com/asciidoctor/asciidoctor/blob/2c0154469b4fd1c064a2fb7028a4bce637bfa3fc/lib/asciidoctor/parser.rb#L908-L911

Then, it's calling `block.lines` but `block` is a `Document` and does not have `lines` method:

https://github.com/asciidoctor/asciidoctor/blob/2c0154469b4fd1c064a2fb7028a4bce637bfa3fc/lib/asciidoctor/parser.rb#L1058-L1061

Opal provides a stub `method_missing_stub`. It should raise a `NoMethodError` but instead it's looping on `inspect`:

```js
return self.$raise($$($nesting, 'NoMethodError').$new("" + "undefined method `" + (symbol) + "' for " + (self.$inspect()), symbol, args), nil, self.$caller(1));
```

`inspect` is calling `self.$instance_variable_get(i).$inspect()`. 
`self.$instance_variable_get(i)` actually returns the `Document` object and the infinite loop begins!

I will run the same code with Asciidoctor (Ruby) in order to spot any difference.

Mogztter:
I get the actual error with Asciidoctor (Ruby):

```
	7: from test.rb:19:in `<main>'
	6: from /home/guillaume/.rvm/gems/ruby-2.7.2@asciidoctor/gems/asciidoctor-2.0.16/lib/asciidoctor/convert.rb:78:in `convert'
	5: from /home/guillaume/.rvm/gems/ruby-2.7.2@asciidoctor/gems/asciidoctor-2.0.16/lib/asciidoctor/load.rb:84:in `load'
	4: from /home/guillaume/.rvm/gems/ruby-2.7.2@asciidoctor/gems/asciidoctor-2.0.16/lib/asciidoctor/document.rb:537:in `parse'
	3: from /home/guillaume/.rvm/gems/ruby-2.7.2@asciidoctor/gems/asciidoctor-2.0.16/lib/asciidoctor/parser.rb:97:in `parse'
	2: from /home/guillaume/.rvm/gems/ruby-2.7.2@asciidoctor/gems/asciidoctor-2.0.16/lib/asciidoctor/parser.rb:385:in `next_section'
	1: from /home/guillaume/.rvm/gems/ruby-2.7.2@asciidoctor/gems/asciidoctor-2.0.16/lib/asciidoctor/parser.rb:892:in `next_block'
/home/guillaume/.rvm/gems/ruby-2.7.2@asciidoctor/gems/asciidoctor-2.0.16/lib/asciidoctor/parser.rb:1046:in `build_block': asciidoctor: FAILED: <stdin>: Failed to load AsciiDoc document - undefined method `lines' for #<Asciidoctor::Document:0x00005561da2a1390> (NoMethodError)
Did you mean?  lineno
```

So the code is incorrect and should throw an error but it should not go into infinite loop. Probably an Opal issue, I will try to reproduce it.

I think the root cause is that the `AbstractNode` holds a reference of the `Document`: https://github.com/asciidoctor/asciidoctor/blob/2c0154469b4fd1c064a2fb7028a4bce637bfa3fc/lib/asciidoctor/abstract_node.rb#L18 and the `Document` extends `AbstractBlock` which extends `AbstractNode`.

In other words, the instance `Document#123` holds a reference of `Document#123` as an instance variable named `document`.

It seems that MRI is smart enough to not go into infinite loop:

```
#<Asciidoctor::Document:0x0000560cc1721608 @document=#<Asciidoctor::Document:0x0000560cc1721608 ...>
```

As you can see, MRI outputs `#<Asciidoctor::Document:0x0000560cc1721608 ...>` to avoid the infinite loop.

Mogztter:
I was able to reproduce it: https://github.com/opal/opal/issues/2384

Mogztter:
@mojavelinux might be a bug upstream since https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/master/lib/textql-block.rb does not work anymore (same error)

mojavelinux:
The extension is incorrect. `parse_content` return the parent (since it can add arbitrary number of blocks to the parent provided). The extension should be returning nil.

mojavelinux:
The root of the problem is that the process method should not be returning the result of parseContent. Instead, it should return undefined (meaning the blocks have already been added).

Mogztter:
To clarify, the extension should not return a `Document` otherwise it won't work. `parse_content` returns a `Document`: https://github.com/asciidoctor/asciidoctor/blob/5857ca755c38f9eeaded34dae4b4f6871249d2eb/lib/asciidoctor/extensions.rb#L213-L217 so basically we need to create a wrapper:

```js
const registry = asciidoctor.Extensions.create()
registry.block(function () {
  this.named('test')
  this.process((parent) => {
    const wrapper = this.createOpenBlock(parent, [], {})
    this.parseContent(wrapper, '*Hello world*', { 'some': 'attribute' })
    return wrapper
  })
})
const html = asciidoctor.convert('[test]\nreplace me', { extension_registry: registry })
```
Not sure if this is intended or not.

Here's another example in Ruby: https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/e3d5219679b60be7b6710004b44621419e5fc49b/lib/compound-block.rb

Mogztter:
> The root of the problem is that the process method should not be returning the result of parseContent. Instead, it should return undefined (meaning the blocks have already been added).

Indeed, the following also works:

```js
const registry = asciidoctor.Extensions.create()
registry.block(function () {
  this.named('test')
  this.process((parent) => {
    this.parseContent(parent, '*Hello world*', { 'some': 'attribute' })
  })
})
const html = asciidoctor.convert('[test]\nreplace me', { extension_registry: registry })
```

mojavelinux:
It's more accurate to say that `parse_document` returns the parent you gave it.

There is a test in core that shows how you can return the return value of `parse_content`, but only if the parent you passed in is itself a block. See https://github.com/asciidoctor/asciidoctor/blob/main/test/extensions_test.rb#L1575-L1611

Mogztter:
> It's more accurate to say that `parse_document` returns the parent you gave it.

👍🏻 
Would you accept an enhancement in Asciidoctor core that adds an explicit error message when the extension does not return a `Block`? We could catch the `NoMethodError` error and add some context? The cause would be `NoMethodError` and the error would be something like: `extension #<Asciidoctor::BlockProcessor:0x00001234abc> must return a Block or nil`.

Mogztter:
Superseded https://github.com/asciidoctor/asciidoctor.js/pull/1519

mojavelinux:
The policy in core has been to not add defensive programming checks in areas like this. I would rather focus on a) handling the situation gracefully and b) documenting what should be done.

I do think there is a bug in core. The reason it's going for the `lines` method is because it's assuming the return value is a Block. However, the code would otherwise work if the return value is any sort of AbstractBlock. So I think we need an extra check for the type so that we only do this special handing (automatic parsing of children) for a Block.

I'll file an issue in core and we can follow-up there and discuss.

mojavelinux:
> Would you accept an enhancement in Asciidoctor core that adds an explicit error message when the extension does not return a Block? 

I did some more digging into this and it turns out that core was doing the wrong thing in several instances. You're right that it should not try to invoke `.lines` if the type is not a `Block`. It should also ignore the return value if it matches the parent argument. This avoids unintended crashes in a number of cases without needing to warn the extension author about it. See https://github.com/asciidoctor/asciidoctor/commit/de1ba697bdbc8da1a6598269f4b0093c83887f5b. This fix will be included in 2.0.17.

Mogztter:
> It should also ignore the return value if it matches the parent argument.

I like this change 👍🏻 

> This fix will be included in 2.0.17.

🎉

PR #1491, opened 2021-12-04
resolves #1475 bridge common Ruby object methods

### LINKED ISSUE #1475: `TypeError: converter.$==` is not a function on custom converter
Hi, im the maintainer of the asciidoc slides vscode extension which depends on @asciidoctor/core@2.2.5 and @asciidoctor/reveal.js: 5.0.0-rc.1, and when i try to convert a document by

```javascript
const vscode = require('vscode');
var asciidoctor = require('@asciidoctor/core')()
var asciidoctorRevealjs = require('@asciidoctor/reveal.js')
asciidoctorRevealjs.register()
var content = '= Document title'
asciidoctor.convert(asciidocText, { safe: 'safe', backend: 'revealjs', attributes: { docDir: this.absoluteDocumentDirectory } });
```

I get the following error message:

```
rejected promise not handled within 1 second: TypeError: converter.$== is not a function
```

This error seams to appear since vscode version >= 1.61.2 and did not appear before.
Is it a misconfiguration on my side or may be a bug on your side?

Mogztter:
It's more a workaround than an actual solution but should be good enough.

resolves #1475

### DIFF
--- a/packages/core/spec/node/asciidoctor.spec.js
+++ b/packages/core/spec/node/asciidoctor.spec.js
@@ -2460,6 +2460,18 @@ getAttribute('fragment'): ${typeof node.getAttribute('fragment') === 'undefined'
       }
     }
 
+    it('should bridge common Ruby object methods', () => {
+      asciidoctor.ConverterFactory.register(new DummyConverter(), ['dummy'])
+      const converterFactory = asciidoctor.ConverterFactory.getDefault()
+      const dummyConverterInstance = converterFactory.create('dummy')
+      const html5ConverterInstance = converterFactory.create('html5')
+      expect(dummyConverterInstance['$=='](dummyConverterInstance)).to.be.true()
+      expect(html5ConverterInstance['$=='](html5ConverterInstance)).to.be.true()
+      expect(dummyConverterInstance['$=='](html5ConverterInstance)).to.be.false()
+      expect(typeof html5ConverterInstance.$send).to.equal('function')
+      expect(typeof dummyConverterInstance.$send).to.equal('function')
+      expect(dummyConverterInstance.$send('convert', { getContent: () => 'Hello World' }, 'paragraph')).to.equal('Hello World')
+    })
     it('should get inline anchor attributes', () => {
       asciidoctor.ConverterFactory.register(new XrefConverter(), ['xref'])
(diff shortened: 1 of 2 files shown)


Mogztter:
@mojavelinux I would be interesting in hearing your thoughts. The main issue is when we call `ConverterFactory.register` with an instance (created from a plain JavaScript class):

```js
class DummyConverter {
  // ...
}

asciidoctor.ConverterFactory.register(new DummyConverter(), ['dummy'])
```

In this case, the instance is missing quite a few Ruby methods.
It's possible to use `Opal.klass` to bridge a native JS class but as far as I know it's not possible to bridge a instance:

https://github.com/opal/opal/blob/a46adeac3cb7b5e13a8ac9e77dd3aa11278ccef8/opal/corelib/runtime.js#L625-L651

Also, we would need to specify the "superclass" in the register function.

mojavelinux:
It seems reasonable to me. I'll admit, I'm not a big fan of this porcelain APIs because of the bridging that has to happen underneath. (I prefer to work within Opal). But I understand that it's important to a certain group of users...and thus the details should be hidden as much as possible in those cases.

Mogztter:
> It seems reasonable to me.

👍🏻 

> I'll admit, I'm not a big fan of this porcelain APIs because of the bridging that has to happen underneath. (I prefer to work within Opal). But I understand that it's important to a certain group of users...and thus the details should be hidden as much as possible in those cases

Yes it's a trade-off, the porcelain API is not perfect and won't give the full power of using Opal directly but it's definitely more pleasant to use when you are not familiar with Opal.

flobilosaurus:
Hi, is there any plan when this fix is integrated ?

jayczech23:
Hi all, also interested in hearing the timeline for integration of this fix.

Thank you :D .

Mogztter:
@jayczech23 Since it's a low-level bug fix I'm curious to know why you are waiting for this fix? Are you using [the AsciiDoc Slides VS code extension](https://github.com/flobilosaurus/vscode-asciidoc-slides)?

jayczech23:
@Mogztter, I am actually attempting to implement RevealJS into an Electron app, came across this error when using Asciidoctor RevealJS API described in [docs ](https://docs.asciidoctor.org/reveal.js-converter/latest/setup/node-js-setup/#using-the-javascript-api)

Issue #1483, opened 2021-11-30
self.parseContent(parent, content, attrs) throws a TypeError: attributes.$dup is not a function exception.


uniqueck:
I wrote an extension for antora and i use the method parseContent(parent, content, attrs).
When I pass the provided attrs from my extension to the method parseContent, I got the error 
 
`TypeError: attributes.$dup is not a function exception.`.

As an workaround I can use Opal.hash(attrs) and pass this to the method, but I think we can do this internally.

```javascript
function jiraIssuesBlockMacro (context) {
  return function () {
    const self = this
    self.named('jiraIssues')
    self.positionalAttributes(['target', 'jql'])
    self.process((parent, target, attrs) => {
      const doc = parent.getDocument()
      const projectKey = target
      const jql = attrs.jql || "resolution='Unresolved' ORDER BY priority DESC, duedate ASC"
      const jiraClient = new Jira(doc)
      const issues = jiraClient.searchIssues(jql)

      var content = []
      
      return self.parseContent(parent, content.join('\n'), attrs)

    })
  }
}
```

Mogztter:
Duplicates #1036

PR #1508, opened 2021-12-19
add simple reproducable test case for extension uses parseContent

uniqueck:
Add minimal test case for #1483


packages/core/spec/share/extensions/jira-block-macro.cjs
```diff
@@ -0,0 +1,41 @@
...
+          const issue = issues[i]
+          content.push('|' + issue.key)
+          content.push('|' + issue.fields.priority.name)
+          content.push('|' + issue.fields.created)
+          content.push('|' + (issue.fields.assignee && issue.fields.assignee.displayName) || 'not assigned')
+          content.push('|' + issue.fields.summary)
+        }
+        content.push('|====')
+
+        return self.parseContent(parent, content.join('\n'), attrs)
```
Mogztter, line 29: This is incorrect, you should do:

```suggestion
        self.parseContent(parent, content.join('\n'), attrs)
```

Mogztter:
Thanks for providing a test case.

Superseded #1519

PR #1523, opened 2021-12-31
resolves #1507 introduce an in-memory cache (Node.js)

### LINKED ISSUE #1507: Implement the content URI caching feature in a Node environment
Asciidoctor provides a built-in cache for content URI (based on `open-uri-cached` gem).
Currently, this feature is not available in Asciidoctor.js because the `open-uri-cached` is not transpile from Ruby to JavaScript.

I think it could be great to implement this feature in a Node environment. This feature will use the `cache-uri` attribute to enable the cache (similar to what Asciidoctor Ruby is doing: https://docs.asciidoctor.org/asciidoc/latest/directives/include-uri/#caching-uri-content)

To implement this feature, we could use: https://github.com/isaacs/node-lru-cache in order to set expiration time depending on response headers, configure a default max size, etc...

ggrossetie:
resolves #1507


Mogztter:
This pull request provides a custom implement of `open-uri/cached`.
The implement is heavily inspired by: https://github.com/isaacs/node-lru-cache but does not rely on it (mostly to avoid additional dependencies).

This pull requests also introduces an API to interact with the cache:

```js
Asciidoctor.Cache.enable()
Asciidoctor.Cache.disable()
Asciidoctor.Cache.setMax(500)
Asciidoctor.Cache.reset()
```

packages/core/src/template-asciidoctor-node.js
```diff
@@ -28,5 +31,40 @@ export default function (moduleConfig) {
...
+  Opal.Asciidoctor.Cache = {
+    disable: function () {
+      // QUESTION: should we also reset cache?
+      const openUriSingleton = Opal.OpenURI.$singleton_class()
+      if (Opal.OpenURI['$respond_to?']('original_open_uri')) {
+        openUriSingleton.$send('remove_method', 'open_uri')
+        openUriSingleton.$send('alias_method', 'open_uri', 'original_open_uri')
+      }
+    },
+    reset: function() {
```
Mogztter, line 45: We might rename this function to `clear`. That way it's "clear" (pun intended) that the function will clear the cache (and not reset the max value)

packages/core/lib/asciidoctor/js/asciidoctor_ext/node/helpers.rb
```diff
@@ -0,0 +1,35 @@
...
+  # Returns The [Boolean] return value of Kernel#require if the library can be loaded.
+  # Otherwise, if on_failure is :abort, Kernel#raise is called with an appropriate message.
+  # Otherwise, if on_failure is :warn, Kernel#warn is called with an appropriate message and nil returned.
+  # Otherwise, nil is returned.
+  def require_library name, gem_name = true, on_failure = :abort
+    if name == 'open-uri/cached'
+      `return Opal.Asciidoctor.Cache.enable()`
+    end
+    original_require_library name, gem_name, on_failure
+  end
```
Mogztter, line 33: @mojavelinux I'm a bit concerned about this hack, could we find a better "hook" in Asciidoctor core?

Mogztter:
Waiting for a change in Asciidoctor core to properly hook the in-memory cache.

PR #1550, opened 2022-01-19
resolves #1475 bridge common Ruby object methods (backport v2.2.x)

### LINKED ISSUE #1475: `TypeError: converter.$==` is not a function on custom converter

Mogztter:
We are using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from which is not available in IE.

I think we should mention that in the next release but I don't want to include a polyfill.

### DIFF
--- a/packages/core/spec/node/asciidoctor.spec.js
+++ b/packages/core/spec/node/asciidoctor.spec.js
@@ -2416,6 +2416,18 @@ getAttribute('fragment'): ${typeof node.getAttribute('fragment') === 'undefined'
       }
     }
 
+    it('should bridge common Ruby object methods', () => {
+      asciidoctor.ConverterFactory.register(new DummyConverter(), ['dummy'])
+      const converterFactory = asciidoctor.ConverterFactory.getDefault()
+      const dummyConverterInstance = converterFactory.create('dummy')
+      const html5ConverterInstance = converterFactory.create('html5')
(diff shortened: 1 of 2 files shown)


mojavelinux:
It also uses the spread assignment, which wouldn't work in IE either. However, a more portable and likely more efficient way of doing it is:

```js
var args = Array.prototype.slice.call(arguments, 1)
```

Mogztter:
> It also uses the spread assignment, which wouldn't work in IE either. However, a more portable and likely more efficient way of doing it is:

Good suggestion, thanks!

PR #1653, opened 2022-09-16
remove call to convert() on return value of process method for inline macro extension

mojavelinux:
- the process method should return an Inline object, not a string


Mogztter:
👍

mojavelinux:
Thanks for merging! In order to make this live, I think it needs to be backported to the v2.2.x branch too.

PR #1659, opened 2022-09-24
Add `getSectionNumeral()` function

benjaminleonard:
Currently using `$sectnum()` in a custom asciidoc section converter—figured it might be useful to get the function wrapper included and typed.

First time contributor, hopefully it's all formatted correctly.

### DIFF
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -745,6 +745,15 @@ Section.prototype.setSectionName = function (value) {
 }
 
 /**
+ * Set the section number of this section.
+ * @returns {string}
+ * @memberof Section
+ */
+Section.prototype.getSectionNumber = function () {
+  return this.$sectnum()
+}
+
+/**
  * Get the flag to indicate whether this is a special section or a child of one.
  * @returns {boolean}
  * @memberof Section
--- a/packages/core/src/asciidoctor-core-api.js
+++ b/packages/core/src/asciidoctor-core-api.js
@@ -749,7 +749,7 @@ Section.prototype.setSectionName = function (value) {
  * @returns {string}
  * @memberof Section
  */
-Section.prototype.getSectionNumber = function () {
+Section.prototype.getSectionNumeral = function () {
   return this.$sectnum()
 }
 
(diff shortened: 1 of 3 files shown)


Mogztter:
That's look good! Thank you 😄 
As mentioned by Dan on Zulip, could you please use `getSectionNumeral` as the primary name (and `getSectionNumber` as an alias)?

> When you file the issue, I propose the name getSectionNumeral(). The value is not always a number, which is why we use that terminology. However, I'd also support adding getSectionNumber() as an alias.

https://asciidoctor.zulipchat.com/#narrow/stream/279645-users.2Fasciidoctor.2Ejs/topic/Accessing.20sectnum.20with.20asciidoctor.2Ejs/near/298429578

benjaminleonard:
Oh I missed that, done!

I can add the alias too if you would like—and have an example.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -744,6 +744,15 @@ Section.prototype.setSectionName = function (value) {
   this.sectname = value
 }
 
+/**
+ * Get the section number of this section.
```
Mogztter, line 748:
```suggestion
 * Get the section numeral of this section.
```

packages/core/types/index.d.ts
```diff
@@ -2247,6 +2247,11 @@ export namespace Asciidoctor {
      */
     setSectionName(value: string): void;
 
+    /**
+     * Get the section number of this section.
```
Mogztter, line 2251:
```suggestion
     * Get the section numeral of this section.
```

Mogztter:
> I can add the alias too if you would like—and have an example.

I think that would be nice since a lot of users might assume that setnum is short for section number (numeral is less common)

benjaminleonard:
Done! (I think)

packages/core/types/index.d.ts
```diff
@@ -2247,6 +2247,11 @@ export namespace Asciidoctor {
      */
     setSectionName(value: string): void;
 
+    /**
+     * Get the section numeral of this section.
+     */
+    getSectionNumeral(): string;
+
```
Mogztter, line 2254: Almost done, we also need to add `getSectionNumber` in the definition type.

benjaminleonard: Ha believe it or not I did this and had not yet saved the file.

Mogztter:
It looks good, thanks again, I will merge it once the build is successful

PR #1686, opened 2023-01-30
Include `cjs` in release

benjaminleonard:
Hi, when importing @asciidoctor/core@3.0.0-alpha.4  with Remix (https://remix.run/) I get the following error:

`TypeError: The "path" argument must be of type string or an instance of URL. Received undefined`

Applying to this line:
https://github.com/asciidoctor/asciidoctor.js/blob/aac44ade02c1ad67f1154fb49243fd0f32f09f51/packages/core/src/template-asciidoctor-node.js#L12

This seems to be because `import.meta.url` is not available in `cjs`. Essentially, Remix converts `esm` packages to `cjs` to run on the server (see here: https://remix.run/docs/en/v1/pages/gotchas#importing-esm-packages).

I'm proposing we keep the release as it is, but copy over the `cjs` as a fallback so it can be used as a dual CommonJS/ES module package.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -2,7 +2,12 @@
   "name": "@asciidoctor/core",
   "version": "3.0.0-alpha.4",
   "description": "Asciidoctor - the core library",
-  "main": "dist/node/asciidoctor.js",
+  "exports": {
+    ".": {
+     "require": "./dist/node/asciidoctor.cjs",
+     "import": "./dist/node/asciidoctor.js"
+    }
+  },
   "browser": "dist/browser/asciidoctor.js",
   "module": "dist/browser/asciidoctor.js",
   "type": "module",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -2,10 +2,11 @@
   "name": "@asciidoctor/core",
   "version": "3.0.0-alpha.4",
   "description": "Asciidoctor - the core library",
+  "main": "dist/node/asciidoctor.js",
   "exports": {
     ".": {
-     "require": "./dist/node/asciidoctor.cjs",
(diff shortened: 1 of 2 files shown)


mojavelinux:
I'm in favor as Antora will need this as well. I have no plans to switch Antora to ESM right now as it requires a major change to the build process. Having a CJS export would be very helpful.

packages/core/package.json
```diff
@@ -2,7 +2,12 @@
   "name": "@asciidoctor/core",
   "version": "3.0.0-alpha.4",
   "description": "Asciidoctor - the core library",
-  "main": "dist/node/asciidoctor.js",
+  "exports": {
+    ".": {
+     "require": "./dist/node/asciidoctor.cjs",
+     "import": "./dist/node/asciidoctor.js"
+    }
+  },
```
ggrossetie, line 10: Can we keep `main` in addition to `exports`? I believe that some web bundlers/CDNs are using `main` as an hint to tell if the module is ESM-compatible or not.

benjaminleonard: 👍🏻 Done

ggrossetie:
Thanks!

benjaminleonard:
If you get the chance I would love an alpha release including this PR.

Thanks!

ggrossetie:
At this point, I think I should release a "rc". 
I've been experimenting a bit with Asciidoctor.js 3.x ESM and I haven't noticed any issue.

I will probably upgrade to the latest release of Opal as well (from 1.5.1 to [1.7.3](https://github.com/opal/opal/releases/tag/v1.7.3)).

benjaminleonard:
That'd be perfect! Thanks

PR #1695, opened 2023-04-19
Fix conditional export

benjaminleonard:
I'm guessing this worked with `npm link` because the `src` dir exists. I can confirm the change works importing the `cjs` with `remix`.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asciidoctor/core",
-  "version": "3.0.0-rc.2",
+  "version": "3.0.0-rc.1",
   "description": "Asciidoctor - the core library",
   "type": "module",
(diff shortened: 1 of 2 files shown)


packages/core/package.json
```diff
@@ -9,7 +9,7 @@
   "exports": {
     "node": {
       "import": "./dist/node/asciidoctor.js",
-      "require": "./src/node/asciidoctor.cjs"
```
ggrossetie, line 12: My bad 😅

ggrossetie:
@benjaminleonard could you please revert this commit [`f492a14` (#1695)](https://github.com/asciidoctor/asciidoctor.js/pull/1695/commits/f492a1477e057dd20dcbe05b303757872326d66c) ? We have an automated process to publish release (bump version, tag...)

Issue #1588, opened 2022-02-09
Refine package exports


ggrossetie:
We should refine exports as described in: https://nodejs.org/api/packages.html#packages_package_entry_points

Currently, we define the following:

```json
{
  "type": "module"
  "main": "dist/node/asciidoctor.js",
  "browser": "dist/browser/asciidoctor.js",
  "module": "dist/browser/asciidoctor.js",
}
```

This is not good enough because web bundlers prioritize `module` when the target is `node`.
As a result, we need to 

```json
{
  "type": "module"
  "main": "dist/node/asciidoctor.cjs",
  "browser": "dist/browser/asciidoctor.umd.js",
  "module": "dist/browser/asciidoctor.module.js",
  "exports": {
    ".": {
      "node": {
        "import": "./dist/node/asciidoctor.js",
        "require": "./dist/node/asciidoctor.cjs",
        "default": "./dist/node/asciidoctor.cjs"
      },
      "browser": "./dist/browser/asciidoctor.umd.js",
      "umd": "./dist/browser/asciidoctor.umd.js",
      "import": "./dist/browser/asciidoctor.module.js",
      "default": "dist/browser/asciidoctor.umd.js"
    }
  }
}
```

Open questions:
- Should we support `unpkg` (alias `umd:main`)? https://github.com/developit/microbundle#-installation--setup--
- Should we minify browser files?
- Should we use `development` / `production` in order to provide unminified files in development and minified files in production? 
- According to: https://webpack.js.org/guides/package-exports/#providing-different-versions-depending-on-target-environment we might need to configure `module` in the `exports` in addition to `import`?

As you can see, the above definition prioritize the browser. I think this is safer since `exports` is not broadly supported by web bundlers (which most of the time target the browser).

Mogztter:
Without UMD:

```json
{
  "type": "module"
  "main": "dist/node/asciidoctor.cjs",
  "browser": "dist/browser/asciidoctor.js",
  "module": "dist/browser/asciidoctor.js",
  "exports": {
    ".": {
      "node": {
        "import": "./dist/node/asciidoctor.js",
        "require": "./dist/node/asciidoctor.cjs",
        "default": "./dist/node/asciidoctor.js"
      },
      "default": "dist/browser/asciidoctor.js"
    }
  }
}
```

> "default" - the generic fallback that always matches. Can be a CommonJS or ES module file. This condition should always come last.

https://nodejs.org/api/packages.html#conditional-exports

[...]

PR #1708, opened 2023-07-21
fix(core): typings path and export

sinedied:
Typescript compiler doesn't seem to find the types unless the full path is specified.
And since the types path doesn't match the exports path, it needs to be added there as well

image

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -8,12 +8,19 @@
   "module": "dist/browser/asciidoctor.js",
   "exports": {
     "node": {
-      "import": "./dist/node/asciidoctor.js",
-      "require": "./dist/node/asciidoctor.cjs"
+      "import": {
+        "default": "./dist/node/asciidoctor.js",
+        "types": "./types/index.d.ts"
+      },
+      "require": {
+        "default": "./dist/node/asciidoctor.cjs",
+        "types": "./types/index.d.ts"
+      }
     },
-    "default": "./dist/browser/asciidoctor.js"
+    "default": "./dist/browser/asciidoctor.js",
+    "types": "./types/index.d.ts"
   },
-  "types": "types",
+  "types": "types/index.d.ts",
   "engines": {
     "node": ">=16",
     "npm": ">=8"


packages/core/package.json
```diff
@@ -8,12 +8,19 @@
...
     "node": {
-      "import": "./dist/node/asciidoctor.js",
-      "require": "./dist/node/asciidoctor.cjs"
+      "import": {
+        "default": "./dist/node/asciidoctor.js",
+        "types": "./types/index.d.ts"
+      },
+      "require": {
+        "default": "./dist/node/asciidoctor.cjs",
+        "types": "./types/index.d.ts"
```
ggrossetie, line 17: It seems that we need to specify a `d.cts` types file (CommonJS). Since we are using `export default function` I don't think we should specify `types` on `require`.

sinedied: You may be right regarding the `d.cts` extension, seeing https://github.com/microsoft/TypeScript/issues/52363
But if you look at the answers in the same thread, I think there's still need to specify the `types` explicitly on the `require` conditional, as there's not matching patch TypeScript will not load it.

My current project is configured for ESM so I could not test the CJS part, it would be helpful to check it on a real CJS project. I'll try to build a test project for that next monday

ggrossetie: Thanks that would be helpful!
If it works on CJS we can proceed and merge.

Techassi:
I'm running into the same error. Related issues might be #1711 and #1588.

ggrossetie:
@Techassi if you could confirm that the above configuration works in a CJS project that would be great.

Techassi:
I also don't use CJS style imports, but I can create a small test project and try it out. Providing the correct path is # essential, as currently the `asciidoctor/core` is unusable when using ESM style imports.

Techassi:
I just tested it locally. This set of options in `package.json` seem to work:

```json
{
  "exports": {
    "require": {
      "default": "./dist/node/asciidoctor.cjs",
      "types": "./types/index.d.ts"
    },
    "import": {
      "default": "./dist/node/asciidoctor.js",
      "types": "./types/index.d.ts"
    }
  },
  "types": "types/index.d.ts",
}
```

We might want to change `./types/index.d.ts` to `./types/index.d.cts` in the `require` group of options. I didn't look into generating this file (it should be a simple copy) as the build scripts seem quite convoluted.

**Side note:** It might be good to drop CJS style imports (and types) altogether, as the JS ecosystem is moving to the more modern ESM style. It was quite a hassle to set up this local test project as most modern tooling defaults to ESM.

ggrossetie:
> I just tested it locally. This set of options in package.json seems to work:

OK, so even if the file `index.d.ts` contains ESM `import` and `export` keywords it's working on a CJS project?

> We might want to change ./types/index.d.ts to ./types/index.d.cts in the require group of options. I didn't look into generating this file (it should be a simple copy) as the build scripts seem quite convoluted.

I found a rollup pluign: https://github.com/Swatinem/rollup-plugin-dts

> Side note: It might be good to drop CJS style imports (and types) altogether, as the JS ecosystem is moving to the more modern ESM style. It was quite a hassle to set up this local test project as most modern tooling defaults to ESM.

We dropped CJS in the first release candidate of Asciidoctor.js but the community was vocal about keeping it. Asciidoctor.js code base is now using ESM and we are using rollup to produce CJS. We might remove CJS support in the next major release.

ggrossetie:
Superseded by #1715 (@sinedied I cherry-picked your commit)

Techassi:
> OK, so even if the file `index.d.ts` contains ESM `import` and `export` keywords it's working on a CJS project?

Yes seems like it. I got the correct typing information when using CJS imports and the above `export`.

> I found a rollup pluign: [Swatinem/rollup-plugin-dts](https://github.com/Swatinem/rollup-plugin-dts)

Oh yeah, I actually use it in one of my personal projects as well. Had great success working with it. It might require some output path adjustments tho.

> We might remove CJS support in the next major release.

Would love to see that!

---

Thanks for getting this merged super quick. Will test it in the real project asap!

sinedied:
Thanks @Techassi and @ggrossetie for picking up this and completing it 👍

Techassi:
Do you have any plans on when this will be released @ggrossetie? Currently, I still use a local dependency as a workaround.

ggrossetie:
I will try to push a new release before my vacation (mid January)

PR #1718, opened 2023-12-27
resolves #1719 add context and node_name accessor in the type definition

### LINKED ISSUE #1719: Setting context of Abstract node no longer working?
Hi there.

Ive submitted a PR which adds a setContext to the AbstractNode.

https://github.com/asciidoctor/asciidoctor.js/pull/1718

This was originally done because I thought that the ability to set the value had been removed. 

This line used to work:

`node.context = 'colist'`

But running it against Asciidoctor.js 3.0.2, and my IDE flags up that it cannot access change the value directly, and the running code comes back with:

```
error TS2339: Property 'context' does not exist on type 'AbstractBlock'.
```

which leads me to believe that the context is no longer exposed. (I have a feeling that being able to set it directly in version 2.2.6 may have been a happy accident that that I was taking advantage of.

So why am I fiddling with it. Well I wrote a Asciidoctor extension a few years ago, that uses that to change an ordered list into a callout list. 

https://www.npmjs.com/package/asciidoctor-external-callout

I'd like to update it to make sure it can still work with current and future versions of the API.

RayOffiah:
Hi there 👋🏾

In previous versions of the API, it was possible to change the context and name of a block/node by reassigning it in the source code.

```Javascript
block.context = block.node_name = newName
```

I noticed that the ability to do this has been removed from the API. 

It isn't a function that is going to get used a lot, but some code I wrote a while back does rely on it.

This PR will restore the functionality, but uses proper `set` functions to change the values.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /node_modules/
 npm-debug.log
+/.idea/
--- a/packages/core/.gitignore
+++ b/packages/core/.gitignore
@@ -6,3 +6,6 @@ npm-debug.log
 
 # generated using rollup
 /spec/node/asciidoctor.spec.cjs
+
+# IntelliJ project files
+/.idea/
--- a/packages/core/spec/node/asciidoctor.spec.js
+++ b/packages/core/spec/node/asciidoctor.spec.js
@@ -825,6 +825,22 @@ image::https://asciidoctor.org/images/octocat.jpg[GitHub mascot]`
       expect(blocksWithLineNumber.length >= 18).to.be.true()
     })
 
+    it('should be able to convert the context of a block', () => {
+      const doc = asciidoctor.loadFile(resolveFixture('documentblocks.adoc'))
+      const blocks = doc.findBy({ context: 'ulist' })
+      expect(blocks.length).to.equal(2)
+      blocks[0].context = 'colist'
+      expect(blocks[0].getContext()).to.equal('colist')
+    })
+
+    it('should be able to set the name of a node', () => {
+      const doc = asciidoctor.loadFile(resolveFixture('documentblocks.adoc'))
+      const blocks = doc.findBy({ context: 'ulist' })
+      expect(blocks.length).to.equal(2)
+      blocks[0].node_name = 'colist'
+      expect(blocks[0].getNodeName()).to.equal('colist')
+    })
+
     if (asciidoctorCoreSemVer.gte('200')) {
       // REMIND: Before Asciidoctor 2.0.0 date was not UTC
       it('should get document date (and honor SOURCE_DATE_EPOCH)', () => {
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -2716,6 +2716,10 @@ export class AbstractNode implements Logging {
    *
    * @returns An Array of Strings representing the substitution operation or nothing if no subs are found.
    */
+
+  node_name: string;
+  context: string
+
   resolveSubstitutions(subs: string, type?: string, defaults?: string[], subject?: string): string[] | undefined;
 
   /**
--- a/packages/core/types/tests.ts
+++ b/packages/core/types/tests.ts
@@ -108,6 +108,19 @@ assert(doc.getSourcemap());
 // Block
 const block = processor.Block.create(doc, 'paragraph');
 assert(block.getContext() === 'paragraph');
+
+// Try to alter the block context
+block.context = 'ulist';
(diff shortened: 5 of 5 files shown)


ggrossetie:
Hey! Could you please open an issue so we can discuss this change?


> I noticed that the ability to do this has been removed from the API.

I don't recall removing this API 🤔 

> it was possible to change the context and name of a block/node by reassigning it in the source code.
> It isn't a function that is going to get used a lot, but some code I wrote a while back does rely on it.

Could you please provide an example (in the issue) that rely on it?

RayOffiah:
Yep, no problem. 

I'll work up an issue tomorrow.

packages/core/src/asciidoctor-core-api.js
```diff
@@ -1096,6 +1096,14 @@ AbstractNode.prototype.getNodeName = function () {
...
 }
 
+/**
+ * Sets the name of the node.
+ *
+ * @param {string} nodeName - The name to be set for the node.
+ */
+AbstractNode.prototype.setNodeName = function (nodeName) {
+  this.node_name = nodeName
+}
```
ggrossetie, line 1106: `node_name` is declared as an attribute reader: https://github.com/asciidoctor/asciidoctor/blob/7d856042fdd278f87d4c480180440e519fa1e97c/lib/asciidoctor/abstract_node.rb#L23-L24

@mojavelinux should we use `attr_accessor` on `node_name`?

mojavelinux: I'm going to warn you, we are really getting into some internal state here. The node_name variable has never been formalized and allowing it to be changed is really undefined. If anything, setting the context should set the node_name so it remains internal. But for an inline node, the node_name is "inline_" + the context.

ggrossetie: I'm reluctant to add it to the "porcelain" API. I would be in favour of adding two fields in the TypeScript definition `node_name` and `context`.

That way `block.context = block.node_name = newName` won't raise a warning.

RayOffiah: Mmm, yes. 🤔

It sounds like I'm skating very close to breaking stuff, now. 

I have a vague memory of having to set the node name in the JavaScript API, but that wasn't needed for the Ruby API. I'll try a bit of test code to see if I can do this without fiddling with the node name. 

Or perhaps I need to look at doing this a different way. Rather than alter the context, it would be safer to remove olist block and replace it with a colist built from scratch. (Would be more work though).

mojavelinux:
I'm not saying you can't set the node name, but rather that setting the
context should set the node name as the latter is internal.

packages/core/src/asciidoctor-core-api.js
RayOffiah: > I'm reluctant to add it to the "porcelain" API. I would be in favour of adding two fields in the TypeScript definition `node_name` and `context`.
> 
> 
> 
> That way `block.context = block.node_name = newName` won't raise a warning.

They would be picked up by the underlying Ruby API if I just add two fields in the TypeScript file?

ggrossetie: Yes, I think so.
As far as I recall, Opal is using un-prefixed fields for `attr_reader`.

ggrossetie:
@RayOffiah Looks good!

Could you please add a test in: https://github.com/asciidoctor/asciidoctor.js/blob/main/packages/core/types/tests.ts. It should verify that declaring `node_name` and  `context` as fields is enough.

mojavelinux:
I still don't think node_name should be exposed. Instead, setContext should set context and node_name. In Ruby, context is a symbol and node_name is a string, but in JavaScript there are no symbols (of the Ruby variety), so these two fields can just be synced. There's no reason that context and node_name should be different. This was purely an internal optimization to split them.

packages/core/src/asciidoctor-core-api.js
RayOffiah: Yup, that seems to work 👍🏾

RayOffiah:
> @RayOffiah Looks good!
> 
> 
> 
> Could you please add a test in: https://github.com/asciidoctor/asciidoctor.js/blob/main/packages/core/types/tests.ts. It should verify that declaring `node_name` and  `context` as fields is enough.

Okay, will do. 👍🏾

ggrossetie:
All good!
@mojavelinux I think we can add a `setContext` that sets both `node_name` and `context` at a later point in time.

mojavelinux:
I've made my case ;)

PR #1722, opened 2024-01-22
Fix: Default condition should be last one

roseckyj:
In newer versions of Node.js, the module resolution fails as Node expects the default condition to be the last one:

```
Module not found: Error: Default condition should be last one
```

As discussed [here](https://stackoverflow.com/questions/76127288/module-not-found-error-default-condition-should-be-last-one) this simple switch should fix the import

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -9,16 +9,16 @@
   "exports": {
     "node": {
       "import": {
-        "default": "./dist/node/asciidoctor.js",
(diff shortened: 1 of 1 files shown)


ggrossetie:
Found it: https://nodejs.org/api/packages.html#conditional-exports

> "default" - the generic fallback that always matches. Can be a CommonJS or ES module file. **This condition should always come last.**

PR #1726, opened 2024-03-28
Updated path to default Asciidoctor CSS file (NPM dist) in quick-tour.adoc

nwittwer:
Docs still mention a node_modules path to `asciidoctor.js` that no longer exists, at least as of  "asciidoctor" package v2.2.7. Looks like that has moved to `@asciidoctor/core` package.


ggrossetie:
You are correct, we are using `@asciidoctor/core` and `asciidoctor` package since 2.0.0. Thanks for the fix!

PR #1730, opened 2024-05-12
remove outdated rx patch

mojavelinux:
This rx is now defined in core. See:

https://github.com/asciidoctor/asciidoctor/blob/v2.0.x/lib/asciidoctor/rx.rb#L522-L527

### DIFF
--- a/packages/core/lib/asciidoctor/js/postscript.rb
+++ b/packages/core/lib/asciidoctor/js/postscript.rb
@@ -4,5 +4,3 @@ require 'asciidoctor/extensions'
 
 require 'asciidoctor/js/asciidoctor_ext'
 require 'asciidoctor/js/opal_ext/logger' # override the built-in Logger
-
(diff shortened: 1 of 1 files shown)


ggrossetie:
One test is failing:

```
TypeError: no implicit conversion of NilClass into String
```

It seems related to this change. When the regexp was added in Asciidoctor core?

mojavelinux:
This should be fixed by https://github.com/asciidoctor/asciidoctor/commit/d1a18cfdbc9b1294850fc18c6285a741b05b0a87

mojavelinux:
Ah, I think this change has to be merged after 2.0.23 is released since the Opal-specific rx was added after the 2.0.22 release.

mojavelinux:
The alternative is to forward-port the change to the regexp...though there is also a change to the main code as well, so that might not work on its own.

PR #1748, opened 2025-05-13
Fix missing docs for `resolveSubstitutions` in `AbstractNode`

ryansuhartanto:
Minor type fix for `AbstractNode` class that moves `node_name` and `context` up to fix the documentation for `resolveSubstitutions`.


cceneag:
好的但我有點不清楚內容該說明什麼

PR #1751, opened 2025-05-19
chore: replace eslint by prettier

ggrossetie:


mojavelinux:
I think you should strongly consider biome. It implements the prettier rules, but also offers features beyond that. And it's a binary, so it has zero JavaScript dependencies.

ggrossetie:
Since I’m more familiar with Prettier, I’ll start by replacing ESLint with Prettier and later evaluate whether it makes sense to switch to Biome.
Prettier also has zero dependency: https://www.npmjs.com/package/prettier?activeTab=dependencies. I was getting increasingly frustrated with ESLint and the standard config, which wasn’t compatible and brought in a lot of dependencies 😬

PR #1753, opened 2025-05-27
Add backtick_javascript comments for Opal 2

janbiedermann:
In Opal 2 backtick_javascript defaults to false, because Kernel#backtick is implemented. Files using inline JavaScript must add the `# backtick_javascript: true` magic comment. This PR does that. I hope i got them all.


ggrossetie:
Thanks!
Opal 2.0 👀 👀 👀

PR #1758, opened 2025-11-21
chore(deps): build against Asciidoctor 2.0.26

ggrossetie:


ggrossetie:
@mojavelinux Something changed in the logger and some tests no longer pass. It seems that the options are being ignored:

```js
it('should be able to get logger\'s info', () => {
  const defaultLogger = asciidoctor.LoggerManager.getLogger()
  expect(defaultLogger.getLevel()).to.equal(0) // was 2
  expect(defaultLogger.getFormatter().$$class.displayName).to.equal('::Formatter') // was ::BasicFormatter
  expect(defaultLogger.getProgramName()).to.equal(Opal.nil) // was 'asciidoctor'
  expect(defaultLogger.getMaxSeverity()).to.equal(undefined) // was 3
})
```

I guess that's because Opal doesn't support kwargs ? https://github.com/opal/opal/blob/master/stdlib/logger.rb#L40-L44

ggrossetie:
Upstream issue: https://github.com/opal/opal/issues/2771

PR #1826, opened 2026-06-06
feat: add --extension CLI option and clarify --require contract

ggrossetie:
- Add `--extension` (repeatable) CLI option that loads an extension file and registers it via its exported `register(registry)` function — both ESM (`export function register`) and CommonJS (`module.exports.register`) are supported
- Fix `--require` to load-only: it no longer auto-calls any exported function.
- Update all eight extension examples to use `export function register(registry)` instead of a default export, and update every extension doc page to show both API and CLI usage
 - Add a `[#cli]` section to `register.adoc` explaining the `--extension` / `--require` distinction and showing ESM + CJS extension formats

### Motivation

Users trying to load an extension with `-r` from the CLI were surprised to find it had no effect. The extension examples exported a default function, but `--require` never called it. This PR introduces `--extension` as the dedicated option for extensions, with a clear contract (`register(registry)`), and documents the distinction.

See also: https://asciidoctor.zulipchat.com/#narrow/channel/279645-users.2Fasciidoctor.2Ejs/topic/Use.20extensions.20from.20command.20line.3F/with/599886016

### Breaking change

`--require` no longer calls `lib.register`. Any extension that relied on this must either switch to `--extension` or self-register using `Extensions.register()` at the top level.

### DIFF
--- a/packages/asciidoctor/lib/cli.js
+++ b/packages/asciidoctor/lib/cli.js
@@ -459,13 +459,17 @@ CLI version ${pkg.version}`
   _prepareExtensions(values) {
     const extensionPaths = values.extension
     if (!extensionPaths || extensionPaths.length === 0) return
+    const logger = LoggerManager.getLogger()
     for (const extensionPath of extensionPaths) {
       const lib = requireLibrary(extensionPath)
-      const registerFn = lib?.register ?? lib?.default
-      if (typeof registerFn === 'function') {
+      if (typeof lib?.register === 'function') {
         Extensions.register(function () {
-          registerFn(this)
+          lib.register(this)
         })
+      } else {
+        logger.warn(
+          `Extension '${extensionPath}' does not export a register function and will be ignored.`
+        )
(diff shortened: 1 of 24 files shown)


packages/asciidoctor/lib/cli.js
```diff
@@ -443,9 +451,21 @@ CLI version ${pkg.version}`
...
+    }
+  }
+
+  /** @internal */
+  _prepareExtensions(values) {
+    const extensionPaths = values.extension
+    if (!extensionPaths || extensionPaths.length === 0) return
+    for (const extensionPath of extensionPaths) {
+      const lib = requireLibrary(extensionPath)
+      const registerFn = lib?.register ?? lib?.default
```
ggrossetie: Should we call the default export if `register` is not found?

I'm still torn, because it can make usage easier if users make mistakes, but it also makes the contract less clear since it still works even when the declaration isn't correct.

PR #1830, opened 2026-06-18
fix(highlight): Move Highlight.js to server-side.

carnoxen:
* make highlight.js run in server-side instead of client-side.
* change extracting callouts. 
* create function injecting callouts.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1371,7 +1371,6 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/boolbase/-/boolbase-1.0.0.tgz",
       "integrity": "sha512-JZOSA7Mo9sNGB8+UjSgzdLtokWAky1zbztM3WRLCbZ70/3cTANmQmOdR7y2g+J0e2WXywy1yS468tY+IruqEww==",
-      "dev": true,
       "license": "ISC"
     },
     "node_modules/brace-expansion": {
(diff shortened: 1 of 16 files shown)


ggrossetie:
Thanks for your contribution, it's greatly appreciated.

However, this projet does not accept contribution without first opening an issue.
It's also important to note that this project must remain compatible with the original Ruby implementation of Asciidoctor. Therefore, we cannot use Highlight.js in server-side mode in the existing Highlight.js syntax highlighter, because the Ruby implementation of Asciidoctor uses Highlight.js in client-side mode.

Please note that it is already possible to provide a custom syntax highlighter: https://docs.asciidoctor.org/asciidoctor.js/4.0/extend/syntax-highlighter/custom-syntax-highlighter/#server-side-highlighting

If you think there are things that could be improved or changed to make it easier to implement a server-side syntax highlighter, feel free to discuss them on the Asciidoctor Zulip chat: https://asciidoctor.zulipchat.com/

Also, we have a fairly strict dependency policy, and the core must not have any dependencies.

For all these reasons, I am closing this PR, as it cannot be merged in its current form.

ggrossetie:
> where can I find the policy?

I don't think we've taken the time to document them yet, but we should, because they're important. We already have a "Policies & Governance" section in the Asciidoctor.js documentation, and I think these details should be added there.

The contribution guidelines should also be expanded, since they currently do not mention these policies.

carnoxen:
@ggrossetie Ok, thanks for response. By the way, this request is not something new. It is an old issue from asciidoctor/asciidoctor#3976. I want to change the callout process, from inject -> parse to parse -> inject.

Anyway, rule is rule, right? I'll check ruby library and find the solution.

ggrossetie:
This extension is not yet compatible with Asciidoctor.js 4.0 but you can take a look at: https://github.com/jirutka/asciidoctor-highlight.js/

PR #1847, opened 2026-07-03
feat(highlight): add build-time mode to the highlightjs syntax highlighter

ggrossetie:
### Summary

Adds an experimental **build-time (server-side) mode** to the built-in `highlightjs` source highlighter. Setting `:highlightjs-mode: build` colourises source blocks with [highlight.js](https://highlightjs.org) **at conversion time** instead of shipping the hljs runtime for the browser to run. The output is self-contained (colours baked into the HTML, theme embedded), so no client-side JavaScript is required.

Without the attribute the behaviour is unchanged (client-side), so this is fully backward compatible and needs no registration.

> Motivation: Asciidoctor.js previously had no built-in server-side highlighter (Ruby offers Rouge/Pygments/CodeRay, none available here). highlight.js is written in JavaScript, so — unlike Asciidoctor Ruby — Asciidoctor.js can actually run it during conversion. Suggested by @mojavelinux.

### Usage

```js
const html = await convert(content, {
  standalone: true,
  attributes: {
    'source-highlighter': 'highlightjs',
    'highlightjs-mode': 'build',
  },
})
```

highlight.js is an **optional peer dependency** — install it where you need build mode:

```
npm install highlight.js
```

### What's supported

- **Callouts** — preserved by the core (stripped before highlighting, re-inserted after); the adapter closes/reopens hljs spans at each line boundary so a conum is never swallowed by a multi-line comment or string span.<br/><br/>

    callouts

- **Line numbers** (`linenums`) — `table` (returns a `[html, offset]` tuple so the core re-aligns callouts) and `inline` modes, via `highlightjs-linenums-mode`.
- `start=` for the first line number.<br/><br/>

    linenums-table

- **Emphasised lines** (`highlight=`) — full-width highlight; a trailing callout is pulled inside the span so it stays on the line, and long lines wrap correctly.<br/><br/>

    highlight

- **Theme** (`highlightjs-theme`) and stylesheet delivery (`highlightjs-stylesheet=embed|link`) — the theme CSS is read from the installed package and embedded by default; a small helper stylesheet is injected for the line-number gutters and emphasis markup.<br/><br/>

    langs

### Design notes

- **Optional & never bundled.** highlight.js is loaded lazily via a runtime `require` (obtained from `await import('node:module')`), which rollup does not follow — so it is never bundled and importing the core never pulls it in. A clear, actionable error is thrown if build mode is requested while the package is missing.
- **Browser-safe by construction.** The Node-only build engine lives in a separate module (`highlightjs-build.js`) that the browser build replaces with a `{ supported: false }` stub (rollup plugin, mirroring the existing template-converter stub). In the browser the adapter logs a warning and **falls back to client-side highlighting** rather than failing. No runtime `process.versions.node` sniffing (which would false-negative on Deno/Bun).
- **`highlight()` and `docinfo()` are now awaited by the core**, so syntax highlighter adapters may be **async** — enabling async engines such as Shiki or starry-night, which cannot be integrated with the previous synchronous contract. Existing synchronous adapters are unaffected.
- Implementation-only adapter members are tagged `@internal` so they don't leak into the generated `.d.ts`.

### Changes

- `src/syntaxHighlighter/highlightjs.js` — build mode added to the existing adapter (mode switch, docinfo/format, helper CSS, warnings).
- `src/syntaxHighlighter/highlightjs-build.js` — new Node-only build engine (`buildEngine`).
- `src/syntax_highlighter.js` — `highlight()`/`docinfo()` documented as awaitable; `format()` gains a `transformContent` hook.
- `src/substitutors.js`, `src/converter/html5.js` — `await` the highlighter `highlight()`/`docinfo()` results.
- `rollup.config.js` — browser stub for the build engine.
- `package.json` — `highlight.js` as an optional `peerDependency` (+ dev dependency).
- Docs: rewrite the server-side section of *Syntax Highlighting* and update *Custom Syntax Highlighter* (async contract, offset tuple, build-mode tip).

### Testing

- New `test/highlightjs-build.test.js` (20 cases): colourising, client-mode + browser fallback, callouts (multi-line comment, last line, multiple), line numbers (table/inline, offset, `start=`), emphasis + callouts (single/multiple, full-width), theme embed/link, missing-theme warning.
- Full suite green (`2763 passing`), `npm run check` clean, types regenerated.

### Notes

- Marked experimental in the changelog.
- The browser bundle was verified to not carry the highlight.js loading code and to leave highlight.js as a runtime require in the Node bundle.


mojavelinux:
I'm excited about this, it looks great!

I did notice that you had to change a few methods to be async. That does seem reasonable given the highlighter might have to do fs operations.

ggrossetie:
> I did notice that you had to change a few methods to be async. That does seem reasonable given the highlighter might have to do fs operations.

Yes, `highlight()` and `docinfo()` are now async. I made this change for Highlight.js, but I think it will also be useful for other syntax highlighters like Shiki, which uses an async API:

```js
const html = await codeToHtml(code, {
  lang: 'javascript',
  theme: 'vitesse-dark'
})
```


And the fact that `docinfo()` is async makes it possible to use the async fs API to load files and include them inline in the HTML.

ggrossetie:
There are a few small visual differences depending on whether you choose to number lines in table mode or inline mode.
In table mode, wrapping is disabled to prevent the line numbers from becoming misaligned. Emphasized lines don't highlight the line number in table mode, but they do highlight the line number in inline mode.
And in inline mode, there's no vertical bar separating the line numbers from the code.
I haven't checked the rendering with Rouge to see if there are also differences there and whether that's intentional. @mojavelinux Is there value in keeping both options?

mojavelinux:
> There are a few small visual differences depending on whether you choose to number lines in table mode or inline mode.

That's darn near impossible to avoid. I have tried very hard to make them look the same for the Ruby-based highlighters, but it's never going to be perfect.

I think small differences are acceptable, and can be refined if people feel like there is a better way to display it.

ggrossetie:
That's fair, and I agree it'll never be pixel-perfect across all highlighters.
That said, looking at the list of small differences between table and inline mode, I'm starting to wonder if the `<table>` layout itself is still pulling its weight. It was a reasonable way to get numbered lines a while back, but it's also the reason we have two diverging modes in the first place, and each new gap between them (wrap, separators, etc.) is really just another symptom of that split.

A more modern option, and one that would let us collapse table/inline into a single approach, is a CSS `display: grid` two-column layout combined with a CSS counter instead of a table. It's what Shiki and recent Pygments/Chroma output use:

```html
<pre class="hljs-with-linenums">
  <code>
    <span class="line">const foo = 1;</span>
    <span class="line highlighted">const bar = 2;</span>
  </code>
</pre>
```

It would mean adopting a single `<span class="line">` per line (with `highlighted` added when relevant) instead of the current table/inline split, and moving the numbering entirely into the bundled stylesheet. It might also simplify the offset-tuple plumbing we added for callout realignment.

For what it's worth, I did manage to get a gutter border working in inline mode too, and wrapping combined with highlighting also works there now.

Capture d’écran 2026-07-06 à 17 17 51

So it's not that inline mode can't be pushed further. But between maintaining that and a single grid-based mode with only a bit of extra markup, I'd rather go with the unified approach. Less to keep in sync long-term!

What do you think?

ggrossetie:
In the end I went with two grid cells per line: an empty gutter cell (`<span class="ln"></span>`) plus the code cell (`<span class="line">…</span>`), laid out as a two-column CSS grid on `<code>`. The line numbers are generated
with a CSS `counter()` on the gutter cell (so they're never selected or copied).

We could simplify the HTML to a single `<span class="line">` and generate the number with `.line::before` (the classic Shiki snippet). It's cleaner markup, but we lose flexibility. That empty gutter cell is what buys us, for free and
with no per-block width math:

- an auto-sized gutter (the `auto` grid column grows to fit any number of digits; a `::before` needs a fixed/computed width);
- emphasis that never covers the line number (the highlight lives in the code column only, since the number is a separate cell);
- a separator that spans the full height of a wrapped line (the gutter cell stretches to the row height);
- callouts that stay intact (a `::before` + `text-indent` hanging-indent breaks inline conums).

So the extra empty span isn't really dead weight. It's the gutter cell, and it keeps the number, the emphasis and the separator cleanly independent. Happy to revisit if we decide those niceties aren't worth it.


```html
<pre class="highlightjs highlight">
<code class="language-ruby hljs linenums" style="counter-reset:line 0" data-lang="ruby">
  <span class="ln"></span><span class="line"><!-- ... --></span>
  <span class="ln"></span><span class="line hljs-ln-highlight"><!-- ... --></span>
  <span class="ln"></span><span class="line"><!-- ... --></span>
</code></pre>
```

Here's the output:

localhost_63342_asciidoctor js_examples_highlightjs-build_demo-default html__ijt=mlnerl1vc1pf7bmai9do0vc2gm _ij_reload=RELOAD_ON_SAVE


If we really want to have the cleanest possible HTML, we could add an option that doesn't use a dedicated span for line numbers, while specifying the limitations of this approach in the documentation (we'd have `highlightjs-linenums-mode` with `inline` (extra span) or `pseudo-element`).

For now, my impression is that this solution covers 95% of the needs and allows for a rendering that's actually quite neat.

